PR #4, opened 2022-01-15
'Solution'

cloudprogramm:


app/main.py
```diff
@@ -1,3 +1,12 @@
 def cache(func):
-    # Write your code here
-    pass
+    arr = []
+
+    def inner(*args, **kwargs):
+        if args in arr:
+            print("Getting from cache")
```
IvanRamyk, line 6: Function should return result in any case. But it just prints `Getting from cache` when `args` are in `arr1`

app/main.py
```diff
@@ -1,3 +1,12 @@
 def cache(func):
-    # Write your code here
-    pass
+    arr = []
```
IvanRamyk, line 2: Could you choose another name for this variable, pls. It's unclear what is stored in it.

IvanRamyk (changes requested):

app/main.py
```diff
@@ -1,3 +1,13 @@
 def cache(func):
-    # Write your code here
-    pass
+    cache_list = []
+
+    def inner(*args, **kwargs):
+        if args in cache_list:
+            print("Getting from cache")
+            return func(*args, **kwargs)
```
IvanRamyk, line 7: The function executes func for each call. The main idea is to somehow store the results for previous arguments and return them for calls with the same arguments without executing func.

IvanRamyk (changes requested):

app/main.py
```diff
@@ -1,3 +1,13 @@
...
+    cache_dict = {}
+
+    def inner(*args, **kwargs):
+        if args in cache_dict:
+            print("Getting from cache")
+            return cache_dict[args]
+        else:
+            cache_dict[args] = func(*args)
+            print("Calculating new result")
+            return func(*args, **kwargs)
```
IvanRamyk, line 11: You call `func` twice: 9 & 11 lines
Everything else is OK)

IvanRamyk (changes requested):

app/main.py
```diff
@@ -1,3 +1,12 @@
...
-    pass
+    cache_dict = {}
+
+    def inner(*args, **kwargs):
+        if args in cache_dict:
+            print("Getting from cache")
+            return cache_dict[args]
+        else:
+            cache_dict[args] = func(*args)
+            print("Calculating new result")
```
IvanRamyk, line 10: Now function doesn't return anything in this branch :(

IvanRamyk (changes requested):

IvanRamyk (approved):
Good job!

PR #5, opened 2022-01-15
'Solution'

denyslipin:


app/main.py
```diff
@@ -1,3 +1,12 @@
 def cache(func):
-    # Write your code here
-    pass
+    dic = {}
+
+    def inner(*args):
+        nonlocal dic
```
IvanRamyk, line 5: Why to use nonlocal here? It is unnecessary

app/main.py
```diff
@@ -1,3 +1,12 @@
 def cache(func):
-    # Write your code here
-    pass
+    dic = {}
```
IvanRamyk, line 2: Could you rename this variable pls? Name `dic` doesn't say anything about what is stored in it.

IvanRamyk (changes requested):

IvanRamyk (approved):
Well done!

PR #6, opened 2022-01-16
Solution

NikitaZH1812:


app/main.py
```diff
@@ -1,3 +1,11 @@
 def cache(func):
-    # Write your code here
-    pass
+    cache_dict = {}
+
+    def inner(*args):
+        if args in cache_dict:
+            print("Getting from cache")
```
IvanRamyk, line 6: When `args` are in cache, result should also be returned. This function returns result only in else branch

IvanRamyk (changes requested):

app/main.py
NikitaZH1812: oh, my bad, fixed

IvanRamyk: Good job!

IvanRamyk (approved):

PR #7, opened 2022-01-16
Solution

nikon4uk:


app/main.py
```diff
@@ -1,3 +1,14 @@
 def cache(func):
-    # Write your code here
-    pass
+    history = {}
+
+    def wrapper(*args, **kwargs):
+        nonlocal history
```
IvanRamyk, line 5: You don't need to use nonlocal here. It can be used for reassigning variables. If you need just to change some fields of dict it is unnecessary.

IvanRamyk (changes requested):

IvanRamyk (approved):
Well done!

PR #9, opened 2022-01-16
Solution

Geksus:


app/main.py
```diff
@@ -1,3 +1,11 @@
...
-    pass
+    ls = []
+
+    def wrapper(*args, **kwargs):
+        if args in ls:
+            print("Getting from cache")
+        else:
+            ls.append(args)
+            print("Calculating new result")
+        return func(*args, **kwargs)
```
IvanRamyk, line 10: The function executes `func` for each call. The main idea is to somehow store the results for previous arguments and return them for calls with the same arguments without executing `func`.

IvanRamyk (changes requested):

IvanRamyk (approved):
Well done!

PR #10, opened 2022-01-17
Solution

alix-mak:


app/main.py
```diff
@@ -1,3 +1,21 @@
...
+            print("Getting from cache")
+            return memory[args][1]
+
+        else:
+            memory[args] = func.__name__, func(*args)
+            print("Calculating new result")
+            return func(*args)
+
+    return wrapper
     pass
```
IvanRamyk, line 21: Remove `pass`, pls. We use it only for empty functions

app/main.py
```diff
@@ -1,3 +1,21 @@
...
+
+
 def cache(func):
-    # Write your code here
+
+    memory = {}
+
+    @functools.wraps(func)
+    def wrapper(*args):
+        nonlocal memory
```
IvanRamyk, line 10: You don't need to use nonlocal here. It can be used for reassigning variables. If you need just to change some fields of dict it is unnecessary.

app/main.py
```diff
@@ -1,3 +1,21 @@
...
+        if args in memory and memory[args][0] == func.__name__:
```
IvanRamyk, line 11: You don't actually need to remember or check function name, because decorator creates a new dictionary for each function (line 6). 

It would be needed if memory was a global variable. In that case dictionary would be the same for all decorated functions.

IvanRamyk (changes requested):

IvanRamyk (approved):
Good job!

PR #11, opened 2022-01-17
'Solution'

liuda-chuzdiuk:


app/main.py
```diff
@@ -1,3 +1,13 @@
...
+    unique = set()
+
+    def inner(*args):
+        if args not in unique:
+            unique.add(args)
+            print("Calculating new result")
+            return func(*args)
+        else:
+            print("Getting from cache")
+            return unique
```
IvanRamyk, line 11: Function should return result of executing `func` with given arguments, but it returns `unique` instead

IvanRamyk (changes requested):

app/main.py
IvanRamyk: This result should be stored after first execution with such args

IvanRamyk (approved):
Well done!

PR #12, opened 2022-01-17
'Solution'

serhii56465:


app/main.py
```diff
@@ -1,3 +1,13 @@
 def cache(func):
-    # Write your code here
-    pass
+    caches = {}
+
+    def inner(*args):
+        if args in caches.keys():
```
IvanRamyk, line 5: You can write it without `keys()`:
`if args in caches:`

app/main.py
```diff
@@ -1,3 +1,13 @@
...
+            print("Getting from cache")
+            return caches[args]
+        else:
+            print("Calculating new result")
+            result = func(*args)
+            caches[tuple(args)] = result
```
IvanRamyk, line 11: Variable `args` is already a tuple. You don't need to convert it

IvanRamyk (changes requested):

IvanRamyk (approved):
Well done!

PR #13, opened 2022-01-17
Solution

TPolina:


app/main.py
```diff
@@ -1,3 +1,13 @@
...
+    storage = dict()
+
+    def wrapper(*args):
+        if args in storage:
+            print("Getting from cache")
+            return storage[args]
+        else:
+            print("Calculating new result")
+            storage[args] = func(*args)
+            return func(*args)
```
IvanRamyk, line 11: You call `func` twice. It is better to return stored result

IvanRamyk (changes requested):

IvanRamyk (approved):
Well done!!

PR #1, opened 2022-01-12
Add py-cache-decorator github task

MasterpieceElbow:

PR #14, opened 2022-01-17
'Solution' #1

8Men5Camels:


app/main.py
```diff
@@ -1,3 +1,14 @@
 def cache(func):
-    # Write your code here
-    pass
+    hole = dict()
+
+    def inner(*args):
+        if args not in hole:
+            hole[args] = func(*args)
+            print('Calculating new result')
+            return func
```
IvanRamyk, line 8: Function should return result of executing `func`, not a `func` itself

IvanRamyk (changes requested):

app/main.py
```diff
@@ -1,3 +1,14 @@
 def cache(func):
-    # Write your code here
-    pass
+    hole = dict()
+
+    def inner(*args):
+        if args not in hole:
+            hole[args] = func(*args)
+            print('Calculating new result')
+            return func(*args)
```
IvanRamyk, line 8: Now you call `func` twice) It is better to return stored result)

IvanRamyk (changes requested):

IvanRamyk (approved):
Well done!

PR #15, opened 2022-01-17
Solution

IgorOlifirenko:


app/main.py
```diff
@@ -1,3 +1,17 @@
...
-    pass
+
+    repository = {}
+
+    @functools.wraps(func)
+    def wrapper(*args):
+        if args not in repository:
+            result = func(*args)
+            repository.update({args: result})
+            print("Calculating new result")
```
IvanRamyk, line 13: Function should always return a result. Now it just prints a message in this branch

IvanRamyk (changes requested):

IvanRamyk (approved):
Good job!

PR #17, opened 2022-01-17
'Solution'

alexandra-halytska:


app/main.py
```diff
@@ -1,3 +1,14 @@
 def cache(func):
-    # Write your code here
-    pass
+    functions = dict()
+    arguments = set()
+
+    def inner(*args):
+        if func.__name__ in functions and args in arguments:
```
IvanRamyk, line 6: You don't actually need to handle function names, because there are separate dict/set for each decorated function created in lines 2-3.

app/main.py
```diff
@@ -1,3 +1,14 @@
...
+            print("Getting from cache")
+            return functions[func.__name__]
+        else:
+            print("Calculating new result")
+            functions[func.__name__] = func(*args)
```
IvanRamyk, line 11: Dictionary should store results for different **args** not for different **functions**

IvanRamyk (changes requested):

PR #18, opened 2022-01-17
Develop2

alexandra-halytska:


app/main.py
```diff
@@ -1,3 +1,12 @@
 def cache(func):
-    # Write your code here
-    pass
+    functions = dict()
```
IvanRamyk, line 2: Could you change the name of this variable? It seems that some functions are stored in it

IvanRamyk (changes requested):

PR #21, opened 2022-01-17
Solution

Alexandr-Semenishin:


app/main.py
```diff
@@ -1,3 +1,13 @@
 def cache(func):
-    # Write your code here
-    pass
+    cache_dict = {}
+
+    def inner(*args):
+        if hash(tuple(args)) in cache_dict:
```
IvanRamyk, line 5: Variable `args` is already a tuple. You don't need to convert it

app/main.py
```diff
@@ -1,3 +1,13 @@
...
+            print("Getting from cache")
+            return cache_dict[hash(tuple(args))]
```
IvanRamyk, line 7: When you add something to the dict, you don't need to call hash function. Dict does it by itself

IvanRamyk (changes requested):

app/main.py
Alexandr-Semenishin: the goal is to store not the object itself, but its cache, and take the data also by hash. According to the task "You have a big array with data..."

IvanRamyk: Actually, dict doesn't store keys itself. It use a hash of the key to find a proper place for value in the hash table 
You can check [this](https://pastebin.com/82cJiHeL) script. Sizes of dicts are equal in that example

app/main.py
```diff
@@ -1,3 +1,13 @@
 def cache(func):
-    # Write your code here
-    pass
+    cache_dict = {}
+
+    def inner(*args):
+        if hash(args) in cache_dict:
```
IvanRamyk, line 5: Dict use hash function by itself. You don't need to call it

IvanRamyk (changes requested):

PR #22, opened 2022-01-18
'Solution'

Zikkolion:


app/main.py
```diff
@@ -1,3 +1,14 @@
 def cache(func):
-    # Write your code here
-    pass
+    c_dict = {}
```
IvanRamyk, line 2: Could you change the name of this variable? It is unclear what `c_dict` means

app/main.py
```diff
@@ -1,3 +1,14 @@
...
+
+    def inner(*args):
+        if args in c_dict:
+            print("Getting from cache")
+            return c_dict[args]
+        else:
+            print("Calculating new result")
+            s = func(*args)
```
IvanRamyk, line 10: Could you change this name as well?

IvanRamyk (changes requested):

IvanRamyk (approved):

PR #29, opened 2022-03-19
Solution task about cache decorator. Katiukha

KatiukhaO:


app/main.py
```diff
@@ -1,3 +1,29 @@
...
+
+
+@cache
+def long_time_func(a, b, c):
+    return (a ** b ** c) % (a * c)
+
+
+@cache
+def long_time_func_2(n_tuple, power):
+    return [number ** power for number in n_tuple]
```
TPolina, line 29: there is no need to write functions and decorate them. if this was for testing, better not to push this

TPolina (changes requested):

TPolina (approved):
Good job!

TPolina (approved):

PR #33, opened 2022-05-12
Solution

YuriiKindrat:


app/main.py
```diff
@@ -1,3 +1,11 @@
 def cache(func):
-    # Write your code here
-    pass
+    cash = {}
+
+    def inner(*args, **kwargs):
+        nonlocal cash
```
MasterpieceElbow, line 5: You don't have to write `nonlocal` if you change a mutable variable.

app/main.py
```diff
@@ -1,3 +1,11 @@
...
+        if args not in cash:
+            print("Calculating new result")
+            cash[args] = func(*args, **kwargs)
+        else:
+            print("Getting from cache")
```
MasterpieceElbow, line 10: You should return a result of the function. I have fixed test_main.py today, please update your `tests/test_main.py` from origin/master via `git checkout master -- tests/test_main.py`

MasterpieceElbow (changes requested):

app/main.py
```diff
@@ -1,3 +1,12 @@
 def cache(func):
-    # Write your code here
-    pass
+    cash = {}
+
+    def inner(*args, **kwargs):
+        if args not in cash:
+            print("Calculating new result")
+            cash[args] = func(*args, **kwargs)
+            return func(*args, **kwargs)
```
TPolina, line 8: You call `func` twice. Better return stored data

TPolina (changes requested):

app/main.py
```diff
@@ -1,3 +1,12 @@
...
+    cash = {}
+
+    def inner(*args, **kwargs):
+        if args not in cash:
+            print("Calculating new result")
+            cash[args] = func(*args, **kwargs)
+            return cash[args]
+        else:
+            print("Getting from cache")
+            return cash[args]
```

TPolina (changes requested):

TPolina (approved):

TPolina (approved):

PR #34, opened 2022-05-12
Solution

4ndyua:


app/main.py
```diff
@@ -1,3 +1,11 @@
...
-    pass
+    c_dict = {}
+
+    def inner(*args):
+        if args in c_dict:
+            print('Getting from cache')
+            return c_dict[args]
+        else:
+            print('Calculating new result')
+            c_dict[args] = func(*args)
```
TPolina, line 10: when you calculating new result you should return it)

TPolina (changes requested):

app/main.py
```diff
@@ -1,3 +1,12 @@
...
+            return c_dict[args]
```

TPolina (changes requested):

TPolina (approved):

PR #35, opened 2022-05-12
Solution

MykolaIF:


app/main.py
```diff
@@ -1,3 +1,14 @@
...
+
+    def wrapper(*args):
+        nonlocal dict_with_args
+        if args in dict_with_args:
+            print("Getting from cache")
+            return dict_with_args[args]
+
+        dict_with_args.update({args: func(*args)})
+        print("Calculating new result")
+        return func
```
IvanRamyk, line 12: Here you should return a calculated result, not a `func` itself.
Example how it works:
```
@cache
def add(a, b):
    return a + b


print(add(2, 3))
```
Output:
```
Calculating new result
<function add at 0x1009a3e50>
```
How it should work:
```
Calculating new result
5
```

app/main.py
```diff
@@ -1,3 +1,14 @@
 def cache(func):
-    # Write your code here
-    pass
+    dict_with_args = {}
+
+    def wrapper(*args):
+        nonlocal dict_with_args
```
IvanRamyk, line 5: Nonlocal is only used if you need to redefine the dictionary. If you are only updating its values, this is redundant.

IvanRamyk (changes requested):

app/main.py
```diff
@@ -1,3 +1,13 @@
...
+    dict_with_args = {}
+
+    def wrapper(*args):
+        if args in dict_with_args:
+            print("Getting from cache")
+            return dict_with_args[args]
+
+        dict_with_args.update({args: func(*args)})
+        print("Calculating new result")
+        return func(*args)
```
IvanRamyk, line 11: Here you call the function a second time. The first time you call it in the 9th line.
It's better you use the already calculated value from the dict :)

IvanRamyk (changes requested):

IvanRamyk (approved):
Good job!

PR #36, opened 2022-05-12
Decorator done

Flashmobber:


app/main.py
```diff
@@ -1,3 +1,13 @@
 def cache(func):
-    # Write your code here
-    pass
+    results = {}
+
+    def inner(*args):
+        if args in results is not None:
```
TPolina, line 5: 'is not None' is redundant here

app/main.py
```diff
@@ -1,3 +1,13 @@
...
+            print("Getting from cache")
+            return results[args]
+
+        print("Calculating new result")
+        results[args] = func(*args)
+        return results[args]
```
TPolina, line 11: try to write your decorator in such way that the return statement is not duplicated

TPolina (changes requested):

TPolina (approved):
Good job!

TPolina (approved):

PR #37, opened 2022-05-13
'Solution'

DmytroSvirsa:


app/main.py
```diff
@@ -1,3 +1,13 @@
...
+    results_for_cache = {}
+
+    def inc_sort(*args):
+        if args in results_for_cache:
+            print("Getting from cache")
+            return results_for_cache[args]
+        else:
+            print("Calculating new result")
+            results_for_cache[args] = func(*args)
+            return func(*args)
```
TPolina, line 11: You call `func` twice. Better return stored result

TPolina (changes requested):

app/main.py
```diff
@@ -1,3 +1,13 @@
...
+    results_for_cache = {}
+
+    def inc_sort(*args):
+        if args in results_for_cache:
+            print("Getting from cache")
+            return results_for_cache[args]
+        else:
+            print("Calculating new result")
+            results_for_cache[args] = func(*args)
+            return results_for_cache[args]
```

TPolina (changes requested):

app/main.py

TPolina (changes requested):

TPolina (approved):

PR #38, opened 2022-05-13
'Solution'

Magus6784:


app/main.py
```diff
@@ -1,3 +1,12 @@
...
+    res = {}
+
+    def inner(*args, **kwargs):
+        if args in res:
+            print("Getting from cache")
+            return res[args]
+        else:
+            print("Calculating new result")
+            res[args] = func(*args)
+            return res[args]
```
TPolina, line 11: you have a duplicate return. Better write it once outside if-else statement

TPolina (changes requested):

TPolina (approved):

PR #42, opened 2022-05-13
Develop

BohdanKryven:
Finally made my homework :)

### DIFF
--- a/app/main.py
+++ b/app/main.py
@@ -4,10 +4,9 @@ def cache(func):
     def wrapper(*args):
         if args in cache_dict.keys():
             print("Getting from cache")
-            return cache_dict[args]
         else:
             print("Calculating new result")
             cache_dict[args] = func(*args)
-            return cache_dict[args]
+        return cache_dict[args]
 
     return wrapper


app/main.py
```diff
@@ -1,3 +1,13 @@
...
+    cache_dict = {}
+
+    def wrapper(*args):
+        if args in cache_dict.keys():
+            print("Getting from cache")
+            return cache_dict[args]
+        else:
+            print("Calculating new result")
+            cache_dict[args] = func(*args)
+            return func(*args)
```
TPolina, line 12: you call `func` twice, better return stored data

TPolina (changes requested):

app/main.py
```diff
@@ -1,3 +1,13 @@
...
+    cache_dict = {}
+
+    def wrapper(*args):
+        if args in cache_dict.keys():
+            print("Getting from cache")
+            return cache_dict[args]
+        else:
+            print("Calculating new result")
+            cache_dict[args] = func(*args)
+            return cache_dict[args]
```

TPolina (changes requested):

PR #44, opened 2022-05-13
Made decorator cache that stores results of completed runs with diffe…

netiazhevskyi:
…rent arguments


Abnormaltype (approved):
Well done!

PR #45, opened 2022-05-14
Solution

RomaTsybulia:


app/main.py
```diff
@@ -1,3 +1,12 @@
 def cache(func):
-    # Write your code here
-    pass
+    result = {}
+
+    def caching(*args, **kwargs):
+
+        if f'{args}' in result.keys():
```
TPolina, line 6: there is no need to make string from args

app/main.py
```diff
@@ -1,3 +1,12 @@
...
+            print('Getting from cache')
+        else:
+            result[f'{args}'] = func(*args, **kwargs)
+            print('Calculating new result')
```
TPolina, line 10: you also should return result of the function

TPolina (changes requested):

app/main.py
TPolina: not fixed yet

app/main.py
TPolina: when you calculate new result you didn't return anything

TPolina (changes requested):

app/main.py
```diff
@@ -1,3 +1,14 @@
...
+
+    def caching(*args, **kwargs):
+
+        if args in result.keys():
+            print('Getting from cache')
+            return result[args]
+        else:
+            result[args] = func(*args, **kwargs)
+            print('Calculating new result')
+            return result[args]
```
TPolina, line 12: you have a duplicate `return`, take it outside if-else statement

TPolina (changes requested):

TPolina (approved):

PR #49, opened 2022-05-14
'Solution'

daria-niklaevna:


app/main.py
```diff
@@ -1,3 +1,14 @@
 def cache(func):
-    # Write your code here
-    pass
+    result = {}
+
+    def decorate(*args, **kwargs):
+        nonlocal result
```
TPolina, line 5: this statement is redundant, as dict is a mutable type

app/main.py
```diff
@@ -1,3 +1,14 @@
...
+        if args in result:
+            print('Getting from cache')
+            return result[args]
+        else:
+            result[args] = func(*args, **kwargs)
+            print('Calculating new result')
+            return func(*args, **kwargs)
```
TPolina, line 12: you call `func` twice, better return stored data

TPolina (changes requested):

app/main.py
```diff
@@ -1,3 +1,13 @@
...
+    result = {}
+
+    def decorate(*args, **kwargs):
+        if args in result:
+            print('Getting from cache')
+            return result[args]
+        else:
+            result[args] = func(*args, **kwargs)
+            print('Calculating new result')
+            return result[args]
```

TPolina (changes requested):

TPolina (approved):

PR #50, opened 2022-05-15
'Solution'

Igor-Cegelnyk:


Abnormaltype (changes requested):
All tests should are passed

app/main.py
```diff
@@ -1,3 +1,12 @@
 def cache(func):
-    # Write your code here
-    pass
+    lis = []
```
Abnormaltype, line 2: Bad and non-informative variable name.

It is better to use the dictionary to store since the key will store the parameters, and the value is - the result of the execute function)

app/main.py
```diff
@@ -1,3 +1,12 @@
...
+
+    def inner(*args):
+        if args not in lis:
+            lis.append(args)
```
Abnormaltype, line 6: `If decorated function runs with repeating arguments it should return stored result instead of calling the function again. Decorator cache creating for decorating only functions that take immutable arguments.`

Abnormaltype (changes requested):

Abnormaltype (changes requested):

Abnormaltype (approved):
Well done!

PR #52, opened 2022-05-15
cache_decorator solution

Y-Havryliv:


app/main.py
```diff
@@ -1,3 +1,18 @@
...
+        if sublist in cached_values_list:
+            print("Getting from cache")
+        else:
+            print("Calculating new result")
+
+        cached_values_list += [[*args]]
+
+        return func(*args, **kwargs)
+
+    return inner
```
TPolina, line 18: your approach is incorrect: you should store arguments and the result of the function with these arguments, and return this result without calling function when arguments are the same.

TPolina (changes requested):

TPolina (approved):

TPolina (approved):

PR #55, opened 2022-05-15
Develop

Hectorovich:


app/main.py
```diff
@@ -1,3 +1,16 @@
-def cache(func):
-    # Write your code here
-    pass
+def cache(func):
+    data = dict()
+
+    def wrapper(*args):
+        nonlocal data
```
TPolina, line 5: this statement is redundant here, because you are not rewriting data variable, you just modify dict that is under this variable

TPolina (changes requested):

TPolina (approved):

TPolina (approved):

TPolina (approved):

PR #57, opened 2022-05-16
'Solution_16/05_#1'

CheshkoSergey:


app/main.py
```diff
@@ -1,3 +1,14 @@
...
+    def inner(*args, **kwargs):
+
+        list_args = [*args]
+        if list_args in long_time_list:
+            print("Getting from cache")
+        else:
+            print("Calculating new result")
+        long_time_list.append([*args])
+        return func(*args, **kwargs)
+    return inner
```
TPolina, line 14: your approach is incorrect: you should store arguments and the result of the function with these arguments and return this result without calling function again when arguments are the same.

TPolina (changes requested):

app/main.py
```diff
@@ -1,3 +1,15 @@
 def cache(func):
-    # Write your code here
-    pass
+
+    long_time_list = {}
+
+    def inner(*args, **kwargs):
+
+        list_args = [*args]
+        if str(list_args) in long_time_list:
```
TPolina, line 8: you perform redundant operations: `args` is a tuple, so you can just store it as key in dictionary instead of transforming it into list and then string.

app/main.py
```diff
@@ -1,3 +1,15 @@
...
+            print("Getting from cache")
+            return long_time_list[str(list_args)]
+        else:
+            print("Calculating new result")
+            long_time_list[str(list_args)] = func(*args, **kwargs)
+            return func(*args, **kwargs)
```
TPolina, line 14: you call `func` twice, better return stored data

TPolina (changes requested):

app/main.py
```diff
@@ -1,3 +1,14 @@
...
+
+    def inner(*args, **kwargs):
+
+        if tuple(args) in long_time_list:
+            print("Getting from cache")
+            return long_time_list[tuple(args)]
+        else:
+            print("Calculating new result")
+            long_time_list[tuple(args)] = func(*args, **kwargs)
+            return long_time_list[tuple(args)]
```
TPolina, line 13: you have duplicate `return`, take it outside it-else statement

TPolina (changes requested):

app/main.py
```diff
@@ -1,3 +1,13 @@
 def cache(func):
-    # Write your code here
-    pass
+
+    long_time_list = {}
+
+    def inner(*args, **kwargs):
+
+        if tuple(args) in long_time_list:
```
TPolina, line 7: there is no need to convert `args` into `tuple` because `args` is already a `tuple`

app/main.py
```diff
@@ -1,3 +1,13 @@
...
+            print("Getting from cache")
+        else:
+            print("Calculating new result")
+            long_time_list[tuple(args)] = func(*args, **kwargs)
```
TPolina, line 11: same here

app/main.py
```diff
@@ -1,3 +1,13 @@
...
+        return long_time_list[tuple(args)]
```
TPolina, line 12: and here

TPolina (changes requested):

MasterpieceElbow (approved):

PR #61, opened 2022-05-16
Solution

vasylhnatiuk:


app/main.py
```diff
@@ -1,3 +1,18 @@
...
+        if mutable:
+            return func(*args)
+        else:
+            if args in cache_dict.keys():
+                print("Getting from cache")
+                return cache_dict[args]
+            else:
+                cache_dict[args] = func(*args)
+                print("Calculating new result")
+                return func(*args)
```
TPolina, line 17: you call `func` twice, better return stored data

app/main.py
```diff
@@ -1,3 +1,19 @@
...
-    pass
+    cache_dict = {}
+
+    def inner(*args):
+        mutable = False
+        for argument in args:
+            if type(argument) in [list, dict, set]:
+                mutable = True
+        if mutable:
+            return func(*args)
```
TPolina, line 10: this check is redundant, because according to task description we will use this decorator only with immutable types of arguments

app/main.py
```diff
@@ -1,3 +1,19 @@
...
+        else:
+            if args in cache_dict.keys():
+                print("Getting from cache")
+                return cache_dict[args]
+            else:
+                cache_dict[args] = func(*args)
+                print("Calculating new result")
+                return cache_dict[args]
```
TPolina, line 18: you have a duplicate `return`, take it outside if-else statement

app/main.py
vasylhnatiuk: I accidentally complicated the task for myself)

TPolina (approved):

PR #62, opened 2022-05-16
Solution

kamilla-boiko:


app/main.py
```diff
@@ -1,3 +1,12 @@
...
-    pass
+    cache_result = set()
+
+    def wrapper(*args, **kwargs):
+        if args not in cache_result:
+            cache_result.add(args)
+            print("Calculating new result")
+        else:
+            print("Getting from cache")
+        return func(*args, *kwargs)
```
Abnormaltype, line 10: your approach is incorrect: you should store the result of the function with certain arguments and return this result without calling this function when the arguments are the same.
hint: use a dictionary as the storage

Abnormaltype (changes requested):

Abnormaltype (approved):
Well done!

PR #63, opened 2022-05-16
'Solution'

YuraBeseda:


app/main.py
```diff
@@ -1,3 +1,26 @@
...
+        local_cache = []
+        for arg in args:
+            local_cache.append(arg)
+        if local_cache in caches:
+            print("Getting from cache")
+        else:
+            print("Calculating new result")
+            caches.append(local_cache)
+
+        return func(*args)
```
TPolina, line 14: your approach is incorrect: you should store arguments and the result of the function with these arguments and return this result without calling function when arguments are the same.

app/main.py
```diff
@@ -1,3 +1,26 @@
...
+
+
+@cache
+def long_time_func(a, b, c):
+    return (a ** b ** c) % (a * c)
+
+
+@cache
+def long_time_func_2(n_tuple, power):
+    return [number ** power for number in n_tuple]
```
TPolina, line 26: there is no need to write functions and decorate them. if it is for testing, still better not to push this

TPolina (changes requested):

TPolina (approved):
Good job!

PR #64, opened 2022-05-19
'Solution'

Yurnerroo:


app/main.py
```diff
@@ -1,3 +1,17 @@
 def cache(func):
-    # Write your code here
-    pass
+    cached_values = []
+
+    def inner(*args):
+
+        lst = [*args]
```
MasterpieceElbow, line 6: You should not transform it into a list because it is already a tuple. Also, [*args] is a bad transformation tuple into a list, better use `list(args)`.

app/main.py
```diff
@@ -1,3 +1,17 @@
...
+
+        if lst in cached_values:
```
MasterpieceElbow, line 8: `cached_values` is better to be a `dict`, because checking it item is in a dict is cheaper than checking if item in a list.

app/main.py
```diff
@@ -1,3 +1,17 @@
...
+            print("Getting from cache")
+        else:
+            print("Calculating new result")
+
+        cached_values.append([*args])
+
+        return func(*args)
```
MasterpieceElbow, line 15: You always calculate a new value and don't use the value calculated before. There is no point in that decorator, rewrite.

MasterpieceElbow (changes requested):

PR #70, opened 2022-05-27
Solution

Splend1ed:


app/main.py
```diff
@@ -1,3 +1,24 @@
 def cache(func):
-    # Write your code here
-    pass
+    cached_data = {}
+
+    def wrap(*args):
+        nonlocal cached_data
+        for key, value in {args: func(*args)}.items():
+            if str(key) not in cached_data:
+                cached_data.update({f"{args}": func(*args)})
```
IvanRamyk, line 8:
```suggestion
                cached_data.update(args: func(*args)})
```
Converting to string isn't correct because you would consider a tuple (1, 2) and a string "(1, 2)" as the same args

IvanRamyk (changes requested):

app/main.py
```diff
@@ -1,3 +1,23 @@
 def cache(func):
-    # Write your code here
-    pass
+    cached_data = {}
+
+    def wrap(*args):
+        nonlocal cached_data
```
Abnormaltype, line 5: Why use nonlocal here? It is unnecessary

Abnormaltype (changes requested):

Abnormaltype (approved):
Good job!

PR #74, opened 2022-06-02
'Solution'

MilManIT:


Abnormaltype (changes requested):
Good job! but fix some suggestion

app/main.py
```diff
@@ -1,3 +1,25 @@
...
+    return (a ** b ** c) % (a * c)
+
+
+@cache
+def long_time_func_2(n_tuple, power):
+    return [number ** power for number in n_tuple]
+
+
+long_time_func(1, 2, 3)
+long_time_func(2, 2, 3)
```
IvanRamyk, line 25: It's bad practice to leave some test calls of functions in your modules. These lines execute every time you import the module. 
You should either remove these calls or wrap them using `if __name__ == "__main__"` (the sense of this condition is discussed in the `Modules and imports` topic)

IvanRamyk (approved):

PR #76, opened 2022-06-05
Solution

Vlad-Peresta:


app/main.py
```diff
@@ -1,3 +1,19 @@
...
+    function_results = {}
+
+    @wraps(func)
+    def func_wrapper(*args):
+        if func.__name__ in function_results and \
+                args in function_results[func.__name__]:
+            print("Getting from cache")
+        else:
+            result = func(*args)
+            function_results.setdefault(func.__name__, {args: result})
```
IvanRamyk, line 14: You don't need different sub-dictionaries for different functions/
When you decorate the function the new dictionary is created (line 5). So you will have a separate instance of `function_results` for every function

IvanRamyk (changes requested):

app/main.py
Vlad-Peresta: I supposed decorator would be more flexible if remove variable function_results outside decorator. There would store all information {function: {args: result}} in the function_results.

IvanRamyk (approved):
Well done!

PR #77, opened 2022-06-06
Solution

muhactive:


app/main.py
```diff
@@ -1,3 +1,19 @@
 def cache(func):
-    # Write your code here
-    pass
+
+    my_memory = {}
```
IvanRamyk, line 3: Please, rename the variable
Use a descriptive name, this dictionary isn't actually your memory)

app/main.py
```diff
@@ -1,3 +1,19 @@
...
+
+            print("Getting from cache")
+            return my_memory[args]
+
+        else:
+
+            my_memory[args] = func(*args)
+            print("Calculating new result")
+
+            return func(*args)
```
IvanRamyk, line 17: You calculate the result twice: in the 14th and 17th lines

IvanRamyk (changes requested):

app/main.py
```diff
@@ -1,3 +1,19 @@
...
+
+            print("Getting from cache")
+            return stores_results[args]
+
+        else:
+
+            stores_results[args] = func(*args)
+            print("Calculating new result")
+
+            return stores_results[args]
```
IvanRamyk, line 17: you have a duplicate `return`, so take it outside the if-else statement; try to rewrite your decorator not to have `return` duplicated

IvanRamyk (changes requested):

PR #78, opened 2022-06-07
Solution

VladyslavKlevanskyi:


app/main.py
```diff
@@ -1,3 +1,13 @@
...
+    cached = {}
+
+    def wrapper(*args):
+        if args not in cached:
+            cached[args] = func(*args)
+            print("Calculating new result")
+            return cached[args]
+        else:
+            print("Getting from cache")
+            return cached[args]
```
IvanRamyk, line 11: You have a duplicate `return`, so take it outside the if-else statement; try to rewrite your decorator not to have `return` duplicated

IvanRamyk (changes requested):

app/main.py
VladyslavKlevanskyi: Done!

IvanRamyk (approved):
Good job!

PR #80, opened 2022-06-10
Solution

AlexBOVUKR:


app/main.py
```diff
@@ -1,3 +1,14 @@
 def cache(func):
-    # Write your code here
-    pass
+    cache_of_args = {}
+
+    def wrapper(*args, **kwargs):
+        i = args
```
Abnormaltype, line 5: you don't need `i` variable, it's unnecessary in this case

Abnormaltype (changes requested):

app/main.py
AlexBOVUKR: ok, going to change

app/main.py
```diff
@@ -1,3 +1,13 @@
...
+
+    def wrapper(*args, **kwargs):
+        if args not in cache_of_args:
+            res = func(*args, **kwargs)
+            cache_of_args[args] = res
+            print("Calculating new result")
+            return res
+        if args in cache_of_args.keys():
+            print("Getting from cache")
+            return cache_of_args[args]
```
Abnormaltype, line 12: Duplication of the `return` statement, You can move it from `if`

Abnormaltype (changes requested):

app/main.py
AlexBOVUKR: there is 2 different **return** from different **if**... i cant understand how correct properly

Abnormaltype: you should return cached valure from `cache_of_args`

AlexBOVUKR: sorry, i dont understand... i tried to run in debager and it works, values saves to Dict and then take and return it from cache(when the arguments repeats) or return and save result (when arguments was new)

Abnormaltype (changes requested):

app/main.py
```diff
@@ -1,3 +1,13 @@
...
-    # Write your code here
-    pass
+    cache_of_args = {}
+
+    def wrapper(*args, **kwargs):
+        if args not in cache_of_args:
+            res = func(*args, **kwargs)
+            cache_of_args[args] = res
+            print("Calculating new result")
+            return res
```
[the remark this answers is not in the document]
AlexBOVUKR: when i remove this return i fail tests

Abnormaltype (approved):
Well done!

PR #86, opened 2022-06-23
Solution

InnaKushnir:


app/main.py
```diff
@@ -1,3 +1,15 @@
 def cache(func):
-    # Write your code here
-    pass
+    cache_dict = {}
+
+    def inner(*args):
+        if len(cache_dict) < 1:
```
IvanRamyk, line 5: This is a redundant condition.
You check if args is present in the line 9

app/main.py
```diff
@@ -1,3 +1,15 @@
...
+            print("Calculating new result")
+            cache_dict[args] = func(*args)
+        else:
+            if args in cache_dict.keys():
```
```suggestion
            if args in cache_dict:
```
More clear way to check key presence

app/main.py
```diff
@@ -1,3 +1,15 @@
...
+                print("Getting from cache")
+            else:
+                print("Calculating new result")
+                cache_dict[args] = func(*args)
+        return cache_dict.get(args)
```
IvanRamyk, line 14:
```suggestion
        return cache_dict[args]
```
More clear way to get value by key if you are sure it is present in the dictionary

IvanRamyk (changes requested):

IvanRamyk (approved):
Well done!

IvanRamyk (approved):
Already approved

PR #95, opened 2022-07-04
'Solution'

Kev1nXD:


app/main.py
```diff
@@ -1,3 +1,15 @@
...
+    cache_results = {}
+
+    def wrapped(*args, **kwargs):
+        function_args = tuple([*args])
+        if function_args not in cache_results:
+            res = func(*args)
+            cache_results[function_args] = res
+            print("Calculating new result")
+            return res
+        else:
```
IvanRamyk, line 11: Please, remove the redundant `else`

app/main.py
```diff
@@ -1,3 +1,15 @@
...
+            print("Getting from cache")
+            return cache_results.get(function_args)
```
IvanRamyk, line 13: At this moment you are sure that args are in the dictionary, so you can use the following way to get the result
```suggestion
            return cache_results[function_args]
```

IvanRamyk (changes requested):

IvanRamyk (approved):
Well done!

PR #97, opened 2022-07-05
use a dictionary to store both args and results

DanRoman-code:


app/main.py
```diff
@@ -1,3 +1,15 @@
 def cache(func):
-    # Write your code here
-    pass
+    list_arguments = {}
```
IvanRamyk, line 2: If it's a dictionary maybe you should change its name?

IvanRamyk (changes requested):
Small code-style  fixes left

IvanRamyk (changes requested):
You haven't provide changes

PR #100, opened 2022-07-07
Solution

shurushku:


app/main.py
```diff
@@ -1,3 +1,13 @@
 def cache(func):
-    # Write your code here
-    pass
+    cache_base = {}
+
+    def inner(*args):
+        nonlocal cache_base
```
MasterpieceElbow, line 5: You should not use nonlocal if you work with the dict items.

MasterpieceElbow (changes requested):

MasterpieceElbow (approved):

PR #101, opened 2022-07-08
Solution

vladimir-byshevets:


app/main.py
```diff
@@ -1,3 +1,13 @@
 def cache(func):
-    # Write your code here
-    pass
+    saved = dict()
+
+    def function(*args):
+        key = str(args)
```
Abnormaltype, line 5: it's not correct, because if in args saved "True" it's a "True", and if saved True it's a "True"
True != "True"

Abnormaltype (changes requested):

Abnormaltype (approved):
well done

PR #106, opened 2022-07-18
Task Сompleted

Glasis9:


app/main.py
```diff
@@ -1,3 +1,10 @@
 def cache(func):
-    # Write your code here
-    pass
+    temp_dict = {}
+    def inner(*args):
+        if args in temp_dict:
+            print(f"Getting from cache")
```
IvanRamyk, line 5: Please, fix flake8 errors: string shouldn't be f-string if it doesn't contain placeholders

IvanRamyk (changes requested):

app/main.py
Glasis9: Done

IvanRamyk (approved):
Well done!

PR #109, opened 2022-07-22
Solution

Paul-Starodub:


app/main.py
```diff
@@ -1,3 +1,13 @@
 def cache(func):
-    # Write your code here
-    pass
+    my_dictionary = {}
```
IvanRamyk, line 2: Do not use the prefix `my_` in the variable names. It is meaningless

IvanRamyk (changes requested):

app/main.py
```diff
@@ -1,3 +1,13 @@
 def cache(func):
-    # Write your code here
-    pass
+    dictionary = {}
+
+    def wrapper(*args):
+        key = args
```
IvanRamyk, line 5: Extra variable

app/main.py
```diff
@@ -1,3 +1,13 @@
...
+        if key not in dictionary.keys():
```
IvanRamyk, line 6:
```suggestion
        if args not in dictionary.keys():
```

IvanRamyk (changes requested):

IvanRamyk (approved):
Well done!

PR #111, opened 2022-07-23
Solution

ansicat:


app/main.py
```diff
@@ -1,3 +1,18 @@
 def cache(func):
-    # Write your code here
-    pass
+    calls_cache = {}
+
+    def wrapper(*args):
+        nonlocal calls_cache
```
Abnormaltype, line 5: Nonlocal calls_cache - this statement is redundant, as dict is a mutable type;

app/main.py
```diff
@@ -1,3 +1,18 @@
...
+        call_key = func.__name__ + "_".join([str(item) for item in args])
```
Abnormaltype, line 6: it's incorrect, store args as key, and the result of the function as value

Abnormaltype (changes requested):

Abnormaltype (commented):
Good job!

Abnormaltype (approved):

Abnormaltype (commented):
it's not your solution, isn't it?

ansicat:
My second commit for local branch 'develop' has comment 'Solution #2' as text. So I mentioned about it in my question above.
Github parsed '#2' from it as a link for alien solution. Yes, it isn't mine.

My question was:
I shouldn't send 'pull request' after fixing mistakes in branch 'develop' and every command 'git push' except the first one, should I?

Abnormaltype (commented):
read this guideline https://github.com/mate-academy/py-task-guideline/blob/main/README.md#implement-the-solution

ansicat:
Got it. Thank you!

PR #112, opened 2022-08-02
Solution

KatyaVasylieva:


app/main.py
```diff
@@ -1,3 +1,15 @@
 def cache(func):
-    # Write your code here
-    pass
+    count = {}
+
+    def wrapper(*args):
+        nonlocal count
```
IvanRamyk, line 5: this statement is redundant, as dict is mutable type;

app/main.py
```diff
@@ -1,3 +1,15 @@
...
+
+        if args in count.keys():
```
```suggestion
        if args in count:
```
This statement could be simplified

IvanRamyk (changes requested):

IvanRamyk (approved):
Well done!

PR #114, opened 2022-08-03
Develop

dmitriyKostiuk:


app/main.py
```diff
@@ -1,3 +1,24 @@
...
+
+    @functools.wraps(func)
+    def wrapper(*args, **kwargs):
+        if (func.__name__, args) in global_cache:
+            print("Getting from cache")
+            return global_cache[(func.__name__, args)]
+        else:
+            print("Calculating new result")
+            # global_cache.append((func.__name__, args))
+        if func.__name__ == "long_time_func":
```
IvanRamyk, line 15: The decorator should work for any function

app/main.py
```diff
@@ -1,3 +1,24 @@
+import functools
+
+
 def cache(func):
-    # Write your code here
-    pass
+    global_cache = dict({})
```
IvanRamyk, line 5: The cache creates a unique for each function because it is placed inside the decorator.

app/main.py
```diff
@@ -1,3 +1,24 @@
...
+
+    @functools.wraps(func)
+    def wrapper(*args, **kwargs):
+        if (func.__name__, args) in global_cache:
```
```suggestion
        if args in global_cache:
```

app/main.py
IvanRamyk: To calculate function result just pass args to the function
```
func(*args)
```

IvanRamyk (changes requested):

app/main.py
```diff
@@ -1,3 +1,18 @@
...
+
+
 def cache(func):
-    # Write your code here
-    pass
+    global_cache = dict({})
+
+    @functools.wraps(func)
+    def wrapper(*args, **kwargs):
+        # if (func.__name__, args) in global_cache:
```
Abnormaltype, line 9: remove unnecessary comment

app/main.py
```diff
@@ -1,3 +1,18 @@
+import functools
+
+
 def cache(func):
-    # Write your code here
-    pass
+    global_cache = dict({})
```
Abnormaltype, line 5:
```suggestion
    global_cache = dict()
```

Abnormaltype (changes requested):

Abnormaltype (approved):
Well done!

PR #121, opened 2022-08-17
Solution

taras-andruschenko:


app/main.py
```diff
@@ -1,3 +1,15 @@
 def cache(func):
-    # Write your code here
-    pass
+    cache_dict = {}
+
+    def wrap_cache(*args):
+        nonlocal cache_dict
```
MasterpieceElbow, line 5: You can make it without `nonlocal`

MasterpieceElbow (changes requested):

app/main.py
```diff
@@ -1,3 +1,15 @@
 def cache(func):
-    # Write your code here
-    pass
+    cache_dict = {}
+
+    def wrap_cache(*args):
+        stored_result = cache_dict.get(args, False)
```
MasterpieceElbow, line 5: You don't need this variable. You use it only in `if` statement but you can simply use `cache_dict[args]`

app/main.py
```diff
@@ -1,3 +1,15 @@
...
+        if args in cache_dict:
+            print("Getting from cache")
+            return stored_result
+        else:
+            print("Calculating new result")
+            func_result = func(*args)
+            cache_dict[args] = func_result
+            return func_result
```
MasterpieceElbow, line 13: You can use a single return statement outside of the if/else statement instead of two. Just get value from the dict.

MasterpieceElbow (changes requested):

MasterpieceElbow (approved):
GJ!

PR #123, opened 2022-08-17
Solution

yessloss:


app/main.py
```diff
@@ -1,3 +1,15 @@
 def cache(func):
     # Write your code here
-    pass
+    dict_results = {}
+
+    def wrapper(*args):
+        nonlocal dict_results
```
DmytroSvirsa, line 6: this statement is redundant, as dict is mutable type;

DmytroSvirsa (changes requested):

app/main.py
```diff
@@ -1,3 +1,14 @@
 def cache(func):
     # Write your code here
```
IvanRamyk, line 2: Please, remove the redundant comment

app/main.py
```diff
@@ -1,3 +1,14 @@
...
-    pass
+    dict_results = {}
+
+    def wrapper(*args):
+        if args in dict_results:
+            print("Getting from cache")
+            return dict_results[args]
+        else:
```
IvanRamyk, line 9: Please, remove the redundant else

IvanRamyk (changes requested):

IvanRamyk (approved):
Good job!

PR #134, opened 2022-08-27
Solution for Cache decorator

Foggy-Cat:


app/main.py
```diff
@@ -1,3 +1,15 @@
 def cache(func):
     # Write your code here
```
IvanRamyk, line 2: Please, remove the redundant comment

app/main.py
```diff
@@ -1,3 +1,15 @@
...
-    pass
+    dict_for_cache = {}
+
+    def inner(*args):
+        key_name_dict = args
```
IvanRamyk, line 6: why don't use just `args`?

IvanRamyk (changes requested):

app/main.py
IvanRamyk: avoid using extra varaibles

app/main.py
```diff
@@ -1,3 +1,15 @@
...
-    pass
+
+    dict_for_cache = {}
+
+    def inner(*args):
+        key_name_dict = args
+        if key_name_dict in dict_for_cache:
+            print("Getting from cache")
+            return dict_for_cache[key_name_dict]
+        new_result = func(*args)
```
IvanRamyk, line 10: avoid using extra variable

IvanRamyk (changes requested):

IvanRamyk (approved):
Good job!

PR #136, opened 2022-08-30
create decorator cache, whom save result function to storage var, if …

LitvinOleksandr:
…function has equals inputs from another function - return result from cache storage var.

### DIFF
--- a/app/main.py
+++ b/app/main.py
@@ -1,3 +1,14 @@
 def cache(func):
-    # Write your code here
-    pass
+    storage = []
+
+    def run(*args):
+        nonlocal storage
+        for numb, val in enumerate(storage):
+            if storage[numb].get(args) is not None:
+                print("Getting from cache")
+                return storage[numb].get(args)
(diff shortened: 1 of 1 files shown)


app/main.py
```diff
@@ -1,3 +1,15 @@
...
-    pass
+    storage_result = {}
+
+    def wrapper(*args):
+
+        if storage_result.get(args) is not None:
+            print("Getting from cache")
+            return storage_result.get(args)
+
+        temp = func(*args)
```
IvanRamyk, line 10: avoid creating extra varaibles

app/main.py
```diff
@@ -1,3 +1,15 @@
...
+        storage_result.update({args: temp})
```
IvanRamyk, line 11:
```suggestion
        storage_result[args] = temp
```

IvanRamyk (changes requested):

IvanRamyk (approved):
Good job!

PR #142, opened 2022-09-04
Solution v.01

dkyrpan:


DmytroSvirsa (changes requested):
make it with only one return

DmytroSvirsa (approved):

PR #144, opened 2022-09-04
Solution

vlad-milkov:


DmytroSvirsa (changes requested):
It did more than needed, you will always get the same function, good job, but make it less complicated

vlad-milkov:
> It did more than needed, you will always get the same function, good job, but make it less complicated

Did you mean it will always work only with long_time_func and long_time_func2? No need to add the possibility of using other functions?

DmytroSvirsa (approved):
Perfect!

PR #147, opened 2022-09-06
Solution

artem-dubinin:


DmytroSvirsa (changes requested):
You dont need to store diferent function, it is already is done by enclosing.

PR #155, opened 2022-09-12
Solution

sbublinskyi:


app/main.py
```diff
@@ -1,3 +1,17 @@
 def cache(func):
-    # Write your code here
-    pass
+    saved = {}
+
+    def inner(*args):
+        func_with_args = (func, args)
```
MasterpieceElbow, line 5: You don't have to use a func object, dict `saved` is separate for each decorated function

MasterpieceElbow (changes requested):

MasterpieceElbow (approved):

PR #158, opened 2022-09-13
Develop

Mazurinka29:


app/main.py
```diff
@@ -1,3 +1,16 @@
 def cache(func):
-    # Write your code here
-    pass
+    cache_of_results = {}
+    cache_of_funcs = []
```
Abnormaltype, line 3: you should create only one store. Would be better dictionary

app/main.py
```diff
@@ -1,3 +1,16 @@
...
+
+    def wrapper(*args, **kwargs):
+        all_args = str(args) + str(**kwargs)
+        if func.__name__ in cache_of_funcs and all_args in cache_of_results:
```
Abnormaltype, line 7: you don't need to use the name of the function here. Store as a key - args, and as a value - result of the function

Abnormaltype (changes requested):
please fix it

Abnormaltype (approved):
Well done!

PR #160, opened 2022-09-14
Solution)

Gr1zzZzzly:


app/main.py
```diff
@@ -1,3 +1,15 @@
...
-    pass
+    catche = {}
+
+    def inner(*args):
+
+        if args in catche:  # check previous results
+            print("Getting from cache")
+        else:
+            # if result not in catch,
+            # add result in cathc and return catch
```
MasterpieceElbow, line 10: Here you have 3 different names: `cache`, `catche` and `catch`. What each of them means? :)

MasterpieceElbow (changes requested):

MasterpieceElbow (approved):
GJ!

PR #164, opened 2022-09-14
Solution for the task

olia-trofymchuk:


app/main.py
```diff
@@ -1,3 +1,15 @@
...
+    def inner(*args, **kwargs):
+
+        if args not in result:
+            res = func(*args, **kwargs)
+            result[args] = res
+            print("Calculating new result")
+            return res
+        else:
+            print("Getting from cache")
+            return result[args]
```
Nattalli, line 13: make one return for two cases , as they both returns result[args]

Nattalli (changes requested):

PR #167, opened 2022-09-14
Solution

misha-doroshenko:


app/main.py
```diff
@@ -1,3 +1,14 @@
...
+
+    def wrapper(*args):
+        if args in dictionary.keys():
+            print('Getting from cache')
+            return dictionary[args]
+        else:
+            res = func(*args)
+            dictionary[args] = res
+            print('Calculating new result')
+            return res
```
MasterpieceElbow, line 12: You can use a single `return` statement out of the `if/else` statement instead of two inside it.

MasterpieceElbow (changes requested):

app/main.py
MasterpieceElbow: Also, `result` is much better than `res`

TPolina (approved):

PR #174, opened 2022-09-15
'3rSolution'

ketstap162:


app/main.py
```diff
@@ -1,3 +1,12 @@
...
+    cache_lib = {}
+
+    def inner(*args):
+        if args not in cache_lib:
+            cache_lib[args] = func(*args)
+            print('Calculating new result')
+            return cache_lib[args]
+        else:
+            print('Getting from cache')
+            return cache_lib[args]
```
Nattalli, line 11: you have 2 the same returns. Use return at once in your function

Nattalli (changes requested):

Nattalli (approved):
well!

PR #179, opened 2022-09-16
Solution Vadim Sabadyr

VadimSabadyr:


app/main.py
```diff
@@ -1,3 +1,27 @@
...
+            print("Getting from cache")
+            return save_value[args]
+        res = func(*args, *kwargs)
+        save_value[args] = res
+        print("Calculating new result")
+        return res
+    return inner
+
+
+@cache
```
Nattalli, line 15: Is it your test functions? Remove all of them, please

app/main.py
```diff
@@ -1,3 +1,27 @@
...
+    save_value = {}
+
+    def inner(*args, **kwargs):
+        if args in save_value:
+            print("Getting from cache")
+            return save_value[args]
+        res = func(*args, *kwargs)
+        save_value[args] = res
+        print("Calculating new result")
+        return res
```
Nattalli, line 11: as I understand - you return a value `save_value[args]` in both cases, so you might to do 1 return inside the `inner` function (but you will have to use a construction `if-else`)
Your solution is good and works but you can make another function view so it will be more readable and easy for understanding

Nattalli (changes requested):

PR #180, opened 2022-09-16
Solution

LiudmylaKulinchenko:


app/main.py
```diff
@@ -1,3 +1,13 @@
...
+    old_results = {}
+
+    def inner(*args):
+        if args in old_results:
+            print("Getting from cache")
+            return old_results[args]
+        result = func(*args)
+        old_results[args] = result
+        print("Calculating new result")
+        return result
```
Nattalli, line 11: see: 
```python
old_results[args] = result
```
that means that if you make `return result` it will be equal to `return old_results[args]`. But in the first case, you have already used `return old_results[args]`
So you make 2 the same returns. Make only 1 return for 2 cases

Nattalli (changes requested):

TPolina (approved):
Well done!

PR #188, opened 2022-09-16
'Solution'

linaSam:


app/main.py
```diff
@@ -1,3 +1,16 @@
 def cache(func):
-    # Write your code here
-    pass
+    info = {}
+
+    def wrapper(*args):
+        nonlocal info
```
Nattalli, line 5: I think that you don't really need this import of nonlocal info

app/main.py
```diff
@@ -1,3 +1,16 @@
...
+        if args in info:
+            print('Getting from cache')
+            for name, inf in info.items():
```
Nattalli, line 8: I think that you can just `return info[args]` - it will be the same as 
```python
for name, inf in info.items():
        if name == args:
                    return inf
```

app/main.py
```diff
@@ -1,3 +1,16 @@
...
+                if name == args:
+                    return inf
+        else:
+            print('Calculating new result')
+            result = func(*args)
+            info[args] = result
+            return result
```
Nattalli, line 15: here you also make `return info[args]`. So you have the same return in both conditions. Then - make 1 return for both cases

Nattalli (changes requested):

app/main.py
```diff
@@ -1,3 +1,12 @@
 def cache(func):
-    # Write your code here
-    pass
+    info = {}
+
+    def wrapper(*args):
+        if args in info:
+            print('Getting from cache')
```
Y-Havryliv, line 6: use double quotes everywhere

Y-Havryliv (changes requested):

Y-Havryliv (approved):
GJ!

PR #189, opened 2022-09-16
Solution

YevheniiDavydenko:


app/main.py
```diff
@@ -1,3 +1,18 @@
...
+    @functools.wraps(func)
+    def wrapper(*args, **kwargs):
+        if args in cache_dict:
+            print("Getting from cache")
+            return cache_dict.get(args)
+        else:
+            result = func(*args)
+            cache_dict.update({args: result})
+            print("Calculating new result")
+            return result
```
Nattalli, line 16: you return the same value for both cases so - use 1 return for them

Nattalli (changes requested):

app/main.py
YevheniiDavydenko: I don't understand what you mean, first return its a value from dict, and second its a value from function

TPolina: Yes, you add a second value to the dict, so you can retrieve it from it. Nataliia means it would be better to make one `return` for both cases and take it outside the if-else statement

TPolina (approved):
Well done!

PR #192, opened 2022-09-17
Solution

pavlo-kapitanets:


app/main.py
```diff
@@ -1,3 +1,13 @@
...
+
+    def inner_function(*args):
+        if args in container:
+            print("Getting from cache")
+            return container[args]
+        else:
+            result_of_func = func(*args)
+            container[args] = result_of_func
+            print('Calculating new result')
+            return result_of_func
```
TPolina, line 12: You can use a single return statement out of the if/else statement instead of two inside it.

TPolina (changes requested):

TPolina (approved):

PR #194, opened 2022-09-18
Solution

AVyha:


app/main.py
```diff
@@ -1,3 +1,15 @@
...
+        if (args, power) in data:
+            print("Getting from cache")
+            return data[args, power]
+        else:
+            print("Calculating new result")
+            if power == 0:
+                data[args, power] = func(*args)
+            else:
+                data[args, power] = func(*args, power)
+            return data[args, power]
```
linaSam, line 14: you have the same return in line 7 . change your code so that it will have return only once

app/main.py
```diff
@@ -1,3 +1,15 @@
 def cache(func):
-    # Write your code here
-    pass
+    data = {}
+
+    def inner(*args, power: int = 0):
```
linaSam, line 4: variable "power" doesn't really do anything. try rewrite your code without it.

linaSam (changes requested):

app/main.py
AVyha: thank you, i wrote this variable for test myself this decorator and forget to clean

linaSam (changes requested):

app/main.py
linaSam: it's classic one) you are welcome!

app/main.py
```diff
@@ -1,15 +1,12 @@
 def cache(func):
     data = {}
 
-    def inner(*args, power: int = 0):
-        if (args, power) in data:
+    def inner(*args):
+        if args in data:
             print("Getting from cache")
-            return data[args, power]
+            return data[args]
```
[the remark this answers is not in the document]
AVyha: i hope in this time all be good) thak you for the review

app/main.py
```diff
@@ -4,9 +4,8 @@ def cache(func):
     def inner(*args):
         if args in data:
             print("Getting from cache")
-            return data[args]
         else:
             print("Calculating new result")
             data[args] = func(*args)
-            return data[args]
+        return data[args]
```
linaSam, line 10: yep, everything great now) nice work!

linaSam (approved):

TPolina (approved):
Well done!

PR #198, opened 2022-09-18
'Solution'

illiariaboy94:


Menselface (changes requested):
In that task we dont need to create long_time_func. You can just simple remove it.

Menselface (approved):

illiariaboy94:
> In that task we dont need to create long_time_func. You can just simple remove it.

Ok, I forgot to delete that.
Thanks

TPolina (approved):
Well done!

PR #203, opened 2022-09-23
Solution

Terrrya:


app/main.py
```diff
@@ -1,3 +1,13 @@
 def cache(func):
-    # Write your code here
-    pass
+    results = {}
+
+    def inner(*args):
+        nonlocal results
```
IvanRamyk, line 5: this statement is redundant, as dict is mutable type

IvanRamyk (changes requested):

app/main.py
Terrrya: I have a very old laptop. It didn't work without this line yesterday, but it works today. It's miracle or i don't know what is it:)

Abnormaltype (approved):
Well done!

PR #205, opened 2022-09-25
Solution

mish-sk:


app/main.py
```diff
@@ -1,3 +1,12 @@
...
+    cache_storage = {}
+
+    def inner(*args, **kwargs):
+        if args not in cache_storage:
+            cache_storage[args] = func(*args)
+            print("Calculating new result")
+            return cache_storage[args]
+        elif args in cache_storage:
+            print("Getting from cache")
+            return cache_storage[args]
```
DmytroSvirsa, line 11: good job, but let's make it with 1 return

DmytroSvirsa (changes requested):

PR #208, opened 2022-09-28
'Solution'

Paul-Maslov:


app/main.py
```diff
@@ -1,3 +1,19 @@
 def cache(func):
-    # Write your code here
-    pass
+
+    c = func.__name__
+    global cash_dict
```
DmytroSvirsa, line 4: are you sure that you need that?

app/main.py
```diff
@@ -1,3 +1,19 @@
...
+    cash_dict = {}
+
+    def inner(*args, **kwargs):
+
+        nonlocal c
+        result = str([item for item in args])
+        result += c
```
DmytroSvirsa, line 11: are you sure that you need that?

DmytroSvirsa (changes requested):

app/main.py
Paul-Maslov: Yes, since the dictionary must be stored as a global variable, which can be accessed later by the following functions

app/main.py
```diff
@@ -1,3 +1,17 @@
 def cache(func):
-    # Write your code here
-    pass
+
+    c = func.__name__
```
DmytroSvirsa, line 3: we can do without it (check the slack chat for mo info)

DmytroSvirsa (changes requested):

Abnormaltype (approved):
Well done!

PR #213, opened 2022-09-29
Solution

marina-bidenko:


app/main.py
```diff
@@ -1,3 +1,14 @@
 def cache(func):
-    # Write your code here
-    pass
+    dt = {}
+
+    def wrapper(*args):
+        if not dt.get(args, False) is False:
```
MasterpieceElbow, line 5: Bad style. You can simply check the key with the `in` keyword.

MasterpieceElbow (changes requested):

PR #214, opened 2022-09-29
Solution

SpecializedBaby:
I can't save logs in global list? Please help me. Where or how I can save it?

### DIFF
--- a/app/main.py
+++ b/app/main.py
@@ -1,12 +1,38 @@
-caches = []
-
-
 def cache(func):
-    def inner(*args, **kwargs):
-        global caches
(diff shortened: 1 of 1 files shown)


DmytroSvirsa (changes requested):
some tests had not pass, the problem in caches variable, You neen it but in another plase and another type, also pay atention you did not return nothing from inner function

SpecializedBaby:
I created local variable in decorator Cache and these conditions confuse(if) me too much, but I can’t figure out how to dynamically pass parameters to the main function.

PR #215, opened 2022-10-01
Develop

ihor-pylat:


app/main.py
```diff
@@ -1,3 +1,28 @@
 def cache(func):
-    # Write your code here
```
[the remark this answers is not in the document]
Abnormaltype: you already have deleted it) If the code has a red background, it means that the code was deleted

app/main.py
```diff
@@ -1,3 +1,28 @@
...
+            cached_list.append(element)
+
+        else:
+            for value in element:
+
+                return value
+
+        return func(*args)
+    return inner
     pass
```
Abnormaltype, line 28: remove unnecessary pass

app/main.py
```diff
@@ -1,3 +1,28 @@
 def cache(func):
-    # Write your code here
+    cached_list = []
+
+    def inner(*args):
+
+        access = True
```
Abnormaltype, line 6: unnecessary variable

app/main.py
```diff
@@ -1,3 +1,28 @@
...
+            print("Calculating new result")
+            cached_list.append(element)
+
+        else:
+            for value in element:
+
+                return value
+
+        return func(*args)
+    return inner
```
Abnormaltype, line 27: you can write this in two conditions

Abnormaltype (changes requested):

app/main.py
```diff
@@ -1,3 +1,21 @@
 def cache(func):
-    # Write your code here
-    pass
+
+    cached_list = []
```
Abnormaltype, line 3: sue dictionary as cache, and store args as key, and result of the func as value

Abnormaltype (changes requested):

app/main.py
```diff
@@ -1,3 +1,20 @@
...
+                cached_list[args] = "Getting from cache"
+                print(cached_list[args])
+
+                for item in args:
+                    return item
+
+        cached_list[args] = "Calculating new result"
+        print(cached_list[args])
+
+        return func(*args)
```
Abnormaltype, line 19: you should return the result from the cache

Abnormaltype (changes requested):

app/main.py
ihor-pylat: When I try to return the result from the cache I get an error ( function missing 2 required positional arguments: 'b' and 'c'). How do I fix this?

app/main.py
```diff
@@ -1,3 +1,20 @@
 def cache(func):
-    # Write your code here
-    pass
+
+    cached_list = {}
+
+    def inner(*args):
+
+        for key, value in cached_list.items():
+            if key == args:
```
Abnormaltype, line 8: just check if args in the cache

app/main.py
```diff
@@ -1,3 +1,20 @@
...
+
+                cached_list[args] = "Getting from cache"
```
Abnormaltype, line 10: you should print this string, not store in the cache

app/main.py
Abnormaltype, line 19: you should store the result.

Abnormaltype (changes requested):

ihor-pylat:
Why I have a fail (Error: Process completed with exit code 1)?  In my PyCharm everything is OK

ihor-pylat:
What is the annotation type for a function argument?
Why are there errors in the tests that I passed before?

IvanRamyk (changes requested):
use `Any` as type annotation for functions. 
If you have more questions ask them in slack

app/main.py
```diff
@@ -1,3 +1,17 @@
...
+    def inner(*args) -> any:
+        if args in cached_list:
+
+            cached_list[args] = "Getting from cache"
+            print("Getting from cache")
+
+            for item in args:
+                return item
+
+        cached_list[args] = "Calculating new result"
```
IvanRamyk, line 13: you should store the result of the function not a message

app/main.py
```diff
@@ -1,3 +1,17 @@
...
+def cache(func: any) -> any:
+    cached_list = {}
+
+    def inner(*args) -> any:
+        if args in cached_list:
+
+            cached_list[args] = "Getting from cache"
+            print("Getting from cache")
+
+            for item in args:
```
IvanRamyk, line 10: why do you iterate over args? 
you should return a result, not an argument

app/main.py
```diff
@@ -1,3 +1,17 @@
...
+                return item
+
+        cached_list[args] = "Calculating new result"
+        print("Calculating new result")
+
+        return func(*args)
```
IvanRamyk, line 16: save result of execution

IvanRamyk (changes requested):

app/main.py
ihor-pylat: When I trying return a result I have a error:   FAILED tests/test_main.py::test_cache_returns_correct_value - assert (2, 3, 3) == 2.  But when I return a argument all is passed. Why?

app/main.py
ihor-pylat: You wrote earlier: sue dictionary as cache, and store args as key, and result of the func as value. I store result and message. Its not correct?

app/main.py
ihor-pylat: I need save "func(*args)" before I return this?

app/main.py
DmytroSvirsa: no store the args as key, and result of function as value

app/main.py
DmytroSvirsa: you need func(*args) if there is no stored result in your dictionary with execution results

DmytroSvirsa (changes requested):

PR #216, opened 2022-10-02
Develop

T-Vlad-96:


app/main.py
```diff
@@ -1,3 +1,23 @@
...
+
+
 def cache(func):
-    # Write your code here
-    pass
+    saved_results = {}
+
+    @functools.wraps(func)
+    def wrapper(*args, **kwargs):
+        if func.__name__ in saved_results:
```
Abnormaltype, line 9: you can write this without func name, store args as key, and a result of the func as value

app/main.py
```diff
@@ -1,3 +1,23 @@
...
+                return saved_results[func.__name__][str(args)]
+            else:
+                print("Calculating new result")
+                saved_results[func.__name__].\
+                    update({str(args): func(*args, **kwargs)})
+                return saved_results[func.__name__][str(args)]
+        else:
+            print("Calculating new result")
+            saved_results[func.__name__] = {str(args): func(*args, **kwargs)}
+            return saved_results[func.__name__][str(args)]
```
Abnormaltype, line 21: remove code duplication

Abnormaltype (changes requested):

app/main.py
T-Vlad-96: hello, sorry for delay, i have proceed my learning, started new topic and accidentally forgot about this task. Could you please explain specifically how to avoid using func.__name__. The cache decorator won't work with different functions if i not store the func_name as a key.
ps: The issue about code duplication is completely clear. It's fixed

Abnormaltype: better write in slack)

DmytroSvirsa (changes requested):
Have any results?

T-Vlad-96:
> Have any results?

yes, the changed was applied

app/main.py
```diff
@@ -1,3 +1,16 @@
-def cache(func):
-    # Write your code here
-    pass
+from typing import Union
+
+
+def cache(func: callable) -> callable:
```
Abnormaltype, line 4:
```suggestion
def cache(func: Сallable) -> Сallable:
```

app/main.py
```diff
@@ -1,3 +1,16 @@
...
+    saved_results = {}
+
+    def wrapper(*args: Union[int, float, str, tuple, frozenset],
+                **kwargs: Union[int, float, str, tuple, frozenset]) -> dict:
```
Abnormaltype, line 8:
```suggestion
    def wrapper(*args, **kwargs) -> dict:
```

Abnormaltype (changes requested):

app/main.py
T-Vlad-96: Can't apply the changes. When i use "callable" with capital "C" i get "NameError: name 'Callable' is not defined. Did you mean: 'callable'?"

Abnormaltype: you should import it
`from typing import Callable`

Abnormaltype (changes requested):

app/main.py
T-Vlad-96: got it, thanks

Abnormaltype (approved):
Well done!

PR #221, opened 2022-10-04

YLobunets:


app/main.py
```diff
@@ -1,3 +1,23 @@
+from functools import wraps
+
+
 def cache(func):
-    # Write your code here
-    pass
+    """New decoration function with cache"""
+    cache_keys = []
+    cache_result = []
+
```
Abnormaltype, line 8: use a dictionary to cache your results. Key as args, and value as result of the function

Abnormaltype (changes requested):

app/main.py
YLobunets: In this case i have to use for to check keys, since it is not important to understand whether the value of the key is in the dictionary.
In general, functions can return any value, including None.
And if each time you convert the dictionary into an array of keys, then this will be a complete analogue of the solution with two arrays.

Change to dictionary?

Abnormaltype (approved):
Well done!

PR #224, opened 2022-10-05
Add Cache decorator

LavrykN:


Abnormaltype (changes requested):
use `""` instead of `''`

Abnormaltype (approved):
Well done!

PR #231, opened 2022-10-13
Solution

Thirteenthskyi:


app/main.py
```diff
@@ -1,3 +1,14 @@
-def cache(func):
-    # Write your code here
-    pass
+def cache(func: object) -> object:
+    cache = {}
```
IvanRamyk, line 2: the cache instance is unique for each function because it is created during decorating the corresponding function

So you can get rid of function name in `signature ` and just use `args`

IvanRamyk (changes requested):

app/main.py
```diff
@@ -1,3 +1,14 @@
...
+
+    def wrapper(*args) -> str:
+        signature = (args)
```
IvanRamyk, line 5: an extra variable
just use `args`

IvanRamyk (changes requested):

app/main.py
```diff
@@ -1,3 +1,13 @@
-def cache(func):
-    # Write your code here
-    pass
+def cache(func: object) -> object:
```
Abnormaltype, line 1:
```suggestion
def cache(func: Callable) -> Callable:
```

app/main.py
```diff
@@ -1,3 +1,13 @@
...
+    cache = {}
+
+    def wrapper(*args) -> str:
```
Abnormaltype, line 4:
```suggestion
    def wrapper(*args) -> str:
```
yu should return not a string

Abnormaltype (changes requested):

app/main.py
```diff
@@ -1,3 +1,13 @@
-def cache(func):
-    # Write your code here
-    pass
+def cache(func: any) -> any:
```
Abnormaltype, line 1:
```suggestion
def cache(func: Callable) -> Callable:
```

app/main.py
```diff
@@ -1,3 +1,13 @@
...
+    cache = {}
+
+    def wrapper(*args) -> func:
```
Abnormaltype, line 4:
```suggestion
    def wrapper(*args) -> Callable:
```

Abnormaltype (changes requested):

IvanRamyk (approved):

PR #238, opened 2022-10-21
make cache decorator

Foksanv:


app/main.py
```diff
@@ -1,3 +1,17 @@
...
+
+
+def cache(func: Callable) -> Callable:
+    store = {}
+
+    def wrapper(*args, **kwargs) -> None:
+        if args not in store:
+            print("Calculating new result")
+            immut = (float, int, tuple, str)
+            if all([isinstance(arg, immut) for arg in args]):
```
YuriiKindrat, line 11: no need to make this check and immut variable here

PR #240, opened 2022-10-21
Solution

evgenijnozdrenko:


app/main.py
```diff
@@ -1,3 +1,13 @@
-def cache(func):
-    # Write your code here
-    pass
+def cache(func: callable) -> callable:
+    dict_res = {}
+
+    def inner(*args) -> callable:
```
Abnormaltype, line 4: use `Callable from typing`

app/main.py
```diff
@@ -1,3 +1,13 @@
...
+        if args not in dict_res:
+            res = func(*args)
```
Abnormaltype, line 6: res is a bad name for variable

app/main.py
```diff
@@ -1,3 +1,13 @@
...
+            dict_res[args] = res
+            print("Calculating new result")
+            return dict_res[args]
+        else:
+            print("Getting from cache")
+            return dict_res[args]
```
Abnormaltype, line 12: you can write this with one `return` statement

Abnormaltype (changes requested):

DmytroSvirsa (approved):

PR #248, opened 2022-10-28
'Solution'

Jorjio22:


app/main.py
```diff
@@ -1,3 +1,17 @@
...
-    # Write your code here
-    pass
+from typing import Callable
+
+
+def cache(func: Callable) -> Callable:
+    func_arg_res = {}
+
+    def inner(*args, **kwargs) -> Callable:
+        if f"{args},{kwargs}" not in func_arg_res:
```
Abnormaltype, line 8: use args as key, and value as a result of the function

Abnormaltype (changes requested):

Serhii-Leonenko (approved):
Good job!

PR #249, opened 2022-10-30
done, all works

phaishuk:


app/main.py
```diff
@@ -1,3 +1,13 @@
...
+
+    def inner(*args) -> callable:
+        nonlocal cache_lib
+        if args not in cache_lib:
+            cache_lib[args] = func(*args)
+            print("Calculating new result")
+            return cache_lib[args]
+        if args in cache_lib:
+            print("Getting from cache")
+            return cache_lib[args]
```
IvanRamyk, line 12: duplicated line 9

app/main.py
```diff
@@ -1,3 +1,13 @@
-def cache(func):
-    # Write your code here
-    pass
+def cache(func: callable) -> callable:
+    cache_lib = {}
+
+    def inner(*args) -> callable:
+        nonlocal cache_lib
```
IvanRamyk, line 5: this statement is redundant, as dict is a mutable type

app/main.py
```diff
@@ -1,3 +1,13 @@
-def cache(func):
-    # Write your code here
-    pass
+def cache(func: callable) -> callable:
+    cache_lib = {}
```
IvanRamyk, line 2: why the name of the variable contains `lib`?

IvanRamyk (changes requested):

app/main.py
phaishuk: I did that because I use a variable in a different scope. I didn't know that it's not necessary to use `nonlocal` in mutable types

app/main.py
phaishuk: because it looks like a library.. That's why I named it in this way. Is it wrong?

app/main.py
phaishuk: redone

phaishuk:
[changes made](https://github.com/mate-academy/py-cache-decorator/pull/249/commits/c8fa84c614fd8526eea311609f13f1bbd9ff9fae)

app/main.py
```diff
@@ -1,3 +1,11 @@
...
-    # Write your code here
-    pass
+def cache(func: callable) -> callable:
+    cache_storage = {}
+
+    def inner(*args) -> callable:
+        if args not in cache_storage:
+            cache_storage[args] = func(*args)
+            print("Calculating new result")
+        elif args in cache_storage:
```
Serhii-Leonenko, line 8: You could use just `else`

Serhii-Leonenko (approved):

app/main.py
phaishuk: Thanks!

PR #255, opened 2022-11-03
Solution

andriy-md:


app/main.py
```diff
@@ -1,3 +1,18 @@
-def cache(func):
-    # Write your code here
-    pass
+def cache(func: callable) -> callable:
```
IvanRamyk, line 1: replace `callable` with `Callable` from `typing`

app/main.py
```diff
@@ -1,3 +1,18 @@
...
+    # dictionary 'archive' will store arguments given to func (as keys)
```
IvanRamyk, line 2: it's not a good practice to write comments for your code a lot. code should be self-descriptive

IvanRamyk: And actually, it is self-descriptive in this function

IvanRamyk: you can use docstring to explain what the function does

app/main.py
```diff
@@ -1,3 +1,18 @@
...
+    # and func(args) as values
+    archive = {}
+
+    def wrapper(*args: tuple) -> callable:
+        # if function has already been called with these argument,
+        # we seek in archive
+        if args in archive.keys():
```
```suggestion
        if args in archive:
```

IvanRamyk (changes requested):

IvanRamyk (approved):
Well done!

PR #256, opened 2022-11-04
Solution

viktory-koroliova:


app/main.py
```diff
@@ -1,3 +1,16 @@
-def cache(func):
-    # Write your code here
-    pass
+from typing import Callable
+
+
+def cache(func: Callable) -> Callable:
+    cached_values = {}
+
+    def inner(*args) -> Callable:
```
```suggestion
    def inner(*args) -> Any:
```
inner could return anything that is returned from `func`

IvanRamyk (changes requested):

Serhii-Leonenko (approved):
Good job!

PR #258, opened 2022-11-10
Solution

anastasiia-tsurkan:


app/main.py
```diff
@@ -1,3 +1,21 @@
...
+        if f"{func.__name__}" not in store.keys():
+            store.update({f"{func.__name__}": {args: func(*args)}})
+            print("Calculating new result")
+        else:
+            if args not in store[f"{func.__name__}"]:
+                store[f"{func.__name__}"].update({args: func(*args)})
+                print("Calculating new result")
+            else:
+                print("Getting from cache")
+        return store[f"{func.__name__}"].get(args)
```
DmytroSvirsa, line 19: this can be done in more simple way
you don`t need to store the function

DmytroSvirsa (changes requested):

app/main.py
anastasiia-tsurkan: Please, check changes

DmytroSvirsa (approved):

PR #275, opened 2022-11-23
'Solution'

LeoT26:


app/main.py
```diff
@@ -1,3 +1,16 @@
...
+    cache_data = {}
+
+    def memory(*args: Callable) -> Any:
+        if args in cache_data:
+            print("Getting from cache")
+            return cache_data[args]
+        print("Calculating new result")
+        func_result = func(*args)
+        cache_data[args] = func_result
+        return func_result
```
Abnormaltype, line 14: you can write this with one return

Abnormaltype (changes requested):

app/main.py
```diff
@@ -1,3 +1,17 @@
...
+from typing import Callable, Any
+
+
+def cache(func: Callable) -> Any:
+    cache_data = {}
+
+    def memory(*args: Callable) -> Any:
+        if args in cache_data:
+            print("Getting from cache")
+        else:
```
OleksandrPozden, line 10: remove unnecessary operator

OleksandrPozden (changes requested):

OleksandrPozden (changes requested):
The file is deleted. Try to checkout to the previous commit and start from that point.

PR #278, opened 2022-11-25
'Solution'

Andrew13X:


app/main.py
```diff
@@ -1,3 +1,17 @@
...
+def cache(func: Callable) -> Callable:
+    result = {}
+
+    def wrapper(*args: Callable) -> Callable:
+        if args not in result:
+            func_result = func(*args)
+            result.update({args: func_result})
+            print("Calculating new result")
+            return func_result
+        else:
```
OleksandrPozden, line 13: Unnecessary ```else```

OleksandrPozden (changes requested):

Abnormaltype (changes requested):
you can write this with one return

Andrew13X:
Did it with one return. I can not make the code working without "else"

IvanRamyk (approved):

PR #279, opened 2022-11-26
Solution

alina-boichenko:


app/main.py
```diff
@@ -1,3 +1,16 @@
...
+from typing import Callable, Any
+
+
+def cache(func: Callable) -> Callable:
+    stores = {}
+
+    def inner(*args: Any) -> Any:
+        if args in stores:
+            print("Getting from cache")
+        else:
```
OleksandrPozden, line 10: without ```else``` , please

OleksandrPozden (changes requested):

app/main.py
alina-boichenko: I don't understand how to write without else

PR #281, opened 2022-11-28
Solution

AnatoliyPilipey:


app/main.py
```diff
@@ -1,3 +1,27 @@
...
-    # Write your code here
-    pass
+from typing import Callable
+
+
+def cache(func: Callable) -> Callable:
+    memory_cache = {}
+
+    def inner(*args, **kwargs) -> list and int:
+        nonlocal memory_cache
```
YuriiKindrat, line 8: No need to use nonlocal here

app/main.py
```diff
@@ -1,3 +1,27 @@
...
+        immute = (dict, set, list)
+        for argumets in args:
+            if type(argumets) in immute:
+                return
```
YuriiKindrat, line 12: U don't need to check the args type 
image

app/main.py
```diff
@@ -1,3 +1,27 @@
...
+
+        memory_cache_key = ""
+        input_num = args[0] if type(args[0]) is tuple else args
+        for each in input_num:
+            memory_cache_key += str(each) + "$"
+        memory_cache_key += str(args[1])
```
YuriiKindrat, line 18: No need to create some memory cache key, u can use args as a key for ur memory_cache dict, cause the args always immutable

YuriiKindrat (changes requested):

AnatoliyPilipey:
In the last code, memory_cache_key is encoding the incoming numbers into a string, and later I put this code as keys in a dict and check for matches.
The decorator receives a tuple as well as a nested tuple. If I find a nested tuple I encode it and add a number going 2 positions at index 1 from the incoming tuple.
If there is no nested tuple, adding a chilo to 2 positions will not lose the uniqueness of the code for the incoming numbers, they will just increase.

PR #284, opened 2022-12-06
Solution

olenaliuby:


app/main.py
```diff
@@ -1,3 +1,16 @@
...
+
+    def wrapper(*args) -> Any:
+        if args in stored_cache:
+            print("Getting from cache")
+            return stored_cache[args]
+        else:
+            print("Calculating new result")
+            result = func(*args)
+            stored_cache[args] = result
+            return result
```
YuriiKindrat, line 15: U can use only one return here and return stored_cache[args]

YuriiKindrat (changes requested):

app/main.py
olenaliuby: Made changes.

Also shortened:
`stored_cache[args] = func(*args)`

and changed annotation for wrapper:
`def wrapper(*args) -> dict:`

Accidentally deleted my last commit.

olenaliuby: Made changes.
Can you look up if it's ok?
Thank you for the review!

PR #285, opened 2022-12-07
Solution

TarasSavchyn:


app/main.py
```diff
@@ -1,3 +1,16 @@
...
-    pass
+from typing import Callable, Any
+cache_archive = {}
+
+
+def cache(func: Callable) -> Callable:
+    def wrapper(*args: Any, **kwargs: Any) -> Any:
+        if f"{func} {args} {kwargs}" not in cache_archive.keys():
+            print("Calculating new result")
+            res = func(*args, **kwargs)
```
katevasilenko, line 9: You don’t need this variable res. You return similar data in line 14 without it. And actually, you can return this only one time, now you repeat yourself

katevasilenko (changes requested):

app/main.py
TarasSavchyn: reworked, tried to make it easier to read

app/main.py
```diff
@@ -1,3 +1,19 @@
-def cache(func):
-    # Write your code here
-    pass
+from typing import Callable, Any
+
+cache_archive = {}
```
IvanRamyk, line 3: move it into the decorator. then for every decorated function will be created a new dictionary to store it results

app/main.py
```diff
@@ -1,3 +1,19 @@
...
+
+
+def cache(func: Callable) -> Callable:
+    def wrapper(*args: tuple, **kwargs: dict) -> Any:
+        save_func_done = f"{func} {args} {kwargs}"
```
IvanRamyk, line 8: if you consider the advise above then you don't need to create this string and just use args as a key

app/main.py
```diff
@@ -1,3 +1,19 @@
-def cache(func):
-    # Write your code here
-    pass
+from typing import Callable, Any
+
+cache_archive = {}
+
+
+def cache(func: Callable) -> Callable:
+    def wrapper(*args: tuple, **kwargs: dict) -> Any:
```
```suggestion
    def wrapper(*args: tuple) -> Any:
```

IvanRamyk: in this task, presume that function accepts only positional arguments

IvanRamyk (changes requested):

app/main.py
```diff
@@ -1,3 +1,19 @@
...
+
+        if args in cache_archive.keys():
+            print("Getting from cache")
+            res = cache_archive[args]
+        else:
+            print("Calculating new result")
+            res = func(*args)
+            cache_archive[args] = res
+
+        return res
```
Abnormaltype, line 17:
```suggestion
        if args in cache_archive.keys():
            print("Getting from cache")
        print("Calculating new result")
        result = func(*args)
        cache_archive[args] = result

        return result
```

Abnormaltype (changes requested):

app/main.py
Abnormaltype, line 17: make it as result, and write with one return

Abnormaltype (changes requested):

app/main.py
TarasSavchyn: > make it as result, and write with one return
remame variable to result)

katevasilenko (approved):

PR #290, opened 2022-12-09
Solution

serhei-pasov:


app/main.py
```diff
@@ -1,3 +1,19 @@
...
+            func_list[args] = func(*args)
+            func_result = func_list.get(args)
+            print("Calculating new result")
+
+            return func_result
+
+        else:
+            print("Getting from cache")
+
+            return func_list.get(args)
```
OleksandrPozden, line 18: remake this without ```else```. Also, remove space between the lines

OleksandrPozden (changes requested):

PR #292, opened 2022-12-10
Dev

drapail:


app/main.py
```diff
@@ -1,3 +1,14 @@
...
+        previous_function_calls[func.__name__] = {}
+
+    def wrapper(*args) -> int:
+        if args not in previous_function_calls[func.__name__]:
+            print("Calculating new result")
+            previous_function_calls[func.__name__][args] = func(*args)
+            return previous_function_calls[func.__name__][args]
+        else:
+            print("Getting from cache")
+            return previous_function_calls[func.__name__][args]
```
katevasilenko, line 13: You repeated yourself here. You can return only one time in function

katevasilenko (changes requested):

app/main.py
```diff
@@ -1,3 +1,16 @@
-def cache(func):
-    # Write your code here
-    pass
+from typing import Any
+
+
+def cache(func: callable) -> callable:
+    previous_function_calls = {}
+    if func.__name__ not in previous_function_calls.keys():
+        previous_function_calls[func.__name__] = {}
```
Abnormaltype, line 7: you can use args - as key, and use value - as a result of the function

Abnormaltype (changes requested):

Vasyl-Smutok (approved):
`previous_function_calls.keys() equal previous_function_calls`
so it is not necessary to use keys()

PR #295, opened 2022-12-12
decorator solution

AnnaKabatova:


YuriiKindrat (changes requested):

app/main.py
```diff
@@ -1,3 +1,14 @@
-def cache(func):
-    # Write your code here
-    pass
+from typing import Callable
+
+
+def cache(func: Callable) -> Callable:
+    cache_dict = {}
+
+    def inner(*args) -> None:
```
```suggestion
    def inner(*args) -> Any:
```
inner could return random results of `func`

IvanRamyk: Any could be imported from typing

IvanRamyk (changes requested):

YuriiKindrat (approved):

PR #296, opened 2022-12-15
Solution

vasyl-main-dev:
Sorry)

### DIFF
--- a/app/main.py
+++ b/app/main.py
@@ -2,15 +2,15 @@ from typing import Callable, Any
 
 
 def cache(func: Callable) -> Callable:
-    cash = {}
+    data = {}
 
     def init(*args) -> Any:
-        if args in cash:
+        if args in data:
             print("Getting from cache")
-            return cash[args]
+            return data[args]
         else:
             print("Calculating new result")
             res = func(*args)
-            cash[args] = res
+            data[args] = res
             return res
     return init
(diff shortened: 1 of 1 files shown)


app/main.py
```diff
@@ -1,3 +1,16 @@
-def cache(func):
-    # Write your code here
-    pass
+from typing import Callable, Any
+
+
+def cache(func: Callable) -> Callable:
+    cash = {}
```
IvanRamyk, line 5:
```suggestion
    cash = {}
```
it means 'готівка' :)

IvanRamyk: please choose another name

IvanRamyk (changes requested):

app/main.py
```diff
@@ -1,3 +1,16 @@
...
+
+    def init(*args) -> Any:
+        if args in data:
+            print("Getting from cache")
+            return data[args]
+        else:
+            print("Calculating new result")
+            res = func(*args)
+            data[args] = res
+            return res
```
OleksandrPozden, line 15:
```suggestion
        if args in data:
            print("Getting from cache")
            return data[args]
        print("Calculating new result")
        res = func(*args)
        data[args] = res
            return res
```

OleksandrPozden (changes requested):

OleksandrPozden (approved):

PR #297, opened 2022-12-15
'Solution'

Vitalii2906:


app/main.py
```diff
@@ -1,3 +1,19 @@
-def cache(func):
-    # Write your code here
-    pass
+def cache(func: callable) -> callable:
+    my_cache = dict()
+
+    def wrapped_func(*args, **kwargs) -> callable:
+        key = str(args) + str(kwargs)
```
YuriiKindrat, line 5: No need to make such thing, u can use just args as a key for ur cache

app/main.py
```diff
@@ -1,3 +1,19 @@
...
+        if key in my_cache:
+            print("Getting from cache")
+            return my_cache[key]
+        value = func(*args, **kwargs)
+        my_cache[key] = value
+        print("Calculating new result")
+        return value
```
YuriiKindrat, line 12: U can make it in one reutrn using my_cache[key]

app/main.py
```diff
@@ -1,3 +1,19 @@
...
+
+    return wrapped_func
+
+
+@cache
+def long_time_func_2(n_tuple: tuple, power: int) -> list:
+    return [number ** power for number in n_tuple]
```
YuriiKindrat, line 19: No need to write this func here

YuriiKindrat (changes requested):

app/main.py
```diff
@@ -1,3 +1,13 @@
...
+    my_cache = dict()
+
+    def wrapped_func(*args, **kwargs) -> callable:
+        key = str(args)
+        if key in my_cache:
+            print("Getting from cache")
+            return my_cache[key]
+        my_cache[key] = func(*args, **kwargs)
+        print("Calculating new result")
+        return my_cache[key]
```
OleksandrPozden, line 11: don't repeat yourself. you can do this differently

OleksandrPozden (approved):

OleksandrPozden (approved):

PR #299, opened 2022-12-19
Solution

IhorPokr:


app/main.py
```diff
@@ -1,3 +1,17 @@
...
+
+    def wrapper(*args, **kwargs) -> Any:
+        if args in cache:
+            print("Getting from cache")
+            return cache[args]
+        else:
+            print("Calculating new result")
+            result = func(*args, **kwargs)
+            cache[args] = result
+            return result
```
YuriiKindrat, line 15: U can make one return here for both cases

YuriiKindrat (changes requested):

YuriiKindrat (approved):

PR #304, opened 2022-12-22
Solution

OleksandrGrk:


katevasilenko (changes requested):
Actually you need only `cache` function here

app/main.py
```diff
@@ -1,3 +1,16 @@
-def cache(func):
-    # Write your code here
-    pass
+from typing import Callable, Any
+
+
+def cache(func: Callable) -> Any:
+    all_number = []
```
IvanRamyk, line 5: extra list, try to check if args are already present in result dict

app/main.py
```diff
@@ -1,3 +1,16 @@
...
+    result = {}
+
+    def check_data(*args: int) -> Callable:
+        if args in all_number:
+            print("Getting from cache")
+        else:
+            result[f"{func.__name__}{args}"] = func(*args)
```
IvanRamyk, line 12:
```suggestion
            result[args] = func(*args)
```
results will be created separately for every decorated function because it's declared inside the decorator. so concatenating function name and args is redundant

IvanRamyk (changes requested):

PR #306, opened 2022-12-23
Solution

YegorT:


app/main.py
```diff
@@ -1,3 +1,18 @@
...
+    @functools.wraps(func)
+    def wrapper(*args) -> Any:
+        if (func.__name__, args) in cache_store:
+            print("Getting from cache")
+            return cache_store[(func.__name__, args)]
+        else:
+            result = func(*args)
+            cache_store[(func.__name__, args)] = result
+            print("Calculating new result")
+            return result
```
YuriiKindrat, line 17: u can use only one return here

YuriiKindrat (changes requested):

app/main.py
```diff
@@ -1,3 +1,18 @@
...
+
+
+def cache(func: Callable) -> Callable:
+    cache_store = {}
+
+    @functools.wraps(func)
+    def wrapper(*args) -> Any:
+        if (func.__name__, args) in cache_store:
+            print("Getting from cache")
+            result = cache_store[(func.__name__, args)]
```
ArsenPidhoretskyi, line 12: As for me, it is redundant.
You can do return cache_store[(func.__name__, args)]
Also, why do you need to save the func name? 
image

ArsenPidhoretskyi: image

ArsenPidhoretskyi (changes requested):

app/main.py
```diff
@@ -1,3 +1,14 @@
-def cache(func):
-    # Write your code here
-    pass
+from typing import Callable, Any
+
+
+def cache(func: Callable) -> Callable:
+    storage = {}
+
+    def wrapper(*args) -> Any:
```
olena-hul, line 7: And annotation for args

olena-hul (changes requested):

ArsenPidhoretskyi (approved):
Well done!

PR #310, opened 2022-12-26
Solution

avkpol:


README.md
```diff
@@ -20,6 +20,20 @@ Also `cache` should print `Getting from cache` when returns stored value and
...
+def cache(func: Callable) -> Callable:
+    def wrapper(*args, **kwargs): -> list
+        already_run = []
+        for i in args:
+            if
+            
+    
+
+
+
```
katevasilenko, line 36: Please don’t fix readme file

app/main.py
```diff
@@ -1,3 +1,19 @@
...
-    pass
+from typing import Callable
+from functools import wraps
+
+
+def cache(func: Callable) -> Callable:
+    storage = {}  # this will be stored in closure because it is nonlocal
+
+    @wraps(func)
+    def wrapper(*args, **kwargs) -> list:
```
katevasilenko, line 9: You can use only rags in this exact task

app/main.py
```diff
@@ -1,3 +1,19 @@
...
+        # Unpack args and kwargs into a tuple to be used as dict keys
+        keys = (tuple(args) + tuple(kwargs.keys()))
```
katevasilenko, line 11: Seems like unnecessary

app/main.py
```diff
@@ -1,3 +1,19 @@
...
+        if keys not in storage:
+            print("Calculating new result")
+            storage[keys] = func(*args, **kwargs)
```
katevasilenko, line 14: Remove kwargs here too

app/main.py
```diff
@@ -1,3 +1,19 @@
-def cache(func):
-    # Write your code here
-    pass
+from typing import Callable
+from functools import wraps
+
+
+def cache(func: Callable) -> Callable:
+    storage = {}  # this will be stored in closure because it is nonlocal
```
katevasilenko, line 6: Please remove comments all over the project

katevasilenko (changes requested):

app/main.py
avkpol: rag - what is it "rag", what do you mean?

app/main.py
avkpol: why should I remove the comments? It is good practice to comment on a code, isn`t it? What if I leave comments for myself?

README.md
avkpol: yes, mistakenly was edited... how to revert it?

app/main.py
IvanRamyk: code should be self-descriptive. comments isn't a common practice in the Python community. 

if you want to describe how the function works, please, use docstrings

IvanRamyk (changes requested):

README.md
IvanRamyk: you could copy the actual readme from mate academy repository and commit/push it

IvanRamyk (approved):
Well done!

PR #311, opened 2022-12-26
Solution

yuliia-stopkyna:


app/main.py
```diff
@@ -1,3 +1,17 @@
...
-    # Write your code here
-    pass
+from typing import Callable
+
+
+def cache(func: Callable) -> Callable:
+    results = {}
+
+    def inner(*args) -> int:
+        nonlocal results
```
katevasilenko, line 8: You don’t need it

app/main.py
```diff
@@ -1,3 +1,17 @@
...
+        if args in results:
+            print("Getting from cache")
+            return results[args]
+        else:
+            result = func(*args)
+            results[args] = result
+            print("Calculating new result")
+            return result
```
katevasilenko, line 16: Please try to write your code using only one `return`

katevasilenko (changes requested):

IvanRamyk (approved):
Well done!

PR #315, opened 2023-01-02
Solution

serg121192:


Abnormaltype (changes requested):
you can use args as the key and the result of the function as a value. Try to do it

serg121192:
> you can use args as the key and the result of the function as a value. Try to do it
I'm sorry, but I thought I'd already realized this idea in my code, didn't it? I'm about this string: _cache[func.__name__][inputs] = func(*inputs). A dictionary _cache, where stored the names of decorated functions as keys of dictionaries, and theirs values are the dictionaries too, where the keys are the args, and values - the results of functions. Or I didn't understand what's your idea. please, explain :)

app/main.py
```diff
@@ -1,3 +1,18 @@
...
-    pass
+from typing import Any
+
+
+def cache(func: callable) -> callable:
+    _cache = {}
+
+    def some_inputs(*inputs: Any) -> str:
+        if func.__name__ not in _cache:
+            _cache[func.__name__] = {}
```
Abnormaltype, line 9: you store **name of the function** -> you should **args**

Abnormaltype: u use dict in dict - it's unnecessary

Abnormaltype (changes requested):

app/main.py
serg121192: > u use dict in dict - it's unnecessary

okay, I'll fix it. thx)

app/main.py
```diff
@@ -1,3 +1,16 @@
-def cache(func):
-    # Write your code here
-    pass
+from typing import Any
+
+
+def cache(func: callable) -> callable:
+    _cache = {}
+
+    def some_inputs(*inputs: Any) -> Any:
```
Vasyl-Smutok, line 7: 1. I would rename `some_inputs` to `wrapper`
2. *inputs: tuple

app/main.py
```diff
@@ -1,3 +1,16 @@
-def cache(func):
-    # Write your code here
-    pass
+from typing import Any
+
+
+def cache(func: callable) -> callable:
+    _cache = {}
+
```
Vasyl-Smutok, line 6: added     `@wraps(func)`

vsmutok (changes requested):

vsmutok (approved):
Well done 👍

PR #319, opened 2023-01-05
Solution

frameq:


app/main.py
```diff
@@ -1,3 +1,16 @@
...
+        if not isinstance(args, Hashable):
+            return func(*args)
+        if args in cache_dict:
+            print("Getting from cache")
+            return cache_dict[args]
+        print("Calculating new result")
+        cache_dict[args] = func(*args)
+        return cache_dict[args]
+    cache_dict = {}
+    return inner
```
Abnormaltype, line 16:
```suggestion
def cache(func: Callable) -> Callable:
    cache_dict = {}
    def inner(*args: Any) -> Any:
        if not isinstance(args, Hashable):
            return func(*args)
        if args in cache_dict:
            print("Getting from cache")
            return cache_dict[args]
        print("Calculating new result")
        cache_dict[args] = func(*args)
        return cache_dict[args]

    return inner
```

Abnormaltype (changes requested):
try to write with one return

Vasyl-Smutok (approved):
well done 👍

PR #320, opened 2023-01-05
'Solution'

andy77andy:


app/main.py
```diff
@@ -1,3 +1,16 @@
...
+
+
+def cache(func: Callable) -> None:
+    store = {}
+
+    def inner(*args) -> str:
+        if args in store:
+            print("Getting from cache")
+            return store[args]
+        else:
```
olena-hul, line 11: Redundant else

app/main.py
```diff
@@ -1,3 +1,16 @@
...
+            result = func(*args)
+            print("Calculating new result")
+            store[args] = result
+        return result
```
olena-hul, line 15: Redundant variable `result`. We can go without it

olena-hul (changes requested):

app/main.py
```diff
@@ -1,3 +1,14 @@
-def cache(func):
-    # Write your code here
-    pass
+from typing import Callable
+
+
+def cache(func: Callable) -> None:
+    store = {}
+
+    def inner(*args) -> str:
```
olena-hul, line 7: Last remark: add an annotation for args. And I doubt about the return annotations. For outer function it should be  `Callable`, but for inner it should definitely not be `str`, I think `Any` fits better. Everything else looks good

olena-hul (changes requested):

Abnormaltype (changes requested):
try to write with one return

app/main.py
andy77andy: hope, i've handled this task finally)

katevasilenko (approved):
Good job

PR #322, opened 2023-01-07
Solution

numerodiciannove:


app/main.py
```diff
@@ -1,3 +1,16 @@
...
+    non_repeating_elements = {}
+
+    def checking(*elements) -> Any:
+        if elements not in non_repeating_elements:
+            print("Calculating new result")
+            non_repeating_elements[elements] = func(*elements)
+            return non_repeating_elements[elements]
+
+        print("Getting from cache")
+        return non_repeating_elements[elements]
```
katevasilenko, line 14: You are self repeating. Please use one return here

katevasilenko (changes requested):

PR #325, opened 2023-01-12
'Solution'

IvanGuculyak:


app/main.py
```diff
@@ -1,3 +1,16 @@
...
-    pass
+from typing import Callable, Any
+
+
+def cache(func: Callable) -> Callable:
+    results = dict()
+
+    def inner(*args: Any) -> Any:
+
+        if args in results.keys():
```
olena-hul, line 9: `.keys()` is redundant

olena-hul (changes requested):

app/main.py
```diff
@@ -1,3 +1,16 @@
-def cache(func):
-    # Write your code here
-    pass
+from typing import Callable, Any
+
+
+def cache(func: Callable) -> Callable:
+    results = dict()
```
Vasyl-Smutok, line 5: try to choose a better name than `result`
`cache_dict` or something like that

app/main.py
```diff
@@ -1,3 +1,16 @@
...
+
+    def inner(*args: Any) -> Any:
```
Vasyl-Smutok, line 7: I would recommend using the name `wrapper`

app/main.py
```diff
@@ -1,3 +1,16 @@
-def cache(func):
-    # Write your code here
-    pass
+from typing import Callable, Any
+
+
+def cache(func: Callable) -> Callable:
+    results = dict()
+
```
Vasyl-Smutok, line 6: is also a good habit in decorators to use
`@wraps(func)`

vsmutok (changes requested):

vsmutok (approved):
Good Job 👍

PR #335, opened 2023-01-21
Solution

bohdan-yatsyna:


app/main.py
```diff
@@ -1,3 +1,14 @@
...
-    # Write your code here
-    pass
+from typing import Callable, Any
+
+
+def cache(func: Callable) -> Any:
+    cashed_functions = {}
+
+    def inner(*args) -> Any:
+        if args not in cashed_functions.keys():
```
ArsenPidhoretskyi, line 8: Not necessary to use .keys here, but it is okay

ArsenPidhoretskyi (approved):
Well done

app/main.py
bohdan-yatsyna: Thank you for your remark.

PR #337, opened 2023-01-26
'Solution'

natalie-goriela:


app/main.py
```diff
@@ -1,3 +1,18 @@
...
-    # Write your code here
-    pass
+from typing import Callable
+
+
+def cache(func: Callable) -> Callable:
+
+    results = {}
+
+    def inner(*args) -> int | str:
```
ArsenPidhoretskyi, line 8: BTW, it can return Any

app/main.py
```diff
@@ -1,3 +1,18 @@
...
+
+        if args in results:
+            print("Getting from cache")
+            return results[args]
+        else:
+            print("Calculating new result")
+            results[args] = func(*args)
+            return results[args]
```
ArsenPidhoretskyi, line 16: Duplicate code in 12 and 16 lines. Move it after if statement

ArsenPidhoretskyi (changes requested):

andrewbelo (approved):

PR #339, opened 2023-01-26
Solution

AlimFreight:


app/main.py
```diff
@@ -1,3 +1,15 @@
...
+
+
+def cache(func: Callable) -> Callable:
+    cache_dict = {}
+
+    def wrapper(*args) -> Any:
+        if args not in cache_dict:
+            print("Calculating new result")
+            result = func(*args)
+            cache_dict[args] = result
```
ArsenPidhoretskyi, line 11: BTW, you can do as follow
```suggestion
            cache_dict[args] = func(*args)
```

ArsenPidhoretskyi (approved):
Good job!

PR #341, opened 2023-01-30
Solution

Rainmanuk:


app/main.py
```diff
@@ -1,3 +1,14 @@
-def cache(func):
-    # Write your code here
-    pass
+
+def cache(func: callable) -> any:
+    results = {}
+
+    def inner(*args) -> any:
+        if args in results.keys():
```
wQuelS, line 6: `in` keyword as well as `for` loop, for example, are iterating dictionary keys by default. So you don't need to write .keys()

app/main.py
```diff
@@ -1,3 +1,14 @@
...
+            print("Getting from cache")
+            return results[args]
+        else:
+            result = func(*args)
+            results.update({args: result})
+            print("Calculating new result")
+            return result
```
wQuelS, line 13: You can use only one return after `if`/`else` statements. For this purpose you need to do like this:
`results[args] = func(*args)`, and that's it.
There are 2 benefits in such approach: 
- You will need only one `return`
- Code will be written in more "pythonic" way

wQuelS (changes requested):

YuriiKindrat (approved):

PR #347, opened 2023-01-31
Solution

SafonovVladimir:
add Solution

### DIFF
--- a/app/main.py
+++ b/app/main.py
@@ -1,3 +1,17 @@
-def cache(func):
-    # Write your code here
-    pass
+from typing import Callable, Any
+
+
+def cache(func: Callable) -> Any:
+    func_result = {}
+
+    def wrapper(*args) -> Any:
+        if args in func_result.keys():
+            print("Getting from cache")
+            return func_result[args]
+        else:
+            res = func(*args)
+            func_result[args] = res
+            print("Calculating new result")
+            return res
+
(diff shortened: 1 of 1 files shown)


app/main.py
```diff
@@ -1,3 +1,17 @@
...
-    # Write your code here
-    pass
+from typing import Callable, Any
+
+
+def cache(func: Callable) -> Any:
+    func_result = {}
+
+    def wrapper(*args) -> Any:
+        if args in func_result.keys():
```
YuriiKindrat, line 8: no need to use keys() method here, it'll use it by default

app/main.py
```diff
@@ -1,3 +1,17 @@
...
+            print("Getting from cache")
+            return func_result[args]
+        else:
+            res = func(*args)
```
YuriiKindrat, line 12: It is better to use full names for variables, result - not res

app/main.py
```diff
@@ -1,3 +1,17 @@
...
+            func_result[args] = res
+            print("Calculating new result")
+            return res
```
YuriiKindrat, line 15: You can use a single return statement out of the if/else statement instead of two inside it.

YuriiKindrat (changes requested):

YuriiKindrat (approved):

PR #349, opened 2023-02-01
Solution

manjustice:


app/main.py
```diff
@@ -1,3 +1,17 @@
...
+
+    def wrapper(*args: Any) -> Any:
+        if args in operations:
+            print("Getting from cache")
+            return operations[args]
+        else:
+            print("Calculating new result")
+            result = func(*args)
+            operations[args] = result
+            return result
```
YuriiKindrat, line 15: You can use a single return statement out of the if/else statement instead of two inside it.

YuriiKindrat (changes requested):

wQuelS (approved):
Nice

PR #351, opened 2023-02-01
Dev roman senyshyn cache decorator 

AndereLion:


app/main.py
```diff
@@ -1,3 +1,29 @@
...
+
+
+@cache
+def long_time_func(name_a: int, name_b: int, name_c: int) -> int:
+    return (name_a ** name_b ** name_c) % (name_a * name_c)
+
+
+@cache
+def long_time_func_2(n_tuple: tuple, power: int) -> list:
+    return [number ** power for number in n_tuple]
```
Serhii-Leonenko, line 29: You can use it for local testing, but if you push it to GitHub you must remove all redundant which you don't have to push according to the task.

app/main.py
```diff
@@ -1,3 +1,29 @@
...
+        ]
+        if in_cache:
+            print("Getting from cache")
+            return in_cache[0]
+        print("Calculating new result")
+        result = func(*args)
+        database.append({"result": result, "arguments": args})
+        return result
+
+    return inner
```
Serhii-Leonenko, line 19: Your approach is incorrect: you use a list and every time you need to iterate through it and spend extra time.
Hint: use a dictionary as the storage

Serhii-Leonenko (changes requested):

app/main.py
```diff
@@ -1,3 +1,16 @@
-def cache(func):
-    # Write your code here
-    pass
+from typing import Callable
+
+
+def cache(func: Callable) -> Callable:
+    database = {}
+
+    def inner(*args) -> Callable:
```
YuriiKindrat, line 7: are u sure that inner returns Callable ?

app/main.py
```diff
@@ -1,3 +1,16 @@
...
+        if args in database:
+            print("Getting from cache")
+            return database[args]
+        print("Calculating new result")
+        result = func(*args)
+        database[args] = result
+        return result
```
YuriiKindrat, line 14: it is better to use only one return here (cause u return the same thing here), for better code style

YuriiKindrat (changes requested):

app/main.py
AndereLion: Fixed.

PR #354, opened 2023-02-02
Solution

ZabFTFT:


app/main.py
```diff
@@ -1,3 +1,16 @@
...
+
+    def inner_function(*args) -> Any:
+        if args in cashed_data:
+            print("Getting from cache")
+            return cashed_data[args]
+        else:
+            print("Calculating new result")
+            func_returned_value = func(*args)
+            cashed_data[args] = func_returned_value
+            return func_returned_value
```
YuriiKindrat, line 15: u can use a single return statement out of the if/else statement instead of two inside it

YuriiKindrat (changes requested):

app/main.py
ZabFTFT: I've changed something. Is it better now? =)

Serhii-Leonenko (approved):
Well done!

PR #355, opened 2023-02-02
Solution

kuznetsb:


app/main.py
```diff
@@ -1,3 +1,19 @@
...
+            return cached_results[func][args]
+        solution = func(*args, **kwargs)
+        if func in cached_results:
+            cached_results[func][args] = solution
+        else:
+            cached_results[func] = {args: solution}
+        print("Calculating new result")
+        return solution
+
+    return inner
```
Serhii-Leonenko, line 19: In your dict `cached_results`   args should be a key, and the result of the function should be a value for this key. You don't need to make a func as a key, because you decorate this function and results will be available only for it.

Serhii-Leonenko (changes requested):

app/main.py
kuznetsb: Thank you, it's clear for me now.

app/main.py
```diff
@@ -1,3 +1,16 @@
...
+    cached_results = {}
+
+    def inner(*args, **kwargs) -> Any:
+        if args in cached_results.keys():
+            print("Getting from cache")
+            return cached_results[args]
+        solution = func(*args, **kwargs)
+        cached_results[args] = solution
+        print("Calculating new result")
+        return solution
```
Serhii-Leonenko, line 14: Try to rewrite your decorator not to have `return` duplicated

Serhii-Leonenko (changes requested):

wQuelS (approved):
Nice work

PR #356, opened 2023-02-02
Solution

ada-krav:


app/main.py
```diff
@@ -1,3 +1,16 @@
...
+    runs_data = {}
+
+    def saver(*args) -> Any:
+        if args in runs_data:
+            print("Getting from cache")
+            return runs_data[args]
+        else:
+            print("Calculating new result")
+            runs_data[args] = func(*args)
+            return runs_data[args]
```
wQuelS, line 14: with those `return`s you have code duplication. In general, it is not a very good practice. Try to fix this issue

wQuelS (changes requested):

app/main.py
ada-krav: Oh, I see. Thank you!

wQuelS (approved):
Good job!

PR #358, opened 2023-02-02
solution

amber-marichi:


app/main.py
```diff
@@ -1,3 +1,19 @@
...
-    pass
+from typing import Callable
+from functools import wraps
+
+
+def cache(func: Callable) -> Callable:
+    cached_data = dict()
+
+    @wraps(func)
+    def wrapper(*args) -> int:
```
YuriiKindrat, line 9: are u sure, that func will return int ? what if decorated func will return float or smth else ?

YuriiKindrat (changes requested):

YuriiKindrat (approved):

PR #359, opened 2023-02-02
Solution

Zenionoi:


app/main.py
```diff
@@ -1,3 +1,20 @@
...
+
+
+def cache(func: Callable) -> Callable:
+    dict__ = dict()
+
+    @wraps(func)
+    def inner(*args) -> Any:
+        nonlocal dict__
+
+        if not dict__.get((*args, inner.__name__,)) is None:
```
wQuelS, line 12: As I see, here the cache key is a tuple of the function's arguments and the function name `(*args, inner.__name__,)`. This can be memory-inefficient if the function's arguments are large, as it creates a new tuple every time the function is called.

P.S It would be better to use `dict.get` method with a default value instead of using not `dict__.get((*args, inner.__name__,)) is None` to check if the key is present in the dictionary. This will simplify the code and make it easier to understand. Just saying ..

app/main.py
```diff
@@ -1,3 +1,20 @@
...
+            print("Getting from cache")
+            return dict__.get((*args, inner.__name__,))
```
wQuelS, line 14: As i said above it is better to use `args` instead of `(*args, inner.__name__,)` as the key in the dictionary. This will reduce memory usage and simplify the code.

app/main.py
```diff
@@ -1,3 +1,20 @@
...
+from typing import Callable, Any
+from functools import wraps
+
+
+def cache(func: Callable) -> Callable:
+    dict__ = dict()
+
+    @wraps(func)
+    def inner(*args) -> Any:
+        nonlocal dict__
```
wQuelS, line 10: As noted below, using `args` instead of `(*args, inner.__name__,)` as the key in the dictionary would simplify the code and make it easier to understand, and this would render the use of `nonlocal` unnecessary.

Also, Using `nonlocal`  or `global` statements to modify variables declared in an enclosing scope can lead to difficulties in understanding and maintaining the code, and can also make it harder to test and debug. You should use 'em if you really need to.

Therefore, it's generally better to avoid using nonlocal or global statements whenever possible and to use other techniques

app/main.py
```diff
@@ -1,3 +1,20 @@
-def cache(func):
-    # Write your code here
-    pass
+from typing import Callable, Any
+from functools import wraps
+
+
+def cache(func: Callable) -> Callable:
+    dict__ = dict()
```
wQuelS, line 6: That is okay, but in general `dict__ = {}`  syntax is more commonly used and is considered more Pythonic.
It is up to you whether to change it or not, btw.

app/main.py
```diff
@@ -1,3 +1,20 @@
...
+    def inner(*args) -> Any:
+        nonlocal dict__
+
+        if not dict__.get((*args, inner.__name__,)) is None:
+            print("Getting from cache")
+            return dict__.get((*args, inner.__name__,))
+        print("Calculating new result")
+
+        dict__[(*args, inner.__name__,)] = func(*args)
+        return dict__[(*args, inner.__name__,)]
```
wQuelS, line 18: the double return statement in the your solution is redundant and can be simplified. The `return dict__[(*args, inner.__name__,)]` statement can be simplified to reduce the number of dictionary lookups and make the code more readable.

wQuelS (changes requested):
You wanna play? Let's play!

P.S. Don't get me wrong. I really love that you know some advanced features. But a lot of things can be simplified in your solution

app/main.py
Zenionoi: Some cached result equals to zero that break some tests when 'if dict__.get(hash__):' used

Zenionoi: '''if dict__.get(hash__):'''

YuriiKindrat (approved):

PR #360, opened 2023-02-03
Solution

NEkropo1:


app/main.py
```diff
@@ -1,3 +1,23 @@
...
+        for item in kwargs.items():
+            func_arg_kwarg_hash += hash(item)
+
+        if cache_hash_result.get(func_arg_kwarg_hash) is not None:
+            print("Getting from cache")
+            return cache_hash_result.get(func_arg_kwarg_hash)
+
+        print("Calculating new result")
+        cache_hash_result[func_arg_kwarg_hash] = func(*args, **kwargs)
+        return cache_hash_result.get(func_arg_kwarg_hash)
```
YuriiKindrat, line 21: You can use a single return statement out of the if/else statement instead of two inside it.

app/main.py
```diff
@@ -1,3 +1,23 @@
...
+
+def cache(func: Callable) -> Callable:
+    cache_hash_result = {}
+
+    @wraps(func)
+    def results_save_wrapper(*args, **kwargs) -> None:
+        func_arg_kwarg_hash = hash(args) + hash(func)
+
+        for item in kwargs.items():
+            func_arg_kwarg_hash += hash(item)
```
YuriiKindrat, line 13: why do u need to hash all values ? try to simplify ur code make a key like this f.e. :
func_arg_kwarg_hash = (*args, *kwargs.values(), hash(func)) no need to make a loop for the same functionlity

YuriiKindrat (changes requested):

app/main.py
```diff
@@ -1,3 +1,20 @@
...
-    pass
+from typing import Callable
+from functools import wraps
+
+
+def cache(func: Callable) -> Callable:
+    cache_hash_result = {}
+
+    @wraps(func)
+    def results_save_wrapper(*args, **kwargs) -> None:
```
Serhii-Leonenko, line 9: I don't agree that the function returns None )

Serhii-Leonenko (changes requested):
Well done!

app/main.py
NEkropo1: image
Any? Union[Any | None] ? not sure which is better)

Serhii-Leonenko: `Any` will be enough )

Serhii-Leonenko (changes requested):
Use just Any.

Serhii-Leonenko (approved):
Well done!

app/main.py
NEkropo1: > why do u need to hash all values ? try to simplify ur code make a key like this f.e. : func_arg_kwarg_hash = (*args, *kwargs.values(), hash(func)) no need to make a loop for the same functionlity

btw this case was kinda what i was wandering about:
image

PR #363, opened 2023-02-03
Solution

yura-hudzovskyi:


app/main.py
```diff
@@ -1,3 +1,19 @@
...
+
+
+def cache(func: Callable) -> Callable:
+    cache_date = {}
+
+    def inner(*args, **kwargs) -> any:
+        if args in cache_date:
+            print("Getting from cache")
+            for date in cache_date:
+                if date == args:
```
YuriiKindrat, line 11: no need to iterate through the dict, u can just call date by the key, like in return

app/main.py
```diff
@@ -1,3 +1,19 @@
...
+                    return cache_date[args]
+        else:
+            result = func(*args, **kwargs)
+            cache_date[args] = result
+            print("Calculating new result")
+            return result
+
+    return inner
```
YuriiKindrat, line 19: You can use a single return statement out of the if/else statement instead of two inside it.

YuriiKindrat (changes requested):

app/main.py
```diff
@@ -1,3 +1,17 @@
...
+
+    def inner(*args, **kwargs) -> any:
+        if args in cache_date:
+            print("Getting from cache")
+            return cache_date[args]
+        else:
+            result = func(*args, **kwargs)
+            cache_date[args] = result
+            print("Calculating new result")
+            return result
```
wQuelS, line 15: you still have 2 `return` statement. Try to solve the task only with one

wQuelS (changes requested):

app/main.py
```diff
@@ -8,10 +8,10 @@ def inner(*args, **kwargs) -> any:
...
-        else:
-            result = func(*args, **kwargs)
-            cache_date[args] = result
-            print("Calculating new result")
-            return result
+
+        result = func(*args, **kwargs)
+        cache_date[args] = result
+        print("Calculating new result")
+        return result
```
wQuelS, line 15: You still didn't solve an issue with 2 pretty much duplicated returns

wQuelS (changes requested):

wQuelS (approved):
Well done

PR #364, opened 2023-02-03
'Solution'

alexlega:


app/main.py
```diff
@@ -1,3 +1,20 @@
...
-    pass
+import typing
+
+
+def cache(func: typing.Callable) -> typing.Callable:
+    cache_func = {}
+
+    def inner(*args) -> typing.Any:
+
+        if ((args), (func.__name__)) in cache_func.keys():
```
YuriiKindrat, line 9: no need to use keys() method here, it search through keys by default

app/main.py
```diff
@@ -1,3 +1,20 @@
...
+            print("Getting from cache")
+            return cache_func[(args), (func.__name__)]
+
+        else:
+            print("Calculating new result")
+            complited_func = func(*args)
+            key_for_func = ((args), (func.__name__))
+            cache_func[key_for_func] = complited_func
+            return complited_func
```
YuriiKindrat, line 18: You can use a single return statement out of the if/else statement instead of two inside it.

YuriiKindrat (changes requested):

app/main.py
```diff
@@ -1,3 +1,19 @@
...
+
+        if ((args), (func.__name__)) in cache_func:
+            print("Getting from cache")
+
+        else:
+            print("Calculating new result")
+            complited_func = func(*args)
+            key_for_func = ((args), (func.__name__))
+            cache_func[key_for_func] = complited_func
+        return cache_func[(args), (func.__name__)]
```
YuriiKindrat, line 17: to much brackets here ((args), (func.__name__)), u can write (args, func.__name__) for same result

YuriiKindrat (changes requested):

wQuelS (approved):
GJ

PR #365, opened 2023-02-03
Solution

Danil228:


app/main.py
```diff
@@ -1,3 +1,15 @@
...
-    # Write your code here
-    pass
+from typing import Callable, Any
+
+
+def cache(func: Callable) -> Callable:
+    results = {}
+
+    def inner(*args, **kwargs) -> Any:
+        if args not in results.keys():
```
YuriiKindrat, line 8: no need to use keys() method here, it will search through keys by default

YuriiKindrat (changes requested):

app/main.py
```diff
@@ -1,3 +1,15 @@
...
+            print("Calculating new result")
+            results[args] = func(*args, **kwargs)
```
YuriiKindrat, line 10: what if func will accept only **kwargs ?

PR #369, opened 2023-02-03
Solution

Anatolii-Poznyak:


app/main.py
```diff
@@ -1,3 +1,14 @@
-def cache(func):
-    # Write your code here
-    pass
+from typing import Callable
+
+
+def cache(func: Callable) -> Callable:
+    data_cache = {}
+
+    def wrapper(*args) -> Callable:
```
Serhii-Leonenko, line 7: Are you sure that it returns Callable?)

Serhii-Leonenko (changes requested):

app/main.py
Anatolii-Poznyak: nope

YuriiKindrat (approved):

PR #370, opened 2023-02-04
Solution

Yevheniy-Lototskiy:


app/main.py
```diff
@@ -1,3 +1,15 @@
-def cache(func):
-    # Write your code here
-    pass
+from typing import Callable, Any
+
+
+def cache(func: Callable) -> Any:
+    incoming_args = {}
+
+    def inner(*args) -> int:
```
Serhii-Leonenko, line 7: It can return not only int. Use Any for this purpose.

app/main.py
```diff
@@ -1,3 +1,15 @@
-def cache(func):
-    # Write your code here
-    pass
+from typing import Callable, Any
+
+
+def cache(func: Callable) -> Any:
```
Serhii-Leonenko, line 4: It returns the function, use another type annotation.

Serhii-Leonenko (changes requested):

wQuelS (approved):
GJ

PR #371, opened 2023-02-04
solution

Yulia-Mala:


app/main.py
```diff
@@ -1,3 +1,15 @@
-def cache(func):
-    # Write your code here
-    pass
+from typing import Callable
+
+
+def cache(in_func: Callable) -> Callable:
+    result_dict = {}
+
+    def inner(*args, **kwargs) -> Callable:
```
Serhii-Leonenko, line 7: Are you sure that it returns Callable?)

app/main.py
```diff
@@ -1,3 +1,15 @@
...
+        kwargs_tuple = tuple(kwargs.items())
+        if (args, kwargs_tuple) in result_dict:
```
Serhii-Leonenko, line 9: it's better to use a variable where you store arguments. For example `storage_key = (args, tuple(kwargs.items()))`. And then use it in your  code.

Serhii-Leonenko (changes requested):

app/main.py
Yulia-Mala: You are right, I don't really know what type it returns because it returns result of some function that will be cached. So it's Any

app/main.py
Yulia-Mala: done!

Yulia-Mala: ok

YuriiKindrat (approved):

PR #373, opened 2023-02-04
Solution: cache decorator

ZhAlexR:


app/main.py
```diff
@@ -1,3 +1,17 @@
...
-    # Write your code here
-    pass
+from typing import Callable, Any
+
+
+def cache(func: Callable) -> Any:
+    cache_dict = {}
+
+    def wrapper(*args) -> Any:
+        if hash(args) in cache_dict:
```
Serhii-Leonenko, line 8: You don't need to use the hash() function, the dictionary will do it under the hood. Just store args as a key.

app/main.py
```diff
@@ -1,3 +1,17 @@
...
+            print("Getting from cache")
+            return cache_dict[hash(args)]
+        else:
+            func_res = func(*args)
+            cache_dict[hash(args)] = func_res
+            print("Calculating new result")
+            return func_res
```
Serhii-Leonenko, line 15: You can use a single return statement out of the if/else statement instead of two inside it.

app/main.py
```diff
@@ -1,3 +1,17 @@
...
+
+def cache(func: Callable) -> Any:
+    cache_dict = {}
+
+    def wrapper(*args) -> Any:
+        if hash(args) in cache_dict:
+            print("Getting from cache")
+            return cache_dict[hash(args)]
+        else:
+            func_res = func(*args)
```
Serhii-Leonenko, line 12: it's a redundant variable

app/main.py
```diff
@@ -1,3 +1,17 @@
-def cache(func):
-    # Write your code here
-    pass
+from typing import Callable, Any
+
+
+def cache(func: Callable) -> Any:
```
Serhii-Leonenko, line 4: It returns a function, use another annotation instead  of Any

Serhii-Leonenko (changes requested):

YuriiKindrat (approved):

PR #377, opened 2023-02-05
'Solution'

AnnaBoichenko:


app/main.py
```diff
@@ -1,3 +1,16 @@
...
+    cache_dict = {}
+
+    def inner(*args, **kwargs) -> Any:
+        if args not in cache_dict:
+            cache_dict[args] = func(*args)
+            print("Calculating new result")
+            return cache_dict[args]
+        else:
+            print("Getting from cache")
+            return cache_dict[args]
```
Serhii-Leonenko, line 14: You can use a single return statement out of the if/else statement instead of two inside it.

Serhii-Leonenko (changes requested):

wQuelS (approved):
Good job

PR #378, opened 2023-02-06
'Solution'

HalynaPetrova:


app/main.py
```diff
@@ -1,3 +1,17 @@
...
-    pass
+from typing import Callable
+
+
+def cache(func: Callable) -> None:
+
+    new_dict = {}
+
+    def inner(*args) -> str:
+        if args not in new_dict.keys():
```
YuriiKindrat, line 9: no need to use keys method here, it will search through keys by default

app/main.py
```diff
@@ -1,3 +1,17 @@
...
+            new_dict[args] = func(*args)
+            print("Calculating new result")
+            return new_dict[args]
+        else:
+            print("Getting from cache")
+            return new_dict[args]
```
YuriiKindrat, line 15: u can use only one return here

YuriiKindrat (changes requested):

PR #379, opened 2023-02-07
Solution

tytomyr:


app/main.py
```diff
@@ -1,3 +1,15 @@
...
-    # Write your code here
-    pass
+from typing import Callable, Any
+
+
+def cache(func: Callable) -> Any:
+    stocked_cache = {}
+
+    def inner(*args) -> Any:
+        nonlocal stocked_cache
```
YuriiKindrat, line 8: no need to use nonlocal here

app/main.py
```diff
@@ -1,3 +1,15 @@
-def cache(func):
-    # Write your code here
-    pass
+from typing import Callable, Any
+
+
+def cache(func: Callable) -> Any:
```
YuriiKindrat, line 4: invalid type annotation for return

YuriiKindrat (changes requested):

wQuelS (approved):
GJ

PR #380, opened 2023-02-07
'Solution'

ykopatko:


wQuelS (changes requested):

app/main.py
```diff
@@ -1,3 +1,14 @@
-def cache(func):
-    # Write your code here
-    pass
+from typing import Callable, Any
+
+
+def cache(func: Callable) -> Callable:
+    dict_cache = {}
+
+    def inner(*args: Any) -> tuple:
```
Serhii-Leonenko, line 7: `inner` can return not only tuple, it's better to use `Any`

Serhii-Leonenko (changes requested):

wQuelS (approved):
Good job

PR #381, opened 2023-02-09
'Solution'

sasha-semenenko:


app/main.py
```diff
@@ -1,3 +1,14 @@
-def cache(func):
-    # Write your code here
-    pass
+from typing import Callable, Any
+
+
+def cache(func: Callable) -> Callable:
+    array = {}
+
+    def wrapper(*args) -> Any:
```

olena-hul (changes requested):

app/main.py
```diff
@@ -1,3 +1,14 @@
-def cache(func):
-    # Write your code here
-    pass
+from typing import Callable, Any
+
+
+def cache(func: Callable) -> Callable:
+    array = {}
+
+    def wrapper(*args: Callable) -> Any:
```
Bohdan-Salamakha, line 7: u can't call string type or dict type etc.
args should be Any

Bohdan-Salamakha (changes requested):
v1

Bohdan-Salamakha (approved):

PR #382, opened 2023-02-09
Solution

yhwh6:


app/main.py
```diff
@@ -1,3 +1,16 @@
...
+from typing import Callable, Any
+
+
+def cache(func: Callable) -> Callable:
+    cache_result = dict()
+
+    def inner(*args) -> Any:
+        if args in cache_result:
+            print("Getting from cache")
+            return cache_result[args]
```
vasylhnatiuk, line 10: you must use DRY, and here you can avoid duplicate code like `return cache_result[args]` and `return result`

vasylhnatiuk (changes requested):

vasylhnatiuk (approved):
well done!

vasylhnatiuk (approved):

PR #386, opened 2023-02-11
Solution

Esxoyne:


app/main.py
```diff
@@ -1,3 +1,18 @@
...
+
+def cache(func: Callable) -> Callable:
+    cache_dict = {}
+
+    def wrapper(*args, **kwargs) -> Callable:
+        arguments = *args, *kwargs.values()
+        if arguments not in cache_dict:
+            print("Calculating new result")
+            result = func(*args, **kwargs)
+            cache_dict[arguments] = result
```
ArsenPidhoretskyi, line 12: Write inline this)

app/main.py
```diff
@@ -1,3 +1,18 @@
...
+            return result
```
ArsenPidhoretskyi, line 13: Move return statement to the end because it will be the same as in line 16

ArsenPidhoretskyi (changes requested):

Vasyl-Smutok (approved):
good job 👍

PR #387, opened 2023-02-12
Solution

flatchers:


app/main.py
```diff
@@ -1,3 +1,15 @@
-def cache(func):
-    # Write your code here
-    pass
+from typing import Callable, Any
+
+
+def cache(func: Callable) -> Callable:
+    cache_dict = {}
+
+    def inner(*args) -> Any:
```
Vasyl-Smutok, line 7: I think wrapper better function name 

Also add annotation to `args`
and use `functools  wraps`

Vasyl-Smutok (changes requested):

Vasyl-Smutok (approved):
Good Job 👍

PR #388, opened 2023-02-12
Solution

un5oul:


app/main.py
```diff
@@ -1,3 +1,18 @@
...
-    # Write your code here
-    pass
+from typing import Callable
+
+
+def cache(func: Callable) -> Callable:
+
+    storage = {}
+
+    def inner(*args) -> Callable:
```
Bohdan-Salamakha, line 8: what if I will call function with named parameters?
add **kwargs
provide typehint for *args and **kwargs

app/main.py
```diff
@@ -1,3 +1,18 @@
...
+        if args in storage:
+            print("Getting from cache")
+            return storage[args]
+
+        print("Calculating new result")
+        result = func(*args)
```
Bohdan-Salamakha, line 14: add **kwargs

app/main.py
```diff
@@ -1,3 +1,18 @@
...
+
+
+def cache(func: Callable) -> Callable:
+
+    storage = {}
+
+    def inner(*args) -> Callable:
+        if args in storage:
+            print("Getting from cache")
+            return storage[args]
```
Bohdan-Salamakha, line 11: use get method to avoid KeyError

app/main.py
```diff
@@ -1,3 +1,18 @@
...
+
+        print("Calculating new result")
+        result = func(*args)
+        storage[args] = result
```
Bohdan-Salamakha, line 15: use get method to avoid KeyError

Bohdan-Salamakha (changes requested):
v1

app/main.py
olena-hul: Using kwargs is not required in this task. But you should still provide annotation for args

app/main.py
olena-hul, line 11: Use just 1 return

app/main.py
olena-hul, line 14: Avoid using extra variable

olena-hul (changes requested):

un5oul:
Solution 2 added*

app/main.py
Bohdan-Salamakha: > That was in readme:
> "Decorator cache creating for decorating only functions that take immutable arguments."
named args != immutable args

Bohdan-Salamakha: long_time_func(a=1, b=2, c=3) will raise an error without **kwargs handling
just to know

app/main.py
```diff
@@ -1,3 +1,16 @@
...
+def cache(func: Callable) -> Callable:
+
+    storage = {}
+
+    def inner(*args, **kwargs) -> Callable:
+        if args in storage:
+            print("Getting from cache")
+        else:
+            print("Calculating new result")
+            storage[args] = func(*args, **kwargs)
```
Bohdan-Salamakha, line 13: use get() method to avoid KeyError like
`storage.get(args)`

app/main.py
```diff
@@ -1,3 +1,16 @@
...
+        return storage[args]
```
Bohdan-Salamakha, line 14: use get() method to avoid KeyError like
storage.get(args)

app/main.py
```diff
@@ -1,3 +1,16 @@
...
-    # Write your code here
-    pass
+from typing import Callable
+
+
+def cache(func: Callable) -> Callable:
+
+    storage = {}
+
+    def inner(*args, **kwargs) -> Callable:
```
Bohdan-Salamakha, line 8: add typehint for args, kwargs
smth like
`*args: Any, **kwargs: Any`

Bohdan-Salamakha (changes requested):
v2

Bohdan-Salamakha (approved):

Bohdan-Salamakha (approved):

PR #390, opened 2023-02-15
'Solution'

kucherenkodmytro:


app/main.py
```diff
@@ -1,3 +1,16 @@
...
+def cache(func: Callable) -> Callable:
+    stored_cache = {}
+
+    def wrapper(*args) -> Any:
+        if args not in stored_cache:
+            result = func(*args)
+            stored_cache[args] = result
+            print("Calculating new result")
+            return result
+        if args in stored_cache:
```
ArsenPidhoretskyi, line 13: use else

app/main.py
```diff
@@ -1,3 +1,16 @@
...
+from typing import Callable, Any
+
+
+def cache(func: Callable) -> Callable:
+    stored_cache = {}
+
+    def wrapper(*args) -> Any:
+        if args not in stored_cache:
+            result = func(*args)
+            stored_cache[args] = result
```
ArsenPidhoretskyi, line 10: Write in one row
And then move return at the end. No need to repeat the same)

app/main.py
```diff
@@ -1,3 +1,16 @@
...
+            print("Calculating new result")
+            return result
+        if args in stored_cache:
+            print("Getting from cache")
+            return stored_cache[args]
```
ArsenPidhoretskyi, line 15: Move return after-if-else statement

ArsenPidhoretskyi (changes requested):

vasylhnatiuk (approved):
Well done!

PR #393, opened 2023-02-17
Solution

antonpiddubnyi:


app/main.py
```diff
@@ -1,3 +1,16 @@
...
+from typing import Callable
+
+
+def cache(func: Callable) -> Callable:
+    result = {}
+
+    def wrapper(*args: (int, tuple)) -> int:
+        if args in result:
+            print("Getting from cache")
+            return result[args]
```
ArsenPidhoretskyi, line 10: code duplicate move it after if

app/main.py
```diff
@@ -1,3 +1,16 @@
-def cache(func):
-    # Write your code here
-    pass
+from typing import Callable
+
+
+def cache(func: Callable) -> Callable:
+    result = {}
+
+    def wrapper(*args: (int, tuple)) -> int:
```
ArsenPidhoretskyi, line 7: Use Union or |

app/main.py
ArsenPidhoretskyi: live 10 and line 14 the same

ArsenPidhoretskyi (changes requested):

app/main.py
antonpiddubnyi: I don't really understand it, what is duplicated?

app/main.py
```diff
@@ -1,3 +1,16 @@
...
+        if args in result:
+            print("Getting from cache")
+            return result[args]
+        else:
+            print("Calculating new result")
+            result[args] = func(*args)
+            return result[args]
```
Grayder0152, line 14:
```suggestion
        if args in result:
            print("Getting from cache")
        else:
            print("Calculating new result")
            result[args] = func(*args)
        return result[args]
```

Grayder0152 (approved):

Grayder0152 (changes requested):
Resolve comments

app/main.py
ArsenPidhoretskyi: https://github.com/mate-academy/py-cache-decorator/blob/master/checklist.md

ArsenPidhoretskyi: See duplicate and checklist
Not fixed

ArsenPidhoretskyi (changes requested):

PR #394, opened 2023-02-19
Solution

SashaLee:


app/main.py
```diff
@@ -1,3 +1,15 @@
-def cache(func):
-    # Write your code here
-    pass
+from typing import Callable, Any
+
+
+def cache(func: Callable) -> Any:
+    data = {}
+
+    def inner(*args) -> Any:
```
Bohdan-Salamakha, line 7: add typehint for args

app/main.py
```diff
@@ -1,3 +1,15 @@
-def cache(func):
-    # Write your code here
-    pass
+from typing import Callable, Any
+
+
+def cache(func: Callable) -> Any:
```
Bohdan-Salamakha, line 4: return type is not None

Bohdan-Salamakha (changes requested):
v1

app/main.py
SashaLee: "functions that take immutable arguments"
int, float, bool, tuple....
But how can I put it all together?

SashaLee: > add typehint for args

"functions that take immutable arguments"
int, float, bool, tuple....
But how can I put it all together?

SashaLee: "functions that take immutable arguments"
int, float, bool, tuple....
But how can I put it all together?

app/main.py
SashaLee: OK. Return type is Callable

app/main.py
Bohdan-Salamakha: from typing import **Any**

app/main.py
```diff
@@ -1,3 +1,15 @@
-def cache(func):
-    # Write your code here
-    pass
+from typing import Callable, Any
+
+
+def cache(func: Callable) -> Callable:
+    data = {}
+
+    def inner(*args: tuple) -> Any:
```
Bohdan-Salamakha, line 7: args: Any

Bohdan-Salamakha (changes requested):
v2

Bohdan-Salamakha (approved):

PR #395, opened 2023-02-20
'Solution'

struhanets:


app/main.py
```diff
@@ -1,3 +1,15 @@
...
+
+def cache(func: Callable) -> None:
+    list_of_results = {}
+
+    def inner(*args: Any, **kwargs: Any) -> None:
+        if isinstance(args, (int, str, float, bool, tuple)):
+            if args not in list_of_results:
+                print("Calculating new result")
+                list_of_results[args] = func(*args)
+                return list_of_results[args]
```
ArsenPidhoretskyi, line 12: add else statement and remove duplicate of return list_of_results[args]

ArsenPidhoretskyi (changes requested):

app/main.py
struhanets: I can't remove the duplicate "list_of_results[args]" because my tests give an error. only with 2 "return" ok.

ArsenPidhoretskyi: No it is false
You can or pin screenshot)
Because all tasks works without duplicate
You must remove line 12 and move line 15 to one left

ArsenPidhoretskyi (changes requested):

app/main.py
```diff
@@ -1,3 +1,15 @@
-def cache(func):
-    # Write your code here
-    pass
+from typing import Callable, Any
+
+
+def cache(func: Callable) -> None:
+    list_of_results = {}
+
+    def inner(*args: Any, **kwargs: Any) -> None:
```
ArsenPidhoretskyi, line 7: Wrong return type!
Also no need to check for instance in this task no needed

ArsenPidhoretskyi (changes requested):

app/main.py
vasylhnatiuk: it must be ` -> Callable`

vasylhnatiuk: because it is decorator

vasylhnatiuk (changes requested):

app/main.py
```diff
@@ -1,3 +1,14 @@
...
+from typing import Callable, Any
+
+
+def cache(func: Callable) -> Callable:
+    list_of_results = {}
+
+    def inner(*args: Any, **kwargs: Any) -> None:
+        if args not in list_of_results:
+            print("Calculating new result")
+            list_of_results[args] = func(*args)
```
Grayder0152, line 10: What if we have the same arguments but for different functions?

app/main.py
Grayder0152: pls do this

Grayder0152 (changes requested):

Grayder0152 (changes requested):
What if we have the same arguments but for different functions?

Vasyl-Smutok (changes requested):
What if we have the same arguments but for different functions?

struhanets:
there will be conflict and error, but not in this case because we have *args.

app/main.py
```diff
@@ -1,3 +1,14 @@
-def cache(func):
-    # Write your code here
-    pass
+from typing import Callable, Any
+
+
+def cache(func: Callable) -> Callable:
+    list_of_results = {}
+
+    def inner(*args: Any, **kwargs: Any) -> None:
```
katevasilenko, line 7:
```suggestion
    def inner(*args: Any, **kwargs: Any) -> Callable:
```

app/main.py
```diff
@@ -1,3 +1,14 @@
-def cache(func):
-    # Write your code here
-    pass
+from typing import Callable, Any
+
+
+def cache(func: Callable) -> Callable:
+    list_of_results = {}
```
katevasilenko, line 5: It’s picky, but actually it’s not a list but a dict

katevasilenko (changes requested):

katevasilenko (approved):
Looks good to me now

PR #396, opened 2023-02-21
'Solution'

govorunenko:


app/main.py
```diff
@@ -1,3 +1,18 @@
...
-    # Write your code here
-    pass
+from typing import Callable, Any
+
+
+def cache(func: Callable) -> Any:
+
+    result_dict = {}
+
+    def wrapper(*args) -> Any:
```
Bohdan-Salamakha, line 8: add args typehint

app/main.py
```diff
@@ -1,3 +1,18 @@
...
+
+        if args not in result_dict:
+            print("Calculating new result")
+            result_dict[args] = func(*args)
+            return result_dict[args]
+
+        else:
+            print("Getting from cache")
```
Bohdan-Salamakha, line 16: redundant return
just print("Getting from cache")

Bohdan-Salamakha (changes requested):
v1

app/main.py
govorunenko: edited

app/main.py
govorunenko: edited

app/main.py
```diff
@@ -1,3 +1,17 @@
...
+    result_dict = {}
+
+    def wrapper(*args: tuple) -> Any:
+
+        if args not in result_dict:
+            print("Calculating new result")
+            result_dict[args] = func(*args)
+            return result_dict[args]
+
+        print("Getting from cache")
```
```suggestion
        else:
             print("Getting from cache")
```

ArsenPidhoretskyi: Better to do like this

ArsenPidhoretskyi (changes requested):

ArsenPidhoretskyi (approved):
Well done!

PR #400, opened 2023-02-22
'Solution'

HalynaPetrova:


app/main.py
```diff
@@ -1,3 +1,16 @@
...
+
+def cache(func: Callable) -> None:
+    my_cash = {}
+
+    def inner(*args) -> str:
+        if args not in my_cash.keys():
+            my_cash[args] = func(*args)
+            print("Calculating new result")
+            return my_cash[args]
+        else:
```
vasylhnatiuk, line 12: use DRY, `else` is uselles here. you can use only one `return`

vasylhnatiuk (changes requested):

HalynaPetrova:
> What if we have common arguments for different functions?

Hello! Add an argument to a class instead of an instance of the class?

HalynaPetrova:
> > What if we have common arguments for different functions?
> 
> Hello! Add an argument to a class instead of an instance of the class?

Oh, sorry, it's not about this task)))
Use argument name "_"?

vasylhnatiuk (approved):
Well done!

PR #402, opened 2023-02-22
Solution

senkmarianna:


app/main.py
```diff
@@ -1,3 +1,17 @@
-def cache(func):
-    # Write your code here
-    pass
+from typing import Callable
+
+
+def cache(func: Callable) -> Callable:
+    cache_dict = {}
+
+    def wrapper(*args: None) -> None:
```
Bohdan-Salamakha, line 7: args is not None
return type is not None

Bohdan-Salamakha (changes requested):
v1

app/main.py
senkmarianna: Any?

app/main.py
```diff
@@ -1,3 +1,17 @@
...
+from typing import Callable, Any
+
+
+def cache(func: Callable) -> Callable:
+    cache_dict = {}
+
+    def wrapper(*args: Any) -> Any:
+        if args in cache_dict:
+            print("Getting from cache")
+            return cache_dict[args]
```
olena-hul, line 10: Return can be used only once at the end of the function

app/main.py
```diff
@@ -1,3 +1,17 @@
...
+        else:
+            print("Calculating new result")
+            result = func(*args)
+            cache_dict[args] = result
+            return result
```
olena-hul, line 15: result variable is not needed

olena-hul (changes requested):

Bohdan-Salamakha (approved):

PR #403, opened 2023-02-23
solution

dminyaylo:


Grayder0152 (changes requested):
What if we have common arguments for different functions?

dminyaylo:
Now it checks for arguments and the value of it. But i have error and dont know how to fix it

app/main.py
```diff
@@ -1,3 +1,13 @@
 def cache(func):
-    # Write your code here
-    pass
+    results = {}
+
+    def wrapper(*args):
```
vasylhnatiuk, line 4: you must add anotation

app/main.py
```diff
@@ -1,3 +1,13 @@
 def cache(func):
```
vasylhnatiuk, line 1: you must add anotation

app/main.py
```diff
@@ -1,3 +1,13 @@
...
+
+    def wrapper(*args):
+        if args in results and results[args] == func(*args):
+            print("Getting from cache")
+            return results[args]
+        else:
+            print("Calculating new result")
+            result = func(*args)
+            results[args] = result
+            return result
```
vasylhnatiuk, line 12: you must use only one `return results[args] `. don't repeat yourself

vasylhnatiuk (changes requested):

app/main.py
```diff
@@ -1,3 +1,13 @@
-def cache(func):
-    # Write your code here
-    pass
+def cache(func: callable) -> dict:
```
Vasyl-Smutok, line 1:
```suggestion
def cache(func: callable) -> Callable:
```

app/main.py
```diff
@@ -1,3 +1,13 @@
...
+    results = {}
+
+    def wrapper(*args) -> any:
```
Vasyl-Smutok, line 4:
```suggestion
    def wrapper(*args) -> Any:
```

Vasyl-Smutok (changes requested):

PR #414, opened 2023-03-06
Solution

Anyta17:


app/main.py
```diff
@@ -1,3 +1,12 @@
-def cache(func):
-    # Write your code here
-    pass
+def cache(func: any) -> any:
```
Bohdan-Salamakha, line 1: both func and return types are Callcable

Bohdan-Salamakha (changes requested):
v1

app/main.py
```diff
@@ -1,3 +1,15 @@
-def cache(func):
-    # Write your code here
-    pass
+from typing import Callable
+
+
+def cache(func: Callable) -> Callable:
+    results = {}
+
+    def inner(*args) -> any:
```
Bohdan-Salamakha, line 7: args: Any
return type is the same

Bohdan-Salamakha (changes requested):
v2

app/main.py
```diff
@@ -1,3 +1,15 @@
-def cache(func):
-    # Write your code here
-    pass
+from typing import Callable
+
+
+def cache(func: Callable) -> Callable:
+    results = {}
+
+    def inner(*args: any) -> any:
```
Bohdan-Salamakha, line 7: "any" is a function
u need "Any"!!!

Bohdan-Salamakha (changes requested):
v3

Bohdan-Salamakha (approved):

PR #415, opened 2023-03-06
'Solution'

nickkozlov90:


app/main.py
```diff
@@ -1,3 +1,17 @@
...
+
+    def inner(*args) -> Any:
+        for _ in args:
+            if not type(_) in {int, float, tuple, str}:
+                return inner
+        if args in results.keys():
+            print("Getting from cache")
+            return results[args]
+        print("Calculating new result")
+        results.update({args: func(*args)})
```
ArsenPidhoretskyi, line 15: Simplify this as:
```suggestion
        results[args] = func(*args)
```

app/main.py
```diff
@@ -1,3 +1,17 @@
...
+from typing import Callable, Any
+
+
+def cache(func: Callable) -> Callable:
+    results = dict()
+
+    def inner(*args) -> Any:
+        for _ in args:
+            if not type(_) in {int, float, tuple, str}:
+                return inner
```
ArsenPidhoretskyi, line 10: This is not needed. You can ignore this

ArsenPidhoretskyi (changes requested):

app/main.py
nickkozlov90: But in task condition it is mentioned that function should get only immutable arguments. No?

app/main.py
```diff
@@ -1,3 +1,14 @@
-def cache(func):
-    # Write your code here
-    pass
+from typing import Callable, Any
+
+
+def cache(func: Callable) -> Callable:
+    results = dict()
+
+    def inner(*args) -> Any:
```
Bohdan-Salamakha, line 7: `args: Any`

Bohdan-Salamakha (changes requested):
v1

app/main.py
ArsenPidhoretskyi: > Decorator cache creating for decorating only functions that take immutable arguments.
It is means other types can not be send

ArsenPidhoretskyi (approved):
Well done!

PR #417, opened 2023-03-07
Solution with type fix

oigod:


app/main.py
```diff
@@ -1,3 +1,17 @@
...
+
+    def wrapper(*args, **kwargs) -> tuple:
+        if args not in cache_dict.keys():
+            result = func(*args)
+            cache_dict[args] = result
+            print("Calculating new result")
+            return result
+
+        print("Getting from cache")
+        return cache_dict[args]
```
YuriiKindrat, line 15: u can use only one return here, cause u are returning the same thing

YuriiKindrat (changes requested):

PR #422, opened 2023-03-07
CACHE DECORATOR

volodinsy:


app/main.py
```diff
@@ -1,3 +1,25 @@
...
+    data_dict = {}
+
+    def inner(*args) -> Callable:
+        if args in data_dict:
+            print("Getting from cache")
+            return data_dict[args]
+        else:
+            print("Calculating new result")
+            data_dict[args] = func(*args)
+            return data_dict[args]
```
katevasilenko, line 14: Try to write with one return

app/main.py
```diff
@@ -1,3 +1,25 @@
...
+
+
+@cache
+def long_time_func(var_a: int, var_b: int, var_c: int) -> int:
+    return (var_a ** var_b ** var_c) % (var_a * var_c)
+
+
+@cache
+def long_time_func_2(n_tuple: tuple, power: int) -> list:
+    return [number ** power for number in n_tuple]
```
katevasilenko, line 25: You don’t need this in the final code version

katevasilenko (changes requested):

PR #424, opened 2023-03-08
CACHE DECORATOR

volodinsy:


app/main.py
```diff
@@ -1,3 +1,24 @@
...
+
+
+@cache
+def long_time_func(var_a: int, var_b: int, var_c: int) -> int:
+    return (var_a ** var_b ** var_c) % (var_a * var_c)
+
+
+@cache
+def long_time_func_2(n_tuple: tuple, power: int) -> list:
+    return [number ** power for number in n_tuple]
```
ArsenPidhoretskyi, line 24: Remove this code it is not needed here)

app/main.py
```diff
@@ -1,3 +1,24 @@
-def cache(func):
-    # Write your code here
-    pass
+from typing import Callable
+
+
+def cache(func: Callable) -> Callable:
+    data_dict = {}
+
+    def inner(*args) -> Callable:
```
ArsenPidhoretskyi, line 7: It does not return Callable it can return Any, because result of function can be any, we can not know it

ArsenPidhoretskyi (changes requested):

PR #428, opened 2023-03-09
Solution

DmytroBondariev:


app/main.py
```diff
@@ -1,3 +1,16 @@
...
+
+    def wrapper(*args) -> int:
+        if str(args) not in cache_dict.keys():
+            print("Calculating new result")
+            result = func(*args)
+            cache_dict[str(args)] = result
+            return cache_dict[str(args)]
+        else:
+            print("Getting from cache")
+            return cache_dict[str(args)]
```
Grayder0152, line 15:
```suggestion
            cache_dict[str(args)] = result
        else:
            print("Getting from cache")
        return cache_dict[str(args)]
```

Grayder0152 (approved):

app/main.py
```diff
@@ -1,3 +1,16 @@
...
-    # Write your code here
-    pass
+from typing import Callable
+
+
+def cache(func: Callable) -> int:
+    cache_dict = {}
+
+    def wrapper(*args) -> int:
+        if str(args) not in cache_dict.keys():
```
ArsenPidhoretskyi, line 8: No need to check as str. Just use if args not in

app/main.py
```diff
@@ -1,3 +1,16 @@
...
+            print("Calculating new result")
+            result = func(*args)
+            cache_dict[str(args)] = result
```
ArsenPidhoretskyi, line 11: write inline this
and also do not use str(args), just args

ArsenPidhoretskyi: Because for example someone call with function with parameter=1 and will have another result if call with "1"

ArsenPidhoretskyi (changes requested):
Changes Requested

Grayder0152 (approved):

PR #429, opened 2023-03-12
Finish task

U-Ra-All:


app/main.py
```diff
@@ -1,3 +1,20 @@
...
-    pass
+from typing import Callable, Any
+import functools
+
+
+def cache(func: Callable) -> Callable:
+    """Keep a cache of previous function calls"""
+
+    @functools.wraps(func)
+    def wrapper_cache(*args, **kwargs) -> Any:
```
Bohdan-Salamakha, line 9: add typehint for args and kwargs

Bohdan-Salamakha (changes requested):
v1

app/main.py
```diff
@@ -1,3 +1,20 @@
...
-    pass
+from typing import Callable, Any
+import functools
+
+
+def cache(func: Callable) -> Callable:
+    """Keep a cache of previous function calls"""
+
+    @functools.wraps(func)
+    def wrapper_cache(*args: Any, **kwargs: Any) -> Any:
```
Bohdan-Salamakha, line 9: incorrect types
args is a tuple
kwargs is a dict

Bohdan-Salamakha (changes requested):
v2

app/main.py
U-Ra-All: > incorrect types args is a tuple kwargs is a dict

Fixed

Bohdan-Salamakha (approved):

PR #431, opened 2023-03-13
Cache decorator solution

Filllick:


app/main.py
```diff
@@ -1,3 +1,15 @@
...
-    # Write your code here
-    pass
+from typing import Callable, Any
+
+
+def cache(func: Callable) -> Callable:
+    cache_database = {}
+
+    def wrapper(*args) -> Any:
+        if tuple(args) not in cache_database:
```
Vasyl-Smutok, line 8: not need to use `tuple` 

Please fix it here and other places 
```suggestion
        if args not in cache_database:
```

app/main.py
```diff
@@ -1,3 +1,15 @@
-def cache(func):
-    # Write your code here
-    pass
+from typing import Callable, Any
+
+
+def cache(func: Callable) -> Callable:
+    cache_database = {}
+
```
Vasyl-Smutok, line 6: also add `@wraps`

Vasyl-Smutok (changes requested):

app/main.py
```diff
@@ -1,3 +1,17 @@
...
+    @functools.wraps(func)
+    def wrapper(*args) -> Any:
+        if args not in cache_database:
+            print("Calculating new result")
+            cache_database[args] = func(*args)
+            return cache_database[args]
+
+        print("Getting from cache")
```
Grayder0152, line 15:
```suggestion
        else:
            print("Getting from cache")
```

Grayder0152 (approved):

PR #435, opened 2023-03-16
v1

NVMBR13:


app/main.py
```diff
@@ -1,3 +1,12 @@
-def cache(func):
-    # Write your code here
-    pass
+def cache(func: callable) -> callable:
+    cached_results = {}
+
+    def inner(*args: any) -> any:
```
olena-hul, line 4: any is wrong annotation. `Any` should be used

app/main.py
```diff
@@ -1,3 +1,12 @@
...
+        if args in cached_results.keys():
```
olena-hul, line 5:
```suggestion
        if args in cached_results:
```

app/main.py
```diff
@@ -1,3 +1,12 @@
...
+            print("Getting from cache")
+            return cached_results[args]
```
olena-hul, line 7: Redundant return

app/main.py
```diff
@@ -1,3 +1,12 @@
...
+        print("Calculating new result")
+        cached_results.update({args: func(*args)})
```
olena-hul, line 9: Move this to else block

olena-hul (changes requested):

Bohdan-Salamakha (changes requested):
don't work on the master branch - bad practice
https://github.com/mate-academy/py-task-guideline/blob/main/README.md#:~:text=git%20checkout%20%2Db%20develop

PR #436, opened 2023-03-16
solution

goldenuni:


app/main.py
```diff
@@ -1,3 +1,16 @@
...
-    # Write your code here
-    pass
+from typing import Any, Callable
+
+
+def cache(func: Callable) -> Callable:
+    memory = {}
+
+    def decorator(*args: tuple, **kwargs: dict) -> Any:
+        key = str(args) + str(kwargs)
```
ArsenPidhoretskyi, line 8: Why you need to convert to str? Remove this in function will only enter *args

app/main.py
```diff
@@ -1,3 +1,16 @@
...
+        if key in memory:
+            print("Getting from cache")
+            return memory[key]
+        value = func(*args, **kwargs)
+        memory[key] = value
```
ArsenPidhoretskyi, line 13: Write inline this

ArsenPidhoretskyi (changes requested):

Bohdan-Salamakha (changes requested):
don't work on the master branch - bad practice
https://github.com/mate-academy/py-task-guideline/blob/main/README.md#:~:text=git%20checkout%20%2Db%20develop

PR #439, opened 2023-03-22
Problem solution example "Cache decorator"

Niki-Alex:


app/main.py
```diff
@@ -1,3 +1,16 @@
...
+    stock = {}
+
+    def wrapper(*args: Any) -> Any:
+        if args in stock:
+            print("Getting from cache")
+            return stock[args]
+        else:
+            print("Calculating new result")
+            result = func(*args)
+            stock[args] = result
```
ArsenPidhoretskyi, line 14: Write inline this

app/main.py
```diff
@@ -1,3 +1,16 @@
...
+            return result
```
ArsenPidhoretskyi, line 15: No need to duplicate code inside if and else move it after if. See your checklist

ArsenPidhoretskyi (changes requested):

app/main.py
Niki-Alex: Gives an error message
image

vasylhnatiuk (approved):
Good job!

PR #441, opened 2023-03-24
Solution cache decorator

vborovets0:


app/main.py
```diff
@@ -1,3 +1,15 @@
-def cache(func):
-    # Write your code here
-    pass
+from typing import Callable, Any
+
+
+def cache(func: Callable) -> Callable:
+    saved_data = {}
+
+    def wrapper(*args: Any, **kwargs: Any) -> Any:
```
Bohdan-Salamakha, line 7: args is a tuple
kwargs is a dict

Bohdan-Salamakha (changes requested):
v1

app/main.py
vborovets0: Do I need to add type annotations for args, kwargs in real projects? (they have default types)

Bohdan-Salamakha (approved):

PR #448, opened 2023-04-04
Solution of decorator

svyshcho:


app/main.py
```diff
@@ -1,3 +1,14 @@
-def cache(func):
-    # Write your code here
-    pass
+from typing import Callable
+
+
+def cache(func: Callable) -> Callable:
+    dict_of_args = {}
+
+    def inner(*args) -> Callable:
```
Bohdan-Salamakha, line 7: add typehint for args

Bohdan-Salamakha (changes requested):
v1

app/main.py
```diff
@@ -1,3 +1,14 @@
-def cache(func):
-    # Write your code here
-    pass
+from typing import Callable
+
+
+def cache(func: Callable) -> Callable:
+    dict_of_args = {}
+
+    def inner(*args: int) -> Callable:
```
Bohdan-Salamakha, line 7: *args is always a tuple cause * includes any count of objects in a tuple

Bohdan-Salamakha (changes requested):
v2

app/main.py
```diff
@@ -1,3 +1,14 @@
...
-    # Write your code here
-    pass
+from typing import Callable
+
+
+def cache(func: Callable) -> Callable:
+    dict_of_args = {}
+
+    def inner(*args: tuple) -> Callable:
+        if args in dict_of_args.keys():
```
olena-hul, line 8:
```suggestion
        if args in dict_of_args:
```

It will result in same as 

```if args in dict_of_args.keys()```

olena-hul (changes requested):

olena-hul (approved):

olena-hul:
Good job!

PR #456, opened 2023-04-11
Solution

yatsiukoleksandr:


app/main.py
```diff
@@ -1,3 +1,16 @@
...
-    pass
+from typing import Callable
+
+
+def cache(func: Callable) -> Callable:
+    result = {}
+
+    def wrapper(*args, **kwargs) -> Callable:
+        key = args
+        if key in result:
```
ArsenPidhoretskyi, line 9:
```suggestion
        if args in result:
```

app/main.py
```diff
@@ -1,3 +1,16 @@
...
+            print("Getting from cache")
+            return result[key]
```
ArsenPidhoretskyi, line 11: remove ths

ArsenPidhoretskyi: duplicate code with last return

ArsenPidhoretskyi (changes requested):

Grayder0152 (approved):

PR #457, opened 2023-04-13
Solution

KittNeKit:


app/main.py
```diff
@@ -1,3 +1,15 @@
-def cache(func):
-    # Write your code here
-    pass
+from typing import Callable, Any
+
+
+def cache(func: Callable) -> Any:
```
olena-hul, line 4: Return annotation is wrong

app/main.py
```diff
@@ -1,3 +1,15 @@
...
+    cache_store = {}
+
+    def inner(*args) -> Any:
```

olena-hul (changes requested):

Bohdan-Salamakha (approved):

PR #462, opened 2023-04-19

alexgubarev88:


app/main.py
```diff
@@ -1,3 +1,19 @@
...
+        if key in wrapper.cache:
+            print("Getting from cache")
+            return wrapper.cache[key]
+        else:
+            print("Calculating new result")
+            result = func(*args, **kwargs)
+            wrapper.cache[key] = result
+            return result
+
+    wrapper.cache = {}
```
ArsenPidhoretskyi, line 18: Move it after 5 line and remove duplicate code

app/main.py
```diff
@@ -1,3 +1,19 @@
...
+    @functools.wraps(func)
+    def wrapper(*args, **kwargs) -> Callable:
+        key = args + tuple(sorted(kwargs.items()))
+        if key in wrapper.cache:
+            print("Getting from cache")
+            return wrapper.cache[key]
+        else:
+            print("Calculating new result")
+            result = func(*args, **kwargs)
+            wrapper.cache[key] = result
```
ArsenPidhoretskyi, line 15: Write this inline

app/main.py
ArsenPidhoretskyi: Because 11 line and 16 will be the same:
return wrapper.cache[key]

ArsenPidhoretskyi (changes requested):

vasylhnatiuk (approved):
Good job!

PR #463, opened 2023-04-19
Solution

ybondaren:


app/main.py
```diff
@@ -1,3 +1,17 @@
...
-    # Write your code here
-    pass
+from typing import Callable
+
+
+def cache(func: Callable) -> Callable:
+    local_cache = {}
+
+    def wrapper(*args) -> int:
+        if all(isinstance(arg, (int, str, float, tuple)) for arg in args):
```
ArsenPidhoretskyi, line 8: No need to do this check. Due to the task it will be always hashable args

app/main.py
```diff
@@ -1,3 +1,17 @@
...
+            if args in local_cache:
+                print("Getting from cache")
+                return local_cache[args]
+            else:
+                value = func(*args)
+                local_cache[args] = value
```
ArsenPidhoretskyi, line 14: Write this inline

ArsenPidhoretskyi (changes requested):

app/main.py
```diff
@@ -1,3 +1,15 @@
...
+    local_cache = {}
+
+    def wrapper(*args) -> int:
+        if args in local_cache:
+            print("Getting from cache")
+            return local_cache[args]
+        else:
+            local_cache[args] = func(*args)
+            print("Calculating new result")
+            return local_cache[args]
```
ArsenPidhoretskyi, line 14: remove duplicate of code(read checklist for example)
line 10 and 14 the same and can be moved from if-else

ArsenPidhoretskyi (changes requested):

vasylhnatiuk (approved):
Well done!

PR #465, opened 2023-04-22
Solution

Dimdimior:


app/main.py
```diff
@@ -1,3 +1,12 @@
-def cache(func):
-    # Write your code here
-    pass
+def cache(func: callable) -> callable:
```
Bohdan-Salamakha, line 1: incorrect typehint
it shoud be capitalized

app/main.py
```diff
@@ -1,3 +1,12 @@
...
+    cache_dict = {}
+
+    def inner(*args: any) -> callable:
```
Bohdan-Salamakha, line 4: incorrect typehint
function doesn't return a function and u don't need typehint for *args in this case

Bohdan-Salamakha (changes requested):
v1

app/main.py
```diff
@@ -1,3 +1,15 @@
-def cache(func):
-    # Write your code here
-    pass
+from typing import Callable
+
+
+def cache(func: Callable) -> Callable:
+    cache_dict = {}
+
+    def inner(*args) -> any:
```
Bohdan-Salamakha, line 7:
```suggestion
    def inner(*args) -> Any:
```

Bohdan-Salamakha (changes requested):
v1

Bohdan-Salamakha (approved):

PR #466, opened 2023-04-23
'Solution'

VTeteruk:


Grayder0152 (changes requested):

app/main.py
```diff
@@ -1,3 +1,16 @@
-def cache(func):
-    # Write your code here
-    pass
+import typing
```
vasylhnatiuk, line 1:
```suggestion
from typing import Callable, Any
```

app/main.py
```diff
@@ -1,3 +1,16 @@
...
+
+
+def cache(func: typing.Callable) -> typing.Callable:
```
vasylhnatiuk, line 4:
```suggestion
def cache(func: Callable) -> Callable:
```

app/main.py
```diff
@@ -1,3 +1,16 @@
...
+    # data_cache == {args: [func_name, result]}
+    data_cache = dict()
+
+    def init(*args) -> typing.Any:
```
vasylhnatiuk, line 8:
```suggestion
    def init(*args) -> Any:
```

app/main.py
vasylhnatiuk: Don't import the full module if it is not necessary.

vasylhnatiuk (changes requested):

vasylhnatiuk (approved):

PR #468, opened 2023-04-23
Cache decorator implemented

thorium-zero:


Oleg-s-s (approved):
Nice)

app/main.py
```diff
@@ -1,3 +1,15 @@
...
+    cache_vault = {}
+
+    def inner(*args) -> Any:
+        if args in cache_vault:
+            print("Getting from cache")
+            return cache_vault[args]
+        print("Calculating new result")
+        result = func(*args)
+        cache_vault[args] = result
+        return result
```
Oleg-s-s, line 14: You are returning the same value, try using only one return.

Oleg-s-s (changes requested):

app/main.py
thorium-zero: Updated...

Oleg-s-s (approved):
Nice :+1:

PR #474, opened 2023-05-03
Create cash decorator

Bohdan11Dii:


app/main.py
```diff
@@ -1,3 +1,16 @@
...
+    cash_dict = {}
+
+    def wrapper(*args) -> Any:
+        if args in cash_dict:
+            print("Getting from cache")
+            return cash_dict[args]
+
+        print("Calculating new result")
+        cash_dict[args] = func(*args)
+        return cash_dict[args]
```
ArsenPidhoretskyi, line 14:
```suggestion
        else:
            print("Calculating new result")
            cash_dict[args] = func(*args)
        return cash_dict[args]
```

ArsenPidhoretskyi: Better to do like this. It will remove all code duplicate

ArsenPidhoretskyi (changes requested):

app/main.py
```diff
@@ -1,3 +1,16 @@
-def cache(func):
-    # Write your code here
-    pass
+from typing import Callable, Any
+
+
+def cache(func: Callable) -> Callable:
+    cash_dict = {}
+
+    def wrapper(*args) -> Any:
```

olena-hul (changes requested):

Bohdan-Salamakha (approved):

PR #478, opened 2023-05-06
'Solution_06/05/23'

TymchenkoV:


app/main.py
```diff
@@ -1,3 +1,14 @@
...
+def cache(func: Callable) -> Callable:
+    results = {}
+
+    def inner(*args) -> Any:
+        if args not in results:
+            print("Calculating new result")
+            results[args] = func(*args)
+            return results[args]
+        else:
+            print("Getting from cache")
```
ArsenPidhoretskyi, line 13:
```suggestion
        else:
            print("Getting from cache")
        return results[args]
```
In else case you does not return anything from function

ArsenPidhoretskyi (changes requested):

app/main.py
```diff
@@ -1,3 +1,15 @@
...
+            return results[args]
```
ArsenPidhoretskyi, line 14: Remove code duplicate read checklist.
Duplicate code lines: 11 and 14

ArsenPidhoretskyi (changes requested):

Vasyl-Smutok (approved):
Good Job 👍

PR #480, opened 2023-05-08
Solution

vladis9:


app/main.py
```diff
@@ -1,3 +1,14 @@
...
+
+def cache(func: Callable) -> Callable:
+    items_in_cached = {}
+
+    def wrapper(*args) -> Callable:
+        if args in items_in_cached.keys():
+            print("Getting from cache")
+            return items_in_cached[args]
+        print("Calculating new result")
+        items_in_cached.update({args: func(*args)})
```
ArsenPidhoretskyi, line 12: Update is redundant in this case. You can do:
items_in_cached[args] = func(*args)

ArsenPidhoretskyi: Also you can use else and than remove the same code: 10 and 13 lines

ArsenPidhoretskyi (changes requested):

ArsenPidhoretskyi (approved):
Well done!

PR #483, opened 2023-05-10
implemented cache decorator

S-SMTN:
Not sure about the size of the function and about the naiming of some variables. But I can't do better. Anyway, LGTM

### DIFF
--- a/app/main.py
+++ b/app/main.py
@@ -1,3 +1,36 @@
-def cache(func):
-    # Write your code here
-    pass
+from typing import Callable, Any
+from functools import wraps
+
+
+def cache(func: Callable) -> Callable:
+    run_log = {}
+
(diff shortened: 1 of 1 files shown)


app/main.py
```diff
@@ -1,3 +1,36 @@
...
-    pass
+from typing import Callable, Any
+from functools import wraps
+
+
+def cache(func: Callable) -> Callable:
+    run_log = {}
+
+    @wraps(func)
+    def wrapper(*args: Any) -> Callable:
```
Bohdan-Salamakha, line 9: return type isn't Callable

Bohdan-Salamakha (changes requested):
v1

Bohdan-Salamakha (approved):

PR #484, opened 2023-05-12
'Solution'

olmatiash:
Hi!

### DIFF
--- a/app/main.py
+++ b/app/main.py
@@ -1,3 +1,20 @@
-def cache(func):
-    # Write your code here
-    pass
+from functools import wraps
+from typing import Any, Callable
+
+
+def cache(func: Callable) -> Callable:
+    memo = {}
+
+    @wraps(func)
+    def wrapper(*args, **kwargs) -> Any:
+        key = args + tuple(sorted(kwargs.items()))
+        if key in memo:
+            print("Getting from cache")
+            return memo[key]
+        else:
+            print("Calculating new result")
+            result = func(*args, **kwargs)
+            memo[key] = result
+            return result
+
+    return wrapper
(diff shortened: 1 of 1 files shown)


app/main.py
```diff
@@ -1,3 +1,20 @@
-def cache(func):
-    # Write your code here
-    pass
+from functools import wraps
+from typing import Any, Callable
+
+
+def cache(func: Callable) -> Callable:
+    memo = {}
```
vasylhnatiuk, line 6: memo is a bad name

app/main.py
```diff
@@ -1,3 +1,20 @@
...
+
+    @wraps(func)
+    def wrapper(*args, **kwargs) -> Any:
+        key = args + tuple(sorted(kwargs.items()))
+        if key in memo:
+            print("Getting from cache")
+            return memo[key]
+        else:
```
vasylhnatiuk, line 14: else is useless here, because you have 3 returns

app/main.py
```diff
@@ -1,3 +1,20 @@
...
+            print("Calculating new result")
+            result = func(*args, **kwargs)
+            memo[key] = result
+            return result
```
vasylhnatiuk, line 18:
```suggestion
            return memo[key]
```

vasylhnatiuk (changes requested):

app/main.py
olmatiash: okay

app/main.py
olmatiash: changed

app/main.py
olmatiash: okay

Grayder0152 (approved):

PR #488, opened 2023-05-13
develop

vilisovv:


app/main.py
```diff
@@ -1,3 +1,15 @@
-def cache(func):
-    # Write your code here
-    pass
+from typing import Callable
+
+
+def cache(func: Callable) -> Callable:
+    cache_item = {}
+
+    def long_time(*args) -> str:
```
Vasyl-Smutok, line 7: I think `wrapper` better name 
also `str` wrong annotation, use `Any`
```suggestion
    def wrapper(*args) -> Any:
```

app/main.py
```diff
@@ -1,3 +1,15 @@
-def cache(func):
-    # Write your code here
-    pass
+from typing import Callable
+
+
+def cache(func: Callable) -> Callable:
+    cache_item = {}
+
```
Vasyl-Smutok, line 6: use `@functools.wraps`

Vasyl-Smutok (changes requested):

ArsenPidhoretskyi (approved):
Well done!

PR #490, opened 2023-05-15
'Solution'

prykhodkom:


app/main.py
```diff
@@ -1,3 +1,16 @@
-def cache(func):
-    # Write your code here
-    pass
+from typing import Callable, Any
+
+
+def cache(func: Callable) -> Any:
```
Bohdan-Salamakha, line 4: return type is Callable too

app/main.py
```diff
@@ -1,3 +1,16 @@
...
+    result_dict = {}
+
+    def wrapper(*args) -> None:
```
Bohdan-Salamakha, line 7: incorrect return type

Bohdan-Salamakha (changes requested):
v1

Bohdan-Salamakha (approved):

PR #494, opened 2023-05-17
'Solution'

VadSsS:


app/main.py
```diff
@@ -1,3 +1,20 @@
...
-    pass
+from functools import wraps
+from typing import Any, Callable
+
+
+def cache(func: Callable) -> Callable:
+    results = {}
+
+    @wraps(func)
+    def wrapper(*args, **kwargs) -> Any:
```
olena-hul, line 9: Add annotations for args and kwargs

app/main.py
```diff
@@ -1,3 +1,20 @@
...
+        key = (args, tuple(sorted(kwargs.items())))
+        if key in results:
+            print("Getting from cache")
+            return results[key]
```
olena-hul, line 13:
```suggestion
```

app/main.py
```diff
@@ -1,3 +1,20 @@
...
+        else:
+            print("Calculating new result")
+            result = func(*args, **kwargs)
+            results[key] = result
+            return result
```
olena-hul, line 18:
```suggestion
        return results[key]
```

app/main.py
```diff
@@ -1,3 +1,20 @@
...
+
+    @wraps(func)
+    def wrapper(*args, **kwargs) -> Any:
+        key = (args, tuple(sorted(kwargs.items())))
+        if key in results:
+            print("Getting from cache")
+            return results[key]
+        else:
+            print("Calculating new result")
+            result = func(*args, **kwargs)
```
olena-hul, line 16:
```suggestion
            result[key] = func(*args, **kwargs)
```

app/main.py
olena-hul: You can move return 1 level above, then you'll use just 1 return for both if and else blocks

olena-hul (changes requested):

Bohdan-Salamakha (approved):

PR #496, opened 2023-05-19
Implementation cache decorator

esemesenko:


app/main.py
```diff
@@ -1,3 +1,18 @@
...
+def cache(func: Callable) -> Callable:
+
+    cache_dict = {}
+
+    def inner(*args) -> Any:
+        nonlocal cache_dict
+        if args in cache_dict:
+            print("Getting from cache")
+            return cache_dict[args]
+        else:
```
Bohdan-Salamakha, line 13: remove redundant "else"

Bohdan-Salamakha (changes requested):
v1

app/main.py
esemesenko: removed redundant "else"

Bohdan-Salamakha (approved):

PR #501, opened 2023-05-22
Solution

sulym:


app/main.py
```diff
@@ -1,3 +1,20 @@
...
+
+    @wraps(func)
+    def wrapper(*args, **kwargs) -> Any:
+        key = args + tuple(sorted(kwargs.items()))
+        if key in cached_results:
+            print("Getting from cache")
+            return cached_results[key]
+        else:
+            result = func(*args, **kwargs)
+            cached_results[key] = result
```
olena-hul, line 16:
```suggestion
            cached_results[key]= func(*args, **kwargs)
```

app/main.py
```diff
@@ -1,3 +1,20 @@
...
+            print("Calculating new result")
+            return result
```
olena-hul, line 18: You can just return cached_results[key] and remove return from the if block

olena-hul (changes requested):

Bohdan-Salamakha (approved):

PR #506, opened 2023-05-25
'Solution'

ant-komarov:


app/main.py
```diff
@@ -1,3 +1,14 @@
...
+def cache(func: Callable) -> Callable:
+    stored_data = {}
+
+    def wrapper(*args) -> int:
+        if args not in stored_data:
+            print("Calculating new result")
+            stored_data[args] = func(*args)
+            return stored_data[args]
+        print("Getting from cache")
+        return stored_data[args]
```
YuriiKindrat, line 13: u are returning the same thing every time, could u use only one return here?)
also does ur wrapper returns int ?

YuriiKindrat (changes requested):

LiudmylaKulinchenko (approved):
Well done! 🎈

PR #507, opened 2023-05-25
Solution

Po-SvinSkipodla:


app/main.py
```diff
@@ -1,3 +1,17 @@
...
-    # Write your code here
-    pass
+from typing import Callable
+
+
+def cache(func: Callable) -> Callable:
+    storage = {}
+
+    def inner(*args) -> int:
+        nonlocal storage
```
Oleg-s-s, line 8: No need to call nonlocal here, read more about usage of it

app/main.py
```diff
@@ -1,3 +1,17 @@
...
+        if args not in storage:
+            print("Calculating new result")
+            storage[args] = func(*args)
+            return storage[args]
+        else:
```
Oleg-s-s, line 13: unnecessary else

Oleg-s-s (changes requested):

YuriiKindrat (approved):

PR #508, opened 2023-05-25
'Solution'

diana-shyrokikh:


Oleg-s-s (changes requested):

app/main.py
```diff
@@ -1,3 +1,15 @@
...
+    saved_results = {}
+
+    def inner(*args) -> Any:
+        if args in saved_results:
+            print("Getting from cache")
+            return saved_results[args]
+
+        print("Calculating new result")
+        saved_results[args] = func(*args)
+        return saved_results[args]
```
YuriiKindrat, line 14: u are using return for same thing twice, could u make this, with only one return ? )

YuriiKindrat (changes requested):

LiudmylaKulinchenko (approved):
Well done 🐣

PR #510, opened 2023-05-25
Solution

AntonShpakovych:


app/main.py
```diff
@@ -1,3 +1,15 @@
-def cache(func):
-    # Write your code here
-    pass
+from typing import Callable
+
+
+def cache(func: Callable) -> Callable:
+    stored_cashe = {}
+
+    def wrapper(*args) -> int:
```
LiudmylaKulinchenko, line 7: Are you sure this function will always return integer value?

LiudmylaKulinchenko (changes requested):

YuriiKindrat (approved):

PR #511, opened 2023-05-25
'Solution'

oleksiikolii:


app/main.py
```diff
@@ -1,3 +1,15 @@
...
+from typing import Callable, Any
+
+
+def cache(func: Callable) -> Callable:
+    cached = {}
+
+    def inner(*args) -> Any:
+        if args not in cached:
+            result = func(*args)
+            cached[args] = result
```
LiudmylaKulinchenko, line 10: No need to use additional variable `result` here

LiudmylaKulinchenko (changes requested):

LiudmylaKulinchenko (approved):
Good job 😎

PR #512, opened 2023-05-25
Solution

Kirontiko:


app/main.py
```diff
@@ -1,3 +1,17 @@
...
-    # Write your code here
-    pass
+from typing import Callable
+
+
+def cache(func: Callable) -> Callable:
+    stored_results = dict()
+
+    def wrapper(*args) -> Callable:
+
```
LiudmylaKulinchenko, line 8: Mind the code style, you don't have an empty line after assigning the previous function. Remove the empty line here

app/main.py
```diff
@@ -1,3 +1,17 @@
...
+        if args in stored_results:
+            print("Getting from cache")
+            return stored_results[args]
+        else:
+            new_result = func(*args)
```
LiudmylaKulinchenko, line 13: Remove unnecessary additional variable here (`new_result`)

LiudmylaKulinchenko (changes requested):

app/main.py
```diff
@@ -1,3 +1,15 @@
-def cache(func):
-    # Write your code here
-    pass
+from typing import Callable
+
+
+def cache(func: Callable) -> Callable:
+    stored_results = dict()
```
YuriiKindrat, line 5: it is more tipical to make like this `stored_results = {}`, but it's up to u

app/main.py
```diff
@@ -1,3 +1,15 @@
...
+
+    def wrapper(*args) -> Callable:
+        if args in stored_results:
+            print("Getting from cache")
+            return stored_results[args]
+        else:
+            stored_results[args] = func(*args)
+            print("Calculating new result")
+            return stored_results[args]
```
YuriiKindrat, line 14: u are returning the same thing every time, could u make this with one return ?

YuriiKindrat (changes requested):

app/main.py
Kirontiko: > it is more tipical to make like this `stored_results = {}`, but it's up to u

I heard that in previous versions of python {} was to assign a set collection. Don't know that they changed it

YuriiKindrat: it seems to me, that it can be the syntax of 2.x python

YuriiKindrat (approved):

app/main.py
Kirontiko: > it seems to me, that it can be the syntax of 2.x python

Thank you for the information! I will use {} instead of dict() in my next solutions

PR #513, opened 2023-05-25
'Solution'

RedCatHot:


app/main.py
```diff
@@ -1,3 +1,14 @@
-def cache(func):
-    # Write your code here
-    pass
+from typing import Callable
+
+
+def cache(func: Callable) -> None:
```
LiudmylaKulinchenko, line 4: Does your function really returns None?

app/main.py
```diff
@@ -1,3 +1,14 @@
...
+    cache = {}
+
+    def inner(*args) -> None:
```
LiudmylaKulinchenko, line 7: Fix annotations also here

app/main.py
```diff
@@ -1,3 +1,14 @@
-def cache(func):
-    # Write your code here
-    pass
+from typing import Callable
+
+
+def cache(func: Callable) -> None:
+    cache = {}
```
LiudmylaKulinchenko, line 5: Better to change the variable name, because it is the same as your function name

LiudmylaKulinchenko (changes requested):

Oleg-s-s (approved):
Good job)

PR #514, opened 2023-05-25
implemented py-cache-decorator

mariiahorbova:


app/main.py
```diff
@@ -1,3 +1,18 @@
...
+    @wraps(func)
+    def inner(*args) -> Callable:
+        if args in cached:
+            print("Getting from cache")
+            return cached[args]
+        else:
+            result = func(*args)
+            cached[args] = result
+            print("Calculating new result")
+            return result
```
LiudmylaKulinchenko, line 17: Try to rewrite your decorator not to have `return` duplicated

app/main.py
```diff
@@ -1,3 +1,18 @@
...
-    pass
+from typing import Callable
+from functools import wraps
+
+
+def cache(func: Callable) -> Callable:
+    cached = {}
+
+    @wraps(func)
+    def inner(*args) -> Callable:
```
LiudmylaKulinchenko, line 9: Does your function really returns Callable?

LiudmylaKulinchenko (changes requested):

app/main.py
mariiahorbova: done

app/main.py
mariiahorbova: done

app/main.py
```diff
@@ -1,3 +1,17 @@
...
+
+    @wraps(func)
+    def inner(*args) -> Any:
+        if args in cached:
+            print("Getting from cache")
+            return cached[args]
+        else:
+            cached[args] = func(*args)
+            print("Calculating new result")
+            return cached[args]
```
LiudmylaKulinchenko, line 16: You can use a single return statement out of the if/else statement instead of two inside it ☀

LiudmylaKulinchenko (changes requested):

app/main.py
mariiahorbova: ☀

YuriiKindrat (approved):

PR #515, opened 2023-05-25
Solution

YKonoplyov:


app/main.py
```diff
@@ -1,3 +1,19 @@
...
-    # Write your code here
-    pass
+from typing import Callable, Any
+
+
+def cache(func: Callable) -> Callable:
+    cache_dict = {}
+
+    def inner(*args) -> Any:
+        nonlocal cache_dict
```
LiudmylaKulinchenko, line 8: Nonlocal is unneeded here, read more about it by the [link](https://www.geeksforgeeks.org/python-nonlocal-keyword/)

app/main.py
```diff
@@ -1,3 +1,19 @@
...
+        res = 0
```
LiudmylaKulinchenko, line 9: Try to rewrite without additional varialbe `res`

app/main.py
```diff
@@ -1,3 +1,19 @@
...
+        if args not in cache_dict:
+            print("Calculating new result")
+            res = func(*args)
+            cache_dict[args] = res
+        elif args in cache_dict:
```
LiudmylaKulinchenko, line 14: Bettter to use `else` instead

LiudmylaKulinchenko (changes requested):

LiudmylaKulinchenko (approved):
Good job 🍬

PR #516, opened 2023-05-25
Solution

volodymyr-komarnyckyi:


app/main.py
```diff
@@ -1,3 +1,15 @@
-def cache(func):
-    # Write your code here
-    pass
+from typing import Callable
+
+
+def cache(func: Callable) -> Callable:
+    calculated_data = {}
+
+    def inner(*args) -> int:
```
LiudmylaKulinchenko, line 7: Are u sure that inner will always return integer value?

LiudmylaKulinchenko (changes requested):

LiudmylaKulinchenko (approved):
Well done!!

PR #519, opened 2023-05-26
'Solution'

yakhlamova:


app/main.py
```diff
@@ -1,3 +1,20 @@
...
+    @wraps(func)
+    def inner(*args, **kwargs) -> Any:
+        key = (args, tuple(sorted(kwargs.items())))
+        if key in results:
+            print("Getting from cache")
+            return results[key]
+        else:
+            print("Calculating new result")
+            result = func(*args, **kwargs)
+            results[key] = result
```
Oleg-s-s, line 17: Unnecessary "result" variable

Oleg-s-s (changes requested):

app/main.py
```diff
@@ -1,3 +1,19 @@
...
+    @wraps(func)
+    def inner(*args, **kwargs) -> Any:
+        key = (args, tuple(sorted(kwargs.items())))
+        if key in results:
+            print("Getting from cache")
+            return results[key]
+        else:
+            print("Calculating new result")
+            results[key] = func(*args, **kwargs)
+            return results[key]
```
YuriiKindrat, line 17: could u use only one return here?)

YuriiKindrat (changes requested):

app/main.py
```diff
@@ -1,3 +1,18 @@
...
+from typing import Callable, Any
+
+
+def cache(func: Callable) -> Callable:
+    results = {}
+
+    @wraps(func)
+    def inner(*args, **kwargs) -> Any:
+        key = (args, tuple(sorted(kwargs.items())))
+        if key in results:
```
Oleg-s-s, line 11: Unnecessary checking, we are receiving only args, you can then iterate over them
```suggestion
    def inner(*args) -> Any:
        if args in results:
```

Oleg-s-s (approved):

PR #531, opened 2023-05-26
Develop4

12-deadblue:


app/main.py
```diff
@@ -1,3 +1,15 @@
...
+    def inner(*args, **kwargs) -> callable:
+        key = args + tuple(sorted(kwargs.items()))
+        if key in results:
+            print("Getting from cache")
+            return results[key]
+        else:
+            print("Calculating new result")
+            result = func(*args, **kwargs)
+            results[key] = result
+            return result
```
YuriiKindrat, line 13: gj, but try to use only one return for ur solution, cause u are returning the same thing twice )

YuriiKindrat (changes requested):

app/main.py
12-deadblue: i am closed this branch can you go to develop5  the main content now there

PR #532, opened 2023-05-27
Solution

Andriy-Sydorenko:


app/main.py
```diff
@@ -1,3 +1,17 @@
...
-    # Write your code here
-    pass
+from typing import Callable, Union
+
+
+def cache(func: Callable) -> Callable:
+    cached_data = {}
+
+    def wrapper(*args, **kwargs) -> Union[int, list]:
+        key = (args, tuple(kwargs.items()))
```
Oleg-s-s, line 8: No need to unpack data

app/main.py
```diff
@@ -1,3 +1,17 @@
...
+        if key in cached_data:
+            print("Getting from cache")
+            result = cached_data[key]
+        else:
+            print("Calculating new result")
+            result = func(*args, **kwargs)
+            cached_data[key] = result
+        return result
```
Oleg-s-s, line 16: result is not necessary variable here

Oleg-s-s (changes requested):

app/main.py
```diff
@@ -1,3 +1,15 @@
-def cache(func):
-    # Write your code here
-    pass
+from typing import Callable, Union
+
+
+def cache(func: Callable) -> Callable:
+    cached_data = {}
+
+    def wrapper(*args) -> Union[int, list]:
```
YuriiKindrat, line 7: what if decorated func will return a dict or tuple ? fix ur annoation for return type )

YuriiKindrat (changes requested):

YuriiKindrat (approved):

PR #534, opened 2023-05-28
Solution

OlhaTryhub:


app/main.py
```diff
@@ -1,3 +1,14 @@
...
-    # Write your code here
-    pass
+from typing import Callable
+
+
+def cache(func: Callable) -> Callable:
+    results_dict = {}
+
+    def inner(*args) -> int:
+        if (*args,) in results_dict:
```
LiudmylaKulinchenko, line 8: The `args` parameter already represents a tuple of arguments, so there is no need to pack it into another tuple `(*args,) `

LiudmylaKulinchenko (changes requested):

Oleg-s-s (approved):
Good job)

LiudmylaKulinchenko (approved):
🌺

PR #535, opened 2023-05-28
Solution

arturiermolenko:
Cache - decorator was created.

### DIFF
--- a/app/main.py
+++ b/app/main.py
@@ -1,10 +1,10 @@
-from typing import Callable
+from typing import Callable, Any
 
 
 def cache(func: Callable) -> Callable:
     cache_dictionary = {}
 
-    def inner(*args, **kwargs) -> any:
+    def inner(*args, **kwargs) -> Any:
         cache_item = func.__name__ + "".join(str(args)) + "".join(str(kwargs))
         if cache_item in cache_dictionary:
             print("Getting from cache")
--- a/app/main.py
+++ b/app/main.py
@@ -1,12 +1,11 @@
 from typing import Callable
 
 
-def cache(func: Callable) -> Callable:
+def cache(func: Callable) -> any:
     cache_dictionary = {}
 
     def inner(*args, **kwargs) -> str:
-        cache_item = str(func.__name__) + "".join(str(args)) + \
-            "".join(str(kwargs))
+        cache_item = func.__name__ + "".join(str(args)) + "".join(str(kwargs))
         if cache_item in cache_dictionary:
             print("Getting from cache")
             return cache_dictionary[cache_item]
--- a/app/main.py
+++ b/app/main.py
@@ -1,10 +1,10 @@
 from typing import Callable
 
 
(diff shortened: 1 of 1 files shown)


app/main.py
```diff
@@ -1,3 +1,16 @@
...
-    pass
+from typing import Callable
+
+
+def cache(func: Callable) -> Callable:
+    cache_dictionary = {}
+
+    def inner(*args, **kwargs) -> str:
+        cache_item = str(func.__name__) + "".join(str(args)) + \
+            "".join(str(kwargs))
```
Bohdan-Salamakha, line 9: u don't need to polymorph item into string

app/main.py
```diff
@@ -1,3 +1,16 @@
-def cache(func):
-    # Write your code here
-    pass
+from typing import Callable
+
+
+def cache(func: Callable) -> Callable:
+    cache_dictionary = {}
+
+    def inner(*args, **kwargs) -> str:
```
Bohdan-Salamakha, line 7: return type can be any

Bohdan-Salamakha (changes requested):

app/main.py
```diff
@@ -1,3 +1,15 @@
-def cache(func):
-    # Write your code here
-    pass
+from typing import Callable
+
+
+def cache(func: Callable) -> any:
```
Bohdan-Salamakha, line 4: invalid return type for cache() func

app/main.py
Bohdan-Salamakha: it's for inner() func

Bohdan-Salamakha (changes requested):

app/main.py
```diff
@@ -1,3 +1,15 @@
-def cache(func):
-    # Write your code here
-    pass
+from typing import Callable
+
+
+def cache(func: Callable) -> Callable:
+    cache_dictionary = {}
+
+    def inner(*args, **kwargs) -> any:
```
EdAlekseiev, line 7: any is the function, you can't use it for type hints. You have to use `typing.Any`. Read more: https://docs.python.org/3/library/typing.html#the-any-type

EdAlekseiev (changes requested):

app/main.py
```diff
@@ -1,3 +1,15 @@
...
+
+
+def cache(func: Callable) -> Callable:
+    cache_dictionary = {}
+
+    def inner(*args, **kwargs) -> Any:
+        cache_item = func.__name__ + "".join(str(args)) + "".join(str(kwargs))
+        if cache_item in cache_dictionary:
+            print("Getting from cache")
+            return cache_dictionary[cache_item]
```
olena-hul, line 11: Try to use just 1 return statement, now you're using 2

olena-hul (changes requested):

olena-hul (approved):

olena-hul:
Good job!

PR #536, opened 2023-05-29
done

maccarets:


app/main.py
```diff
@@ -1,3 +1,14 @@
...
+def cache(func: Callable) -> Any:
+    storage = {}
+
+    def inner(*args) -> Any:
+        if args not in storage:
+            print("Calculating new result")
+            storage[args] = func(*args)
+            return storage[args]
+        print("Getting from cache")
+        return storage[args]
```
Oleg-s-s, line 13: Better make this through 1 return

Oleg-s-s (changes requested):

app/main.py
```diff
@@ -1,3 +1,14 @@
-def cache(func):
-    # Write your code here
-    pass
+from typing import Callable, Any
+
+
+def cache(func: Callable) -> Any:
```
YuriiKindrat, line 4: are u sure about type annotation for return type? it seems to me, that cache func returns callable

YuriiKindrat (changes requested):

app/main.py
maccarets: u r right

PR #538, opened 2023-05-29
The function is working.

stepband:
Perhaps not very elegantly, but the function is working.

### DIFF
--- a/app/main.py
+++ b/app/main.py
@@ -1,3 +1,18 @@
-def cache(func):
-    # Write your code here
-    pass
+from typing import Any, Callable
+
+
+def cache(func: Callable) -> Callable:
+    results = {}
+
+    def wrapper(*args) -> Any:
+        key = args
+        if key in results:
+            print("Getting from cache")
+            return results[key]
+        else:
+            print("Calculating new result")
+            result = func(*args)
+            results[key] = result
+            return result
+
+    return wrapper


app/main.py
```diff
@@ -1,3 +1,18 @@
...
+
+    def wrapper(*args) -> Any:
+        key = args
+        if key in results:
+            print("Getting from cache")
+            return results[key]
+        else:
+            print("Calculating new result")
+            result = func(*args)
+            results[key] = result
```
ArsenPidhoretskyi, line 15: Write this inline

ArsenPidhoretskyi (changes requested):

app/main.py
stepband: I rewrote it according to the instructions

app/main.py
```diff
@@ -1,3 +1,18 @@
...
+            return result
```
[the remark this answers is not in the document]
stepband: Moved after the if-else block.

app/main.py
```diff
@@ -1,3 +1,18 @@
...
+
+    def wrapper(*args) -> Any:
+        key = args
+        if key in results:
+            print("Getting from cache")
+            return results[key]
+        else:
+            print("Calculating new result")
+
+        results[key] = func(*args)
```
```suggestion
            print("Calculating new result")
            results[key] = func(*args)
```

ArsenPidhoretskyi: After this return will be the same in if you can remove it

ArsenPidhoretskyi (changes requested):

app/main.py
```diff
@@ -1,3 +1,17 @@
...
-    # Write your code here
-    pass
+from typing import Any, Callable
+
+
+def cache(func: Callable) -> Callable:
+    results = {}
+
+    def wrapper(*args) -> Any:
+        key = args
```
ArsenPidhoretskyi, line 8: FYI this line is redundant

ArsenPidhoretskyi (approved):

PR #539, opened 2023-05-30
'Solution'

Anastasia-Su:


app/main.py
```diff
@@ -1,3 +1,26 @@
...
+def cache(func: Callable) -> Callable:
+    results = {}
+
+    def inner(*args: Union[int, Tuple[int, float]]) -> Union[int, list]:
+        if args in results:
+            print("Getting from cache")
+            return results[args]
+        else:
+            result_value = func(*args)
+            results[args] = result_value
```
ArsenPidhoretskyi, line 13: You can write this inline and just move return after if-else statement

ArsenPidhoretskyi (changes requested):

Grayder0152 (approved):

PR #540, opened 2023-05-30
Develop5

12-deadblue:


app/main.py
```diff
@@ -1,3 +1,13 @@
...
+    results = {}
+
+    def inner(*args, **kwargs) -> callable:
+        key = args + tuple(sorted(kwargs.items()))
+        if key in results:
+            print("Getting from cache")
+        else:
+            print("Calculating new result")
+            res = func(*args, **kwargs)
+            results[key] = res
```
Oleg-s-s, line 11: Unnecessary res variable

app/main.py
```diff
@@ -1,3 +1,12 @@
-def cache(func):
-    # Write your code here
-    pass
+def cache(func: callable) -> callable:
+    results = {}
+
+    def inner(*args) -> callable:
+        key = args
```
Oleg-s-s, line 5: Why you created this variable?

Oleg-s-s (changes requested):

app/main.py
12-deadblue: for explicitly

PR #544, opened 2023-05-31

verdene0:


app/main.py
```diff
@@ -1,3 +1,19 @@
...
+def cache(func: Callable) -> Callable:
+    cash_storage = {}
+
+    def wrapper(*args) -> Any:
+        key_cash = args
+        if key_cash not in cash_storage.keys():
+            cash_storage[key_cash] = func(*args)
+            print("Calculating new result")
+
+            return cash_storage[key_cash]
```
ArsenPidhoretskyi, line 13: No need to duplicate move return after if-else statement

ArsenPidhoretskyi (changes requested):

vasylhnatiuk (approved):

PR #545, opened 2023-06-02
Solution

MarkShapka:


app/main.py
```diff
@@ -1,3 +1,17 @@
...
-    pass
+from typing import Callable, Any
+
+dictionary = {}
+
+
+def cache(func: Callable) -> Any:
+
+    def inner(*args, **kwargs) -> Any:
+        key = func, args
```
ArsenPidhoretskyi, line 9: No need to store as key func
Only args.

app/main.py
```diff
@@ -1,3 +1,17 @@
...
+        if key in dictionary:
+            print("Getting from cache")
+            return dictionary[key]
+        result = func(*args, **kwargs)
+        dictionary[key] = result
```
ArsenPidhoretskyi, line 14: Do this inline

ArsenPidhoretskyi (changes requested):

Bohdan-Salamakha (changes requested):

Bohdan-Salamakha (approved):

PR #546, opened 2023-06-03

vkleshko:


app/main.py
```diff
@@ -1,3 +1,18 @@
-def cache(func):
-    # Write your code here
-    pass
+from typing import Callable, Any
+
+
+def cache(func: Callable[..., Any]) -> Callable[..., Any]:
```
vsmutok, line 4:
```suggestion
def cache(func: Callable) -> Callable:
```

app/main.py
```diff
@@ -1,3 +1,18 @@
...
+    memo = {}
```
vsmutok, line 5: I thin `memo` is bad variable name
also `res` bad name

app/main.py
```diff
@@ -1,3 +1,18 @@
...
+    def wrapper(*args, **kwargs) -> Any:
+        key = (args, tuple(kwargs))
+        if key in memo:
+            print("Getting from cache")
+            return memo[key]
+        else:
+            print("Calculating new result")
+            res = func(*args, **kwargs)
+            memo[key] = res
+            return res
```
```suggestion
        if key in memo:
            print("Getting from cache")
        else:
            print("Calculating new result")
            memo[key] = func(*args, **kwargs)
        return memo[key]
```

vsmutok (changes requested):

PR #549, opened 2023-06-09
Solution

Starchild666:


app/main.py
```diff
@@ -1,3 +1,16 @@
...
+    data_collection = {}
+
+    def inner(*args) -> object:
+        if args in data_collection.keys():
+            print("Getting from cache")
+            return data_collection[args]
+        else:
+            print("Calculating new result")
+            data = func(*args)
+            data_collection[args] = data
```
ArsenPidhoretskyi, line 14: write this inline

ArsenPidhoretskyi: And then you will have similar code inside if and else

ArsenPidhoretskyi (changes requested):

app/main.py
```diff
@@ -1,3 +1,14 @@
-def cache(func):
-    # Write your code here
-    pass
+from typing import Callable
+
+
+def cache(func: Callable) -> Callable:
+    data_collection = {}
+
```
vsmutok, line 6: use @functools.wraps

vsmutok (changes requested):

ArsenPidhoretskyi (approved):
Well done!

PR #551, opened 2023-06-11
Solution

NazarLytvyniuk:


app/main.py
```diff
@@ -1,3 +1,16 @@
...
-    # Write your code here
-    pass
+from typing import Callable, Any
+
+
+def cache(func: Callable) -> Callable:
+    results = {}
+
+    def inner(*args) -> Any:
+        if args in results:
```
Dimosphen1, line 8: Great work here! As an alternative implementation, you may use [assignment expression](https://peps.python.org/pep-0572/) in the if statement that will allow you to check if the value is truthy and return the already retrieved value in a variable. This will not require repeating this operation in the return statement and you can omit using `args in results` check as well which will lead to better optimization.

Dimosphen1 (commented):
Great implementation of the decorator pattern in Python, works as intended.

Dimosphen1 (approved):

app/main.py
NazarLytvyniuk: Thanks, I applied it

PR #554, opened 2023-06-12
solution

nschcur:


app/main.py
```diff
@@ -1,3 +1,17 @@
...
+    results = {}
+
+    def wrapper(*args: Any) -> Any:
+        if args in results:
+            print("Getting from cache")
+            return results[args]
+        else:
+            print("Calculating new result")
+            result = func(*args)
+            results[args] = result
```
ArsenPidhoretskyi, line 14: Write this inline. And then move return after if-else

ArsenPidhoretskyi (changes requested):

app/main.py
```diff
@@ -1,3 +1,15 @@
...
+
+    def wrapper(*args: Any) -> Any:
+        if args in results:
+            print("Getting from cache")
+            return results[args]
+        else:
+            print("Calculating new result")
+        return results.setdefault(args, func(*args))
+
+    return wrapper
```
Grayder0152, line 15: add new line after return

Grayder0152 (approved):

PR #557, opened 2023-06-14
created cache decorator

vsafian:


app/main.py
```diff
@@ -1,3 +1,17 @@
...
-    # Write your code here
-    pass
+from typing import Callable
+
+
+def cache(func: Callable) -> Callable:
+    cash_dict = {}
+
+    def function_cacher(*args, **kwargs) -> Callable:
+        nonlocal cash_dict
```
ArsenPidhoretskyi, line 8: This is not needed)

app/main.py
```diff
@@ -1,3 +1,17 @@
...
+        args_kvargs_key = f"{args} {kwargs}"
```
ArsenPidhoretskyi, line 9: No need to do like this. Function only accepts *args

ArsenPidhoretskyi (changes requested):

app/main.py
vasylhnatiuk: remove it

vasylhnatiuk (changes requested):

vasylhnatiuk (changes requested):
remove it

vasylhnatiuk (approved):

PR #558, opened 2023-06-15
'Solution'

Rudge0:


app/main.py
```diff
@@ -1,3 +1,22 @@
-def cache(func):
-    # Write your code here
-    pass
+from typing import Callable, Any
+
+
+def cache(func: Callable) -> Callable:
+    data = {}
+    result_history = {}
```
Grayder0152, line 6: pls use only one dict for this task

Grayder0152 (changes requested):

app/main.py
Rudge0: done

vasylhnatiuk (changes requested):

vasylhnatiuk (changes requested):
use only ONE return

vasylhnatiuk (approved):

PR #563, opened 2023-06-19
Solution

pnakongit:


app/main.py
```diff
@@ -1,3 +1,19 @@
...
+
+
+def cache(func: Callable) -> Callable:
+    _cache = {}
+
+    @wraps(func)
+    def wrapper(*args: not MutableSequence) -> Any:
+        if args in _cache:
+            print("Getting from cache")
+            return _cache[args]
```
vasylhnatiuk, line 12: use else and one return instead of two return

vasylhnatiuk (changes requested):

Grayder0152 (approved):

PR #565, opened 2023-06-20
Solution

Quiet-Klirik:


app/main.py
```diff
@@ -1,3 +1,20 @@
-def cache(func):
-    # Write your code here
-    pass
+from typing import Callable, Any
+
+
+def cache(func: Callable) -> Callable:
+    _cache = dict()
+
```
Polyakiv-Andrey, line 6: Here you should use the decarator from functools

app/main.py
```diff
@@ -1,3 +1,20 @@
-def cache(func):
-    # Write your code here
-    pass
+from typing import Callable, Any
+
+
+def cache(func: Callable) -> Callable:
+    _cache = dict()
```
Polyakiv-Andrey, line 5: Since your function is called c cache, you use the _cache variable as a dictionary, it is better to rename it to cache_dict, the underscore in python is used for encapsulation, you will learn about this later

Polyakiv-Andrey (changes requested):

Dimosphen1 (approved):
LGTM, nice use of `wraps` decorator!

app/main.py
```diff
@@ -1,3 +1,22 @@
-def cache(func):
-    # Write your code here
-    pass
+from typing import Callable, Any
+from functools import wraps
```
Dimosphen1, line 2: functools import should precede typing

PR #571, opened 2023-06-24
Solution

fotomassimo:


app/main.py
```diff
@@ -1,3 +1,20 @@
...
+from typing import Callable, Any
+
+
+def cache(func: Callable) -> Callable:
+    storage = {}
+
+    def inner(*args) -> Any:
+
+        func_arguments_and_results = storage.get(func)
+        if func in storage and args in func_arguments_and_results:
```
Nattalli, line 10: I think that this check: `func in storage` is unused, you can just remove it

Nattalli: instead of `args in func_arguments_and_results` you can just use `args in storage`

app/main.py
```diff
@@ -1,3 +1,20 @@
...
+            print("Getting from cache")
+            res = func_arguments_and_results.get(args)
+        else:
+            res = func(*args)
+            print("Calculating new result")
+            func_arguments_and_results.update({args: res}) if func in storage \
+                else storage.update({func: {args: res}})
```
Nattalli, line 17: it is also can be just `storage[args] = res`

Nattalli (changes requested):

Nattalli (approved):

PR #575, opened 2023-06-29
Implemented cache decorator

lubart2517:


app/main.py
```diff
@@ -1,3 +1,25 @@
...
+from functools import wraps
+
+
+def cache(func: Callable) -> Callable:
+    launch_cache = dict()
+
+    @wraps(func)
+    def wrapper(*args, **kwargs) -> None:
+        key = str(args) + str(kwargs)
+        if func.__name__ in launch_cache:
```
Polyakiv-Andrey, line 11: try rewriting the code without checking func.__name__

Polyakiv-Andrey (changes requested):

app/main.py
```diff
@@ -1,3 +1,25 @@
...
+                print("Getting from cache")
+                return launch_cache[func][key]
+            else:
+                print("Calculating new result")
+                launch_cache[func][key] = func(*args, **kwargs)
+                return launch_cache[func][key]
+        else:
+            launch_cache[func] = dict()
+            print("Calculating new result")
+            launch_cache[func][key] = func(*args, **kwargs)
```
Dimosphen1, line 22: `cache` function should not behave as Singleton, storing function names it was ever applied to with their arguments and results. It is applied to one function at a time, which is why storing function names and nesting input values and results is excessive. Rewrite `cache` with the use of only one if statement and do not store the function name in `launch_cache`. At the same time, the task describes that only `args` will be passed to a function, so you may not process `kwargs` here. As a hint, you can implement the `launch_cache` dict with keys being function args and values being calculation results. By the presence of similar keys, you can determine whether these args were ever processed.

Dimosphen1 (changes requested):
Change in the `cache` function logic is requested.

Polyakiv-Andrey (approved):
Good job!

PR #577, opened 2023-06-30
Solution cache decorator

OlPlaksa:


app/main.py
```diff
@@ -1,3 +1,15 @@
...
+    cache_data = {}
+
+    def wrapper(*args: None) -> None:
+        if args in cache_data:
+            print("Getting from cache")
+            return cache_data[args]
+        else:
+            print("Calculating new result")
+            cache_data[args] = func(*args)
+            return cache_data[args]
```
ArsenPidhoretskyi, line 14: Move this after if-else statement

ArsenPidhoretskyi: Because the same code in both

ArsenPidhoretskyi (changes requested):

ArsenPidhoretskyi (changes requested):
Changes requested

app/main.py
```diff
@@ -1,3 +1,14 @@
-def cache(func):
-    # Write your code here
-    pass
+from typing import Callable
+
+
+def cache(func: Callable) -> Callable:
+    cache_data = {}
+
```
vsmutok, line 6: add `wraps`

vsmutok (changes requested):

app/main.py
vasylhnatiuk: you must add like here https://www.geeksforgeeks.org/python-functools-wraps-function/

app/main.py
```diff
@@ -1,3 +1,24 @@
...
+        if args in cache_data:
+            print("Getting from cache")
+        else:
+            print("Calculating new result")
+            cache_data[args] = func(*args)
+        return cache_data[args]
+    return wrapper
+
+
+@cache
```
vasylhnatiuk, line 17: bellow code is reduntant

vasylhnatiuk (changes requested):
add wraps https://www.geeksforgeeks.org/python-functools-wraps-function/

vsmutok (approved):
Well done 👍

PR #581, opened 2023-07-03

katryana:


app/main.py
```diff
@@ -1,3 +1,16 @@
-def cache(func):
-    # Write your code here
-    pass
+from typing import Any
+
+
+def cache(func) -> Any:
+    data_save = {}
+
+    def inner(*args: tuple[(int, float, tuple, str)]) -> Any:
```
Kroll410, line 7: If you want have a consistency with typings, do not pass `Any` for determined result

PR #583, opened 2023-07-03
Solution

alecilchenko:


app/main.py
```diff
@@ -1,3 +1,16 @@
-def cache(func):
-    # Write your code here
-    pass
+from typing import Callable
+
+
+def cache(func: Callable) -> Callable:
+    result_args_dict = {}
+
+    def inner(*args) -> None:
```
pavlejviki, line 7: wrong return type annotation.

app/main.py
```diff
@@ -1,3 +1,16 @@
...
+        if args in result_args_dict:
+            print("Getting from cache")
+            return result_args_dict[args]
+        else:
+            value = func(*args)
```
pavlejviki, line 12: no need to create additional variable

pavlejviki (changes requested):

app/main.py
```diff
@@ -1,3 +1,15 @@
...
+from typing import Callable
+
+
+def cache(func: Callable) -> Callable:
+    result_args_dict = {}
+
+    def inner(*args) -> Callable:
+        if args in result_args_dict:
+            print("Getting from cache")
+            return result_args_dict[args]
```
pavlejviki, line 10:
```suggestion

```

app/main.py
```diff
@@ -1,3 +1,15 @@
...
+        else:
+            result_args_dict[args] = func(*args)
+            print("Calculating new result")
+            return result_args_dict[args]
```
pavlejviki, line 14:
```suggestion
    return result_args_dict[args]
```

pavlejviki (changes requested):

app/main.py
```diff
@@ -1,3 +1,15 @@
...
+    result_args_dict = {}
+
+    def inner(*args) -> Callable:
+        if args in result_args_dict:
+            print("Getting from cache")
+        else:
+            result_args_dict[args] = func(*args)
+            print("Calculating new result")
+            return result_args_dict[args]
+        return result_args_dict[args]
```
Dimosphen1, line 14: Add a code suggestion made by @pavlejviki, and remove line 13 from your code. There is no need for code duplication for these two return statements.

app/main.py
```diff
@@ -1,3 +1,15 @@
-def cache(func):
-    # Write your code here
-    pass
+from typing import Callable
+
+
+def cache(func: Callable) -> Callable:
+    result_args_dict = {}
+
+    def inner(*args) -> Callable:
```
Dimosphen1, line 7: The return type of inner is not `Callable`, it would be a result of function calculations that could be `Any` by its type.

Dimosphen1, line 7: You could consider using decorator from `functools` for `inner` to preserve the name and docstring of a function this decorator will be applied to.

Dimosphen1 (changes requested):

PR #589, opened 2023-07-11
Solution

H1kki:


app/main.py
```diff
@@ -1,3 +1,14 @@
-def cache(func):
-    # Write your code here
-    pass
+def cache(func: callable) -> callable:
+    results = {}
+
+    def wrapper(*args, **kwargs) -> callable:
```
Dimosphen1, line 4: There is no need to process `**kwargs` in this decorator, limit processing to just `*args`

app/main.py
```diff
@@ -1,3 +1,14 @@
...
+        key = args + tuple(sorted(kwargs.items()))
+        if key in results:
+            print("Getting from cache")
+            return results[key]
+        else:
```
Dimosphen1, line 9: Delete this `else` statement, code after the `return` statement will not execute anyway.

app/main.py
Dimosphen1, line 4: The output of `wrapper` may be of any type, do not limit it to callable.

app/main.py
```diff
@@ -1,3 +1,14 @@
-def cache(func):
-    # Write your code here
-    pass
+def cache(func: callable) -> callable:
```
Dimosphen1, line 1: For type-hinting use types imported from the `typing` module.

app/main.py
Dimosphen1, line 4: You may consider using a decorator from `functools` module to the `wrapper` function as well.

Dimosphen1 (changes requested):
Requested changes to type-hinting and logic.

app/main.py
```diff
@@ -1,3 +1,18 @@
...
-    pass
+from typing import Callable, Any
+from functools import wraps
+
+
+def cache(func: Callable) -> Callable:
+    results = {}
+
+    @wraps(func)
+    def wrapper(*args) -> Any:
```
Polyakiv-Andrey, line 9: if the function accepts args, then it must also accept kwargs

Polyakiv-Andrey (changes requested):

PR #590, opened 2023-07-12
Implemented cache decorator

RPodolets:


maxkatkalov (approved):
nice

app/main.py
```diff
@@ -1,3 +1,16 @@
...
-    pass
+from typing import Callable
+
+
+def cache(func: Callable) -> Callable:
+    cached_calcs = {}
+
+    def wrapper(*args, **kwargs) -> int | float:
+        if args not in cached_calcs:
+            print("Calculating new result")
```
Kroll410, line 9: Remove the print

app/main.py
```diff
@@ -1,3 +1,16 @@
-def cache(func):
-    # Write your code here
-    pass
+from typing import Callable
+
+
+def cache(func: Callable) -> Callable:
+    cached_calcs = {}
+
+    def wrapper(*args, **kwargs) -> int | float:
```
Kroll410, line 7: You can't be sure about typings here, so I'd suggest make it generic or omit them

Kroll410 (changes requested):

app/main.py
RPodolets: The task tells that prints should be.  Tests fails whiout them

Kroll410 (approved):

PR #595, opened 2023-07-17
'Solution'

Lebwa1337:


app/main.py
```diff
@@ -1,3 +1,12 @@
-def cache(func):
-    # Write your code here
-    pass
+def cache(func: tuple) -> str:
+    archive = {}
+
+    def inner(*args, **kwargs) -> None:
```
Polyakiv-Andrey, line 4: Your function returns not None, but Callable

Polyakiv-Andrey (changes requested):

app/main.py
```diff
@@ -1,3 +1,15 @@
...
+
+    def inner(*args, **kwargs) -> Callable:
+        if args not in archive:
+            result = func(*args, **kwargs)
+            archive[args] = result
+            print("Calculating new result")
+        else:
+            print("Getting from cache")
+        return archive[args]
+    return inner
```
Dimosphen1, line 15: Regarding type annotation, the previous comment was not implemented in a full manner. `cache` function should return `Callable` because we will return the `inner` function, this is the only option, not `str`. At the same time, the `inner` function should [return](https://docs.python.org/3/library/typing.html#typing.Any) `Any` from `typing`, because the result may differ depending on the input and we cannot foresee all possible options.

Dimosphen1 (changes requested):
Several minor changes were requested regarding typing.

app/main.py
```diff
@@ -1,3 +1,15 @@
-def cache(func):
-    # Write your code here
-    pass
+from typing import Callable, Any
+
+
+def cache(func: tuple) -> Callable:
```
Dimosphen1, line 4: Also one addition here, func that is passed as an argument to `cache` should be not `tuple` but `Callable` as well. Pay attention to typing for incoming arguments as well.

Dimosphen1 (approved):
LGTM, nice work!

PR #596, opened 2023-07-18
Solution

ip-ilya:


app/main.py
```diff
@@ -1,3 +1,19 @@
-def cache(func):
-    # Write your code here
-    pass
+from typing import Callable, Any
+
+
+def cache(func: Callable) -> Callable:
+    cache_dict = {}
+
+    def inner(*args) -> Any:
```
Polyakiv-Andrey, line 7: Your function returns not any, but Callable

app/main.py
```diff
@@ -1,3 +1,19 @@
...
+        nonlocal cache_dict
```
Polyakiv-Andrey, line 8: nonlocal in this task should be omitted

Polyakiv-Andrey (changes requested):

app/main.py
ip-ilya: but wht not `Any ` if  `return` cache_dict[args]` from inner can be int, list, str, etc. (depends on the type of *args)

Polyakiv-Andrey (approved):
Good job!

PR #598, opened 2023-07-23
'Solution'

niksgawrilenko:


app/main.py
```diff
@@ -1,3 +1,16 @@
...
+def cache(func: None) -> None:
+    func_results = {}
+
+    def wrapper(*args, **kwargs) -> None:
+        key = (args, frozenset(kwargs.items()))
+
+        if key in func_results:
+            print("Getting from cache")
+            return func_results[key]
+        else:
```
Dimosphen1, line 10: There is no need for `else`, the code will not execute after the `return` statement.

app/main.py
```diff
@@ -1,3 +1,16 @@
-def cache(func):
-    # Write your code here
-    pass
+def cache(func: None) -> None:
+    func_results = {}
+
+    def wrapper(*args, **kwargs) -> None:
```
Dimosphen1, line 4: You could limit processing to `args` only, there is no need to process `kwargs` for this task.

Dimosphen1, line 4: You could consider adding a decorator from `functools` for `wrapper`.

app/main.py
```diff
@@ -1,3 +1,16 @@
-def cache(func):
-    # Write your code here
-    pass
+def cache(func: None) -> None:
```
Dimosphen1, line 1: You should change typing for `cache`, because `func` will be `Callable`, as well as the returning `wrapper` will be `Callable`.

app/main.py
Dimosphen1, line 4: `wrapper` won't return None, it will return `Any` value according to the function result.

Dimosphen1 (changes requested):

app/main.py
niksgawrilenko: done

niksgawrilenko: done

niksgawrilenko: done

app/main.py
niksgawrilenko: done

app/main.py
```diff
@@ -1,3 +1,21 @@
...
+from typing import Any, Callable
+from functools import wraps
+
+
+def cache(func: Callable) -> Callable:
+    func_results = {}
+
+    @wraps(func)
+    def wrapper(*args) -> Any:
+        key = args
```
pavlejviki, line 10: It's not that essential to create an additional variable to store args, you can straight  use args for your check:
**if args in func_results:**
But the rest looks good.

pavlejviki (approved):
Well done!

PR #599, opened 2023-07-26
Solution-py-cache-decorator

Nadiia-developer:


app/main.py
```diff
@@ -1,3 +1,29 @@
...
-    pass
+import hashlib
+import json
+from typing import Callable, Any
+
+
+def cache(func: Callable) -> Callable:
+    cache = {}
+
+    def compute_key(args: tuple, kwargs: dict) -> str:
```
Polyakiv-Andrey, line 9: 1) no additional functions other than the main one should be written inside the decorator
2) You don't need this function to solve this problem

app/main.py
```diff
@@ -1,3 +1,29 @@
...
+        key = hashlib.md5(
+            json.dumps(args, sort_keys=True).encode("utf-8")
+        ).hexdigest()
+        key += hashlib.md5(
+            json.dumps(kwargs, sort_keys=True).encode("utf-8")
+        ).hexdigest()
+        return key
+
+    def wrapper(*args, **kwargs) -> Any:
```
Polyakiv-Andrey, line 18: use the wrapper from functools to return the wrapped function's metadata

app/main.py
```diff
@@ -1,3 +1,29 @@
...
+        key = compute_key(args, kwargs)
+        if key in cache:
```
Polyakiv-Andrey, line 20: try  **if args not in cache:**  instead  **key = compute_key(args, kwargs), if key in cache:**

Polyakiv-Andrey (changes requested):

PR #600, opened 2023-07-27
Solution

olgierrd:


app/main.py
```diff
@@ -1,3 +1,17 @@
-def cache(func):
-    # Write your code here
-    pass
+from typing import Any, Callable
+
+
+def cache(func: Callable) -> Any:
+    dict_of_results = {}
+
+    def inner(*args: Any, **kwargs: Any) -> Any:
```
Dimosphen1, line 7: There is no need to process `kwargs` for this task, you can omit that.

app/main.py
```diff
@@ -1,3 +1,17 @@
...
+        func_arguments = args
```
Dimosphen1, line 8: Assigning one variable to another is unreasonable, just use `args` directly.

app/main.py
```diff
@@ -1,3 +1,17 @@
...
+        if func_arguments not in dict_of_results:
+            print("Calculating new result")
+            dict_of_results[func_arguments] = func(*args, **kwargs)
+            return dict_of_results[func_arguments]
+        else:
```
Dimosphen1, line 13: There is no need in this `else` statement, code will not execute after `return` anyway.

app/main.py
```diff
@@ -1,3 +1,17 @@
-def cache(func):
-    # Write your code here
-    pass
+from typing import Any, Callable
+
+
+def cache(func: Callable) -> Any:
```
Dimosphen1, line 4: `cache` will rather return the `inner` function, which is `Callable`. You could be more specific here.

app/main.py
Dimosphen1, line 7: You could use a decorator from `functools` for `inner` here.

Dimosphen1 (changes requested):

app/main.py
```diff
@@ -1,3 +1,17 @@
...
-    pass
+from typing import Any, Callable
+from functools import wraps
+
+
+def cache(func: Callable) -> Callable:
+    dict_of_results = {}
+
+    @wraps(func)
+    def inner(*args: Any) -> Any:
```
Polyakiv-Andrey, line 9: args and kwargs don't need type annotations. And add kwargs here

Polyakiv-Andrey (changes requested):

Dimosphen1 (approved):
LGTM, nice work!

PR #603, opened 2023-07-28
solution

vkleshko:


app/main.py
```diff
@@ -1,3 +1,16 @@
-def cache(func):
-    # Write your code here
-    pass
+from typing import Callable, Any
+
+
+def cache(func: Callable) -> Callable:
+    cache = {}
```
ArsenPidhoretskyi, line 5: Better to use another name, because the function named the same

ArsenPidhoretskyi: Maybe cache_info?

ArsenPidhoretskyi (changes requested):

PR #604, opened 2023-07-29
Mission accomplished, attempt number 1

DaniilHinzburh:


app/main.py
```diff
@@ -1,3 +1,18 @@
-def cache(func):
-    # Write your code here
-    pass
+from typing import Callable
+
+
+def cache(func: Callable) -> Callable:
+    results = {}
+
+    def wrapper(*args, **kwargs) -> Callable:
```
vasylhnatiuk, line 7:
```suggestion
    def wrapper(*args) -> Callable:
```

app/main.py
```diff
@@ -1,3 +1,18 @@
...
+        key = str(args) + str(kwargs)
+        if key in results:
```
vasylhnatiuk, line 9:
```suggestion
        if  args in results:
```

app/main.py
```diff
@@ -1,3 +1,18 @@
...
-    # Write your code here
-    pass
+from typing import Callable
+
+
+def cache(func: Callable) -> Callable:
+    results = {}
+
+    def wrapper(*args, **kwargs) -> Callable:
+        key = str(args) + str(kwargs)
```
vasylhnatiuk, line 8: delete this line

vasylhnatiuk (changes requested):
try to use only one return results[args]

app/main.py
```diff
@@ -1,3 +1,17 @@
...
-    # Write your code here
-    pass
+from typing import Callable
+
+
+def cache(func: Callable) -> Callable:
+    results = {}
+
+    def wrapper(*args) -> Callable:
+        if str(args) in results:
```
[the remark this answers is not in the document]
vasylhnatiuk: why do you use str()?

app/main.py
```diff
@@ -1,3 +1,17 @@
...
+            print("Getting from cache")
+            return results[str(args)]
+        else:
+            print("Calculating new result")
+            result = func(*args)
+            results[str(args)] = result
+            return result
```
vasylhnatiuk, line 15:
```suggestion
        return  results[args]
```

vasylhnatiuk: use only one return

vasylhnatiuk (changes requested):

PR #605, opened 2023-07-30
Created decorator cache

KolBohdan:


app/main.py
```diff
@@ -1,3 +1,16 @@
...
+
+
+def cache(func: Callable) -> Any:
+    memory = {}
+
+    def inner(*args) -> str:
+        if args in memory:
+            print("Getting from cache")
+            return memory.get(args)
+        else:
```
Dimosphen1, line 11: There is no need in this `else` statement, code will not execute after `return`. Delete it.

app/main.py
```diff
@@ -1,3 +1,16 @@
-def cache(func):
-    # Write your code here
-    pass
+from typing import Callable, Any
+
+
+def cache(func: Callable) -> Any:
+    memory = {}
+
+    def inner(*args) -> str:
```
Dimosphen1, line 7: You could consider using a decorator from `functools` for `inner`.

Dimosphen1 (changes requested):

app/main.py
Dimosphen1: The return type of `str` inner could be not only `str`, actually it can be any type.

app/main.py
```diff
@@ -1,3 +1,17 @@
-def cache(func):
-    # Write your code here
-    pass
+from typing import Callable, Any
+import functools
+
+
+def cache(func: Callable) -> Any:
```
Dimosphen1, line 5: `cache` return type annotation should be `Callable`, not `Any`, because `inner` is always callable.

app/main.py
```diff
@@ -1,3 +1,17 @@
-def cache(func):
-    # Write your code here
-    pass
+from typing import Callable, Any
+import functools
```
Dimosphen1, line 2: `import` should precede `from ...`

app/main.py
```diff
@@ -1,3 +1,17 @@
...
+
+
+def cache(func: Callable) -> Any:
+    memory = {}
+
+    @functools.wraps(func)
+    def inner(*args) -> Any:
```
Dimosphen1, line 9: Add here `**kwargs` and pass them to the `func` as well. Their processing is not needed for this task, but presence would be preferable.

Dimosphen1 (changes requested):

PR #610, opened 2023-08-02
Solution

VolodymyrSeniv:
I had a problem with the last test which is "test_unnecessary_comment". It shows me that there is no such a file as main.py. I don't know how to fix that

### DIFF
--- a/app/main.py
+++ b/app/main.py
@@ -1,3 +1,16 @@
-def cache(func):
-    # Write your code here
-    pass
+from typing import Callable
+
+
+def cache(func: Callable) -> Callable:
+    cached_results = {}
+
+    def wrapper(*args, **kwargs) -> list:
+        args_key = args + tuple(kwargs.items())
+        if args_key in cached_results:
+            print("Getting from cache")
+            return cached_results[args_key]
+        result = func(*args, **kwargs)
+        cached_results[args_key] = result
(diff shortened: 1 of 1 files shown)


app/main.py
```diff
@@ -1,3 +1,16 @@
-def cache(func):
-    # Write your code here
-    pass
+from typing import Callable
+
+
+def cache(func: Callable) -> Callable:
+    cached_results = {}
+
+    def wrapper(*args, **kwargs) -> list:
```
Bohdan-Salamakha, line 7: add additional typehint for list inner objects like list[YOUR_VALUE_TYPE]
for example: list that contains string elements has typehint list[str

Bohdan-Salamakha (changes requested):

app/main.py
```diff
@@ -1,3 +1,19 @@
-def cache(func):
-    # Write your code here
-    pass
+from typing import Callable, TypeVar
+
+
+T = TypeVar("T")
```
Bohdan-Salamakha, line 4: what is it???
remove this redundant cryptic row

Bohdan-Salamakha (changes requested):

app/main.py
VolodymyrSeniv: so adding "-> list[int]" is anough?

PR #613, opened 2023-08-03
implemented decorator cache

innyshka:


app/main.py
```diff
@@ -1,3 +1,16 @@
-def cache(func):
-    # Write your code here
-    pass
+from typing import Callable, Any
+
+
+def cache(func: Callable) -> Any:
```
eduardhabryd, line 4: This function returns Callable. I think it's better to specify it: def cache(func: Callable) -> Callable

app/main.py
```diff
@@ -1,3 +1,16 @@
...
+    saved_cache_dict = {}
+
+    def inner(*args) -> Any:
```
eduardhabryd, line 7: This function returns int

eduardhabryd (changes requested):
Missing proper function annotations for cache and inner functions. Everything else seem to be fine. Nice Work!

Kroll410 (approved):

PR #614, opened 2023-08-03
Solution

eduardhabryd:


anna-lybid (approved):

app/main.py
```diff
@@ -1,3 +1,19 @@
...
-    # Write your code here
-    pass
+from typing import Callable
+
+
+def cache(func: Callable) -> Callable:
+    cached = []
+
+    def wrapper(*args, **kwargs) -> int:
+        cache_tuple = (*args, frozenset(kwargs.items()))
```
Kroll410, line 8: Why do we freeze ordering from kwargs.items()?

Kroll410 (changes requested):

app/main.py
eduardhabryd: I wanted to prevent accidental deletion of data in future but now I see that there is no need to use it in that place.

eduardhabryd: > Why do we freeze ordering from kwargs.items()?

That's because if we use regular set, the hash will depend on the order of kwargs, on the other side if we use frozenset the order of elemenent won't matter and won't influence on hash. So we will be able to check or retrieve information from our cached list correctly.

eduardhabryd: > Okay but how the order of the cached elements make an impact here?

When caching results based on the arguments (both positional and keyword arguments) of a function, you need a way to generate a unique and consistent hash for these arguments so that you can efficiently retrieve cached results. Using a simple set to store the arguments wouldn't be sufficient because sets are unordered collections, and their hash values could change based on the order of elements. This means that if you were to call the same function with the same arguments in a different order, you might not be able to retrieve the cached result, even though the actual computation is the same.

By using a frozenset, you're creating an immutable and hashable object that represents the set of keyword arguments, without being affected by their order. This ensures that no matter in which order the keyword arguments were passed, the resulting frozenset will have the same hash value. As a result, you can reliably use this hash value as part of your caching mechanism to retrieve the cached result associated with those specific keyword arguments, regardless of their order.

In conclusion, using a frozenset eliminates any potential inconsistency caused by the order of the keyword arguments and ensures that the caching mechanism works as expected in all scenarios.

Kroll410: Okay, maybe I don't quiet understand here, because you replied me a doc for frozenset 
We're caching this key
image
But should we really do so? Also, if applying frozenset to dict_items object makes sense in terms of obtaining the correct and cached key, can you give me an example?

Kroll410 (changes requested):

eduardhabryd:
Imagine you have the following function that calculates the sum of two numbers and caches the results:

`
@cache
def add(a, b):
    print("Performing addition")
    return a + b
`

Now, let's consider two calls to this function:

`
print(add(2, 3))
print(add(b=3, a=2))
`

In both cases, the same addition operation is being performed, resulting in the value 5. However, the order of the keyword arguments is different in these two calls.

If you were using a regular set to create the cache key, the hash value for the dict_items representing the keyword arguments might vary depending on their order. As a result, these two calls might generate different cache keys, and you wouldn't be able to retrieve the cached result for the second call, even though the computation is the same.

eduardhabryd:
> you can use dict to store data with your cach_tuple as a key, for what purpose do you use list as storage?

You're right. Fixed, please check again.

app/main.py
```diff
@@ -1,3 +1,19 @@
...
+
+def cache(func: Callable) -> Callable:
+    cached = dict()
+
+    def wrapper(*args, **kwargs) -> int:
+        cache_tuple = (*args, frozenset(kwargs.items()))
+        for cached_args, cached_result in cached.items():
+            if cached_args == cache_tuple:
+                print("Getting from cache")
+                return cached_result
```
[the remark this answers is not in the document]
eduardhabryd: > you don't need loop here, just use `cached.get(cache_tuple)`

Fixed, please check.

PR #615, opened 2023-08-03
Implementation of py-cache-decorator

anna-lybid:


Barsh4ec (approved):

app/main.py
```diff
@@ -1,3 +1,15 @@
-def cache(func):
-    # Write your code here
-    pass
+def cache(func: callable) -> callable:
+    dict_cache = {}
+
+    def inner(*args, **kwargs) -> int:
+        key = (args, frozenset(kwargs.items()))
```
Kroll410, line 5: Why do we freeze kwargs ordering here?

Kroll410 (changes requested):

app/main.py
anna-lybid: > Why do we freeze kwargs ordering here?

Just one way to unpack the kwargs to keep them organized. I can change that to tuple(). It would be better?

app/main.py
```diff
@@ -1,3 +1,15 @@
-def cache(func):
-    # Write your code here
-    pass
+def cache(func: callable) -> callable:
+    dict_cache = {}
+
+    def inner(*args, **kwargs) -> int:
+        key = (args, tuple(kwargs.items()))
```
Kroll410, line 5: Check your key in debugger
image

Kroll410 (changes requested):

app/main.py
```diff
@@ -1,3 +1,14 @@
...
-    # Write your code here
-    pass
+def cache(func: callable) -> callable:
+    dict_cache = {}
+
+    def inner(*args, **kwargs) -> str:
+        if args in dict_cache:
+            print("Getting from cache")
+            return dict_cache[args]
+        else:
```
IvanRamyk, line 8: redundant else

IvanRamyk (changes requested):
everything else is fine, great job!

IvanRamyk (approved):

PR #616, opened 2023-08-03
Solution

Barsh4ec:
implemented solution

### DIFF
--- a/app/main.py
+++ b/app/main.py
@@ -1,3 +1,14 @@
-def cache(func):
-    # Write your code here
-    pass
+def cache(func: callable) -> callable:
+    cached = {}
+
+    def inner(*args: int | float) -> int | list:
+        if args not in cached.keys():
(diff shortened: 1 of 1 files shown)


app/main.py
```diff
@@ -1,3 +1,14 @@
-def cache(func):
-    # Write your code here
-    pass
+def cache(func: callable) -> callable:
```
denys-source, line 1: incorrect type annotation. callable is a function

app/main.py
```diff
@@ -1,3 +1,14 @@
...
+    cached = {}
+
+    def inner(*args: int | float) -> int | list:
```
denys-source, line 4: arguments and return value can have any type. use `Any` or `ParamSpec` with `TypeVar` instead

OlegatorLE (approved):
Good!

Kroll410 (approved):

PR #617, opened 2023-08-03
Solution

denys-source:


lgSilay (approved):
Great to improve this task with try/except !)

app/main.py
```diff
@@ -1,3 +1,31 @@
...
+from typing import Callable, ParamSpec, TypeVar
+
+Param = ParamSpec("Param")
+RetType = TypeVar("RetType")
+
+
+def cache(func: Callable[Param, RetType]) -> Callable[Param, RetType]:
+    cache_dict: dict = {}
+
+    @wraps(func)
```
Kroll410, line 11: What the purpose using wraps here?

app/main.py
```diff
@@ -1,3 +1,31 @@
-def cache(func):
-    # Write your code here
-    pass
+from functools import wraps
+from typing import Callable, ParamSpec, TypeVar
+
+Param = ParamSpec("Param")
```
Kroll410, line 4: Specify generics more precisely, if using them

app/main.py
```diff
@@ -1,3 +1,31 @@
...
+RetType = TypeVar("RetType")
+
+
+def cache(func: Callable[Param, RetType]) -> Callable[Param, RetType]:
+    cache_dict: dict = {}
+
+    @wraps(func)
+    def wrapper(*args, **kwargs) -> RetType:
+        for arg in args:
```
Kroll410, line 13: Unify looping through the args and kwargs

app/main.py
```diff
@@ -1,3 +1,31 @@
...
+            try:
+                hash(arg)
+            except TypeError:
+                return func(*args, **kwargs)
+        for value in kwargs.values():
+            try:
+                hash(value)
```
Kroll410, line 20: you can check is this value hashable rather than calling hash func

app/main.py
```diff
@@ -1,3 +1,31 @@
...
+            except TypeError:
+                return func(*args, **kwargs)
+        signature: tuple = (args, tuple(kwargs.items()))
```
Kroll410, line 23: We're caching all the values from args and kwargs as a key?

Kroll410 (changes requested):
Try not to [complicate ](https://peps.python.org/pep-0020/#the-zen-of-python) the code :)

app/main.py
denys-source: i thought that using args and kwargs as a key would prevent returning the wrong output if there is a keyword argument that changes the function's return value

app/main.py
```diff
@@ -1,3 +1,23 @@
...
+
+def cache(func: Callable[Param, RetType]) -> Callable[Param, RetType]:
+    cache_dict: dict = {}
+
+    def wrapper(*args: Param.args, **kwargs: Param.kwargs) -> RetType:
+        all_arguments = args + tuple(kwargs.values())
+        for arg in all_arguments:
+            if not isinstance(arg, Hashable):
+                return func(*args, **kwargs)
+        signature: tuple = (args, tuple(kwargs.items()))
```
IvanRamyk, line 15: I believe this approach doesn't work for such calls:

```
def func(a, b):
  ...

func(1, 2)
func(a=1, b=2)
```
signature are different for those two cases

let's assume that only positional arguments are used

IvanRamyk (changes requested):

IvanRamyk (approved):

PR #620, opened 2023-08-04
Nikita Karpenko py-23 july

ORENYT:
Created a decorator function that hashing a result of previous functions

### DIFF
--- a/app/main.py
+++ b/app/main.py
@@ -1,3 +1,31 @@
-def cache(func):
-    # Write your code here
-    pass
+from typing import Callable
+
+
+def cache(func: Callable) -> Callable:
+    previous_runs = {}
+
+    def inner(*args) -> object:
+        nonlocal previous_runs
+        hash_key = hash_function(func, args)
+        if hash_key in previous_runs.keys():
+            print("Getting from cache")
+            return previous_runs[hash_key]
+        else:
+            print("Calculating new result")
+            result = func(*args)
+            previous_runs[hash_key] = result
+        return result
+
+    return inner
+
+
+def hash_function(function_to_hash: Callable, args: tuple):
+    """
+    Hashing incoming function, using it's name
+    and arguments with a string result.
+    No encryption.
+    """
+    name_of_hashed = str(function_to_hash.__name__)
+    hashed_args = str(args)
+    result_hash = name_of_hashed + hashed_args
+    return result_hash


MarianKovalyshyn (approved):
I think there is no point to implement 'hash_function' because you can pass 'args' as an argument in 'previous_runs' dict and don't call function twice. But in general, looks good.

ORENYT:
> I think there is no point to implement 'hash_function' because you can pass 'args' as an argument in 'previous_runs' dict and don't call function twice. But in general, looks good.

External function was created to develop it in advance, if hashing will be required to hide a name of function or arguments.

app/main.py
```diff
@@ -1,3 +1,31 @@
...
+        else:
+            print("Calculating new result")
+            result = func(*args)
+            previous_runs[hash_key] = result
+        return result
+
+    return inner
+
+
+def hash_function(function_to_hash: Callable, args: tuple) -> str:
```
Kroll410, line 22: > External function was created to develop it in advance, if hashing will be required to hide a name of function or arguments.

Good point, but we don't need it here. Unfortunately, in real work cases it might not be approved for the task description, so do I, sorry :(

Kroll410 (changes requested):

ORENYT:
Changed function to one-line, request checking.

app/main.py
ORENYT: Fixed in next commit

app/main.py
```diff
@@ -1,3 +1,19 @@
...
-    # Write your code here
-    pass
+from typing import Callable
+
+
+def cache(func: Callable) -> Callable:
+    previous_runs = {}
+
+    def inner(*args) -> object:
+        nonlocal previous_runs
```
IvanRamyk, line 8: nonlocal is redundant here because you don't use an assignment operator with this variable and only call its methods.

app/main.py
```diff
@@ -1,3 +1,19 @@
...
+        hash_key = str(func.__name__) + str(args)
```
```suggestion
        hash_key = args
```
No need to add a function name to the key because the cache dict is unique for each function. It is created while decorating.

app/main.py
```diff
@@ -1,3 +1,19 @@
...
+        if hash_key in previous_runs.keys():
+            print("Getting from cache")
+            return previous_runs[hash_key]
+        else:
```
IvanRamyk, line 13: redundant else

IvanRamyk (changes requested):

app/main.py
ORENYT: Changes were accepted.

app/main.py
ORENYT: Else removed

app/main.py
ORENYT: Nonlocal removed

app/main.py
```diff
@@ -1,3 +1,17 @@
...
-    # Write your code here
-    pass
+from typing import Callable
+
+
+def cache(func: Callable) -> Callable:
+    previous_runs = {}
+
+    def inner(*args) -> object:
+        hash_key = args
```
IvanRamyk, line 8: extra variable

IvanRamyk (changes requested):

app/main.py
ORENYT: Removed this variable

IvanRamyk (approved):

PR #625, opened 2023-08-04
Implemented Cache decorator

AndriiHomeniuk1:


app/main.py
```diff
@@ -1,3 +1,13 @@
...
+    cache_dict = {}
+
+    def wrapper(*args):
+        if args not in cache_dict:
+            cache_dict.update({args: func(*args)})
+            print("Calculating new result")
+            return cache_dict[args]
+        else:
+            print("Getting from cache")
+            return cache_dict[args]
```
vasylhnatiuk, line 11: use only one return

vasylhnatiuk (changes requested):

vasylhnatiuk (changes requested):
image

app/main.py
AndriiHomeniuk1: fixed everything)

app/main.py
```diff
@@ -1,3 +1,15 @@
...
-    pass
+from typing import Callable, Any
+
+
+def cache(func: Callable) -> Callable:
+    cache_dict = {}
+
+    def wrapper(*args) -> Any:
+        if args not in cache_dict:
+            cache_dict.update({args: func(*args)})
```
ArsenPidhoretskyi, line 9:
```suggestion
            cache_dict[args] = func(*args)
```
No need to use update

ArsenPidhoretskyi (changes requested):

app/main.py
AndriiHomeniuk1: didn't think about it. Thank you 
Fixed it

AndriiHomeniuk1: Didn't think about it. Thank you
Fixed it

ArsenPidhoretskyi (approved):
Well done!

PR #626, opened 2023-08-05
Cache_decorator

yaroslav-mazurkevych:


OleksandrMoroz24 (approved):
good code, but you add kwargs for input agrument your inner func but then you not check them in code

app/main.py
```diff
@@ -1,3 +1,16 @@
-def cache(func):
-    # Write your code here
-    pass
+from typing import Callable
+
+
+def cache(func: Callable) -> Callable:
+    dict_of_args = {}
+
+    def inner(*args, **kwargs) -> dict:
```
Kroll410, line 7: Why kwargs are not validated as args were?

Kroll410 (changes requested):

app/main.py
```diff
@@ -1,3 +1,16 @@
...
+        if args not in dict_of_args.keys():
+            dict_of_args[args] = func(*args)
+            print("Calculating new result")
+            return dict_of_args[args]
+        else:
```
IvanRamyk, line 12: redundant else

app/main.py
```diff
@@ -1,3 +1,16 @@
...
-    # Write your code here
-    pass
+from typing import Callable
+
+
+def cache(func: Callable) -> Callable:
+    dict_of_args = {}
+
+    def inner(*args, **kwargs) -> dict:
+        if args not in dict_of_args.keys():
```
IvanRamyk, line 8:
```suggestion
        if args not in dict_of_args:
```

IvanRamyk: you don't need to call keys explicitly

IvanRamyk (changes requested):

Y-Havryliv (approved):
GJ!

PR #627, opened 2023-08-05

Illia-Kononenko:


maxkatkalov (approved):
Ok, great.

app/main.py
```diff
@@ -1,3 +1,19 @@
...
+def cache(func: callable) -> callable:
+    cached_results: Dict[Tuple[str, Tuple, frozenset], Any] = {}
+
+    def wrapper(*args: Any, **kwargs: Any) -> Any:
+        key: Tuple[str, Tuple, frozenset] = (func.__name__, args,
+                                             frozenset(kwargs.items()))
+        if key in cached_results:
+            print("Getting from cache")
+            return cached_results[key]
+        else:
```
IvanRamyk, line 13: redundant else

app/main.py
```diff
@@ -1,3 +1,19 @@
...
-    # Write your code here
-    pass
+from typing import Any, Dict, Tuple
+
+
+def cache(func: callable) -> callable:
+    cached_results: Dict[Tuple[str, Tuple, frozenset], Any] = {}
+
+    def wrapper(*args: Any, **kwargs: Any) -> Any:
+        key: Tuple[str, Tuple, frozenset] = (func.__name__, args,
```
IvanRamyk, line 8: No need to add a function name to the key because the cache dict is unique for each function. It is created while decorating.

IvanRamyk (changes requested):

PR #628, opened 2023-08-05
Solution

OleksandrMoroz24:


app/main.py
```diff
@@ -1,3 +1,16 @@
...
-    # Write your code here
-    pass
+from typing import Callable
+
+
+def cache(func: Callable) -> Callable:
+
+    result_cache = {}
+
+    def wrapper(*args) -> None:
```
denys-source, line 8: add type annotation for `*args` and update return value annotation to `Any`

app/main.py
```diff
@@ -1,3 +1,16 @@
...
+        if args in result_cache:
+            print("Getting from cache")
+            return result_cache[args]
+        else:
```
denys-source, line 12: you can remove `else` if there is a `return` statement

WonnaKoo (approved):

Kroll410 (changes requested):

Kroll410 (approved):

PR #629, opened 2023-08-05
Solution

IvanKosin:


app/main.py
```diff
@@ -1,3 +1,18 @@
...
+    def wrapper(*args: Tuple, **kwargs: Dict) -> Any:
+        key: Tuple = args + tuple(kwargs.items())
+        if key in cached_results:
+            print("Getting from cache")
+            return cached_results[key]
+        else:
+            print("Calculating new result")
+            result: Any = func(*args, **kwargs)
+            cached_results[key] = result
+            return result
```
vasylhnatiuk, line 16:
```suggestion
        return cached_results[key]
```

vasylhnatiuk: use only one `return `in `def wrapper`

app/main.py
```diff
@@ -1,3 +1,18 @@
...
+
+
+def cache(func: callable) -> callable:
+    cached_results: Dict[Tuple, Any] = {}
+
+    def wrapper(*args: Tuple, **kwargs: Dict) -> Any:
+        key: Tuple = args + tuple(kwargs.items())
+        if key in cached_results:
+            print("Getting from cache")
+            return cached_results[key]
```
vasylhnatiuk, line 11: delete this

vasylhnatiuk (changes requested):

PR #632, opened 2023-08-06
Implemented decorators task

ZaichenkoViktoriia:


app/main.py
```diff
@@ -1,3 +1,17 @@
-def cache(func):
-    # Write your code here
-    pass
+from typing import Callable
+
+
+def cache(func: Callable) -> str:
```
ochernous, line 4: Maybe we return another type of data?

grungezz (approved):

app/main.py
```diff
@@ -1,3 +1,17 @@
...
+    cached_results = {}
+
+    def wrapper(*args, **kwargs) -> Callable:
+        if args in cached_results:
+            print("Getting from cache")
+            return cached_results[args]
+        else:
```
IvanRamyk, line 11: redundant else

app/main.py
```diff
@@ -1,3 +1,17 @@
-def cache(func):
-    # Write your code here
-    pass
+from typing import Callable
+
+
+def cache(func: Callable) -> str:
+    cached_results = {}
+
+    def wrapper(*args, **kwargs) -> Callable:
```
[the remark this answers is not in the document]
IvanRamyk: +1

app/main.py
IvanRamyk: agree

IvanRamyk (changes requested):

IvanRamyk (approved):

PR #633, opened 2023-08-06
Solution

Ihor-MA:


app/main.py
```diff
@@ -1,3 +1,17 @@
-def cache(func):
-    # Write your code here
-    pass
+from typing import Callable
+
+
+def cache(func: Callable) -> None:
```
ochernous, line 4: Maybe Callable be better than None?

ochernous (approved):

PR #637, opened 2023-08-10
Solution

Shum430:


app/main.py
```diff
@@ -1,3 +1,17 @@
...
+    def wrapper(*args, **kwargs) -> None:
+        arguments = args + tuple(sorted(kwargs.items()))
+        if arguments in caching_place:
+            print("Getting from cache")
+            return caching_place[arguments]
+        else:
+            print("Calculating new result")
+            result = func(*args, **kwargs)
+            caching_place[arguments] = result
+            return result
```
vasylhnatiuk, line 16:
```suggestion
        return caching_place[arguments]
```

app/main.py
```diff
@@ -1,3 +1,17 @@
...
+
+
+def cache(func: Callable) -> Callable:
+    caching_place = {}
+
+    def wrapper(*args, **kwargs) -> None:
+        arguments = args + tuple(sorted(kwargs.items()))
+        if arguments in caching_place:
+            print("Getting from cache")
+            return caching_place[arguments]
```
vasylhnatiuk, line 11: and delete this

vasylhnatiuk (changes requested):
use only one return in def wrapper

app/main.py
```diff
@@ -1,3 +1,16 @@
-def cache(func):
-    # Write your code here
-    pass
+from typing import Callable
+
+
+def cache(func: Callable) -> Callable:
+    caching_place = {}
+
```
Dimosphen1, line 6: You could consider using `wraps` decorator from `functools` here, read more about its advantages in the documentation.

app/main.py
```diff
@@ -1,3 +1,16 @@
...
+    def wrapper(*args, **kwargs) -> None:
```
Dimosphen1, line 7: The return type for `wrapper` should be not `None`, but rather `Any`, because you can apply it to any function.

Dimosphen1 (changes requested):

app/main.py
```diff
@@ -1,3 +1,18 @@
...
+
+    @wraps(func)
+    def wrapper(*args, **kwargs) -> Any:
+        arguments = args + tuple(sorted(kwargs.items()))
+        if arguments in caching_place:
+            print("Getting from cache")
+        else:
+            print("Calculating new result")
+            result = func(*args, **kwargs)
+            caching_place[arguments] = result
```
ArsenPidhoretskyi, line 16: Write this inline

ArsenPidhoretskyi (changes requested):

PR #639, opened 2023-08-12
Solution for py-cache-decorator

Viva1-ua:


app/main.py
```diff
@@ -1,3 +1,17 @@
...
+from typing import Callable, Any
+
+
+def cache(func: Callable) -> Any:
+    stored_cache = {}
+
+    def wrapper(*args: (int, float, str), **kwargs: (tuple, bool)) -> Any:
+        if args not in stored_cache:
+            result = func(*args, **kwargs)
+            stored_cache[args] = result
```
ArsenPidhoretskyi, line 10: Write this inline

ArsenPidhoretskyi: And then move return from if-else statement

ArsenPidhoretskyi (changes requested):

app/main.py
Viva1-ua: Didn't really get it. Do I need to write lines 9 and 10 in one line? And about return also didn't understand, sorry

vasylhnatiuk (approved):
Good job

PR #640, opened 2023-08-13
Solution

dkibalenko:


app/main.py
```diff
@@ -1,3 +1,17 @@
...
-    # Write your code here
-    pass
+from typing import Callable, Any
+
+
+def cache(func: Callable) -> Callable:
+    cache_data = dict()
+
+    def wrapper(*args: Any) -> int | str:
+        for argument in args:
```
pavlejviki, line 8: the task description doesn't specifically asks for the immutability check

app/main.py
```diff
@@ -1,3 +1,17 @@
-def cache(func):
-    # Write your code here
-    pass
+from typing import Callable, Any
+
+
+def cache(func: Callable) -> Callable:
+    cache_data = dict()
+
+    def wrapper(*args: Any) -> int | str:
```
pavlejviki, line 7: If you're trying to make a general caching decorator, using Any as the return type would be better

pavlejviki (changes requested):

Dimosphen1 (approved):
LGTM. Minor comments are left for future consideration.

pavlejviki (approved):
Good job!

PR #642, opened 2023-08-14
Solution

prochigor:


app/main.py
```diff
@@ -1,3 +1,18 @@
...
-    # Write your code here
-    pass
+import time
+from typing import Callable
+
+
+def cache(func: Callable) -> Callable:
+    cache_result = {}
+
+    def add_result(*args, **kwargs) -> Callable:
```
LiudmylaKulinchenko: This func can return not only Callable, it's better to use `Any`

LiudmylaKulinchenko: It is a common convention to name inner function in decorator as `inner`. Consider to rename it

app/main.py
```diff
@@ -1,3 +1,18 @@
...
+        if args in cache_result and cache_result[args][0] >= time.time():
```
LiudmylaKulinchenko: Why do you need time.time() in this solution?

app/main.py
```diff
@@ -1,3 +1,18 @@
...
+            print("Getting from cache")
+            return cache_result[args][1]
+        else:
+            result = func(*args, **kwargs)
+            cache_result[args] = (time.time() + 3, result)
+            print("Calculating new result")
+            return result
```

LiudmylaKulinchenko (changes requested):

app/main.py
prochigor: I see, I don't need that, I don't remember what I thought

fsocie7y (approved):

PR #646, opened 2023-08-15
Solution

AndriiYe:


app/main.py
```diff
@@ -1,3 +1,20 @@
...
+        if func_name not in results:
+            results[func_name] = {}
+        if args in results[func_name]:
+            print("Getting from cache")
+            return results[func_name][args]
+        else:
+            print("Calculating new result")
+            results[func_name][args] = func(*args)
+            return results[func_name][args]
+    return wrapper
```
pavlejviki, line 20: Your results dictionary can use the args tuple as a key, rather than nesting it within another dictionary keyed by function name.

pavlejviki (changes requested):

pavlejviki (approved):
Well done!

PR #649, opened 2023-08-18
Solution

Daniil-Pankieiev:


app/main.py
```diff
@@ -1,3 +1,19 @@
...
+    def wrapper(*args) -> Any:
+        key = args
+        if key in cached_results:
+            print("Getting from cache")
+            return cached_results[key]
+        else:
+            result = func(*args)
+            cached_results[key] = result
+            print("Calculating new result")
+            return result
```
LiudmylaKulinchenko: You can use a single return statement out of the if/else statement instead of two inside it.

app/main.py
```diff
@@ -1,3 +1,19 @@
...
-    pass
+from typing import Callable
+from typing import Any
+
+
+def cache(func: Callable) -> Callable:
+    cached_results = {}
+
+    def wrapper(*args) -> Any:
+        key = args
```
LiudmylaKulinchenko: Not really needed to create new variable here
Also, you can name the parameters in a function definition as you like, but it's a common convention to use `*args`. So, `args` will fit good here 👍

LiudmylaKulinchenko (changes requested):

app/main.py
Daniil-Pankieiev: Fixed. Thanks it really looks better

D1st3f:
All look like good!

D1st3f (approved):

app/main.py
```diff
@@ -1,3 +1,17 @@
...
+    cached_results = {}
+
+    def wrapper(*args) -> Any:
+        if args in cached_results:
+            print("Getting from cache")
+            return cached_results[args]
+        result = func(*args)
+        cached_results[args] = result
+        print("Calculating new result")
+        return result
```
LiudmylaKulinchenko: Try rewrite with one return

LiudmylaKulinchenko (changes requested):

wQuelS (changes requested):
The solution looks quite good, but can you implement it without using the additional intermediate variable `result` ?

Also, fix conflicts that are shown on screenshot
image

wQuelS (approved):
gj

PR #653, opened 2023-08-21
decorator is developed

YaroslavYarovoy:


app/main.py
```diff
@@ -1,3 +1,16 @@
...
+
+def cache(func: Callable) -> Callable:
+    cache_dict = {}
+
+    def inner(*args, **kwargs) -> Any:
+        if (args, tuple(kwargs.items())) in cache_dict:
+            print("Getting from cache")
+            return cache_dict[(args, tuple(kwargs.items()))]
+        else:
+            result = func(*args, **kwargs)
```
pavlejviki, line 12: no need to create additional variables

pavlejviki (changes requested):

app/main.py
```diff
@@ -1,3 +1,15 @@
...
+from typing import Callable, Any
+
+
+def cache(func: Callable) -> Callable:
+    cache_dict = {}
+
+    def inner(*args, **kwargs) -> Any:
+        if (args, tuple(kwargs.items())) in cache_dict:
+            print("Getting from cache")
+            return cache_dict[(args, tuple(kwargs.items()))]
```
pavlejviki, line 10:
```suggestion

```

app/main.py
```diff
@@ -1,3 +1,15 @@
...
+        else:
+            print("Calculating new result")
+            cache_dict[(args, tuple(kwargs.items()))] = func(*args, **kwargs)
+            return func(*args, **kwargs)
```
pavlejviki, line 14:
```suggestion
      return cache_dict[(args, tuple(kwargs.items()))]
```

app/main.py
pavlejviki, line 10: you can use just one return

pavlejviki (changes requested):

app/main.py
YaroslavYarovoy: thanks for the nice tip

PR #655, opened 2023-08-23
Py cache decorator

Dmitriy-Poplinski:


app/main.py
```diff
@@ -1,3 +1,44 @@
-def cache(func):
-    # Write your code here
-    pass
+from functools import wraps
+from typing import Callable, Any
+
+
+def cache(func: Callable) -> Callable:
+    caching_result = {}
+    wraps(func)
```
Dimosphen1, line 7: You need to apply `wraps` as a decorator for `wrapper`, you can see an example [in the documentation](https://docs.python.org/3/library/functools.html#functools.wraps).

app/main.py
```diff
@@ -1,3 +1,44 @@
...
+                "result": result
+            })
+
+            return result
+        else:
+            print("Getting from cache")
+
+            for item in caching_result[func.__name__]:
+                if item["arguments"] == args:
+                    return item["result"]
```
Dimosphen1, line 42: Your logic here is overloaded. You do not need to store the function name as a key for `caching_result`, you could use `args` directly as a key and the function result as a value. This cache decorator will be applied to one function at a time, it doesn't need to track all functions it was applied to. After that, you can check if the same args are already present in the `caching_result` and either retrieve it from there or put a new value. Besides that, try to use only one `return` statement in the `wrapper` and do not call `nonlocal caching_result`, it will be available in the `wrapper` context.

app/main.py
```diff
@@ -1,3 +1,44 @@
...
-    pass
+from functools import wraps
+from typing import Callable, Any
+
+
+def cache(func: Callable) -> Callable:
+    caching_result = {}
+    wraps(func)
+
+    def wrapper(*args: tuple) -> Any:
```
Dimosphen1, line 9: There is no need to use type annotation for `args` and `kwargs`, unless you need to specify their content. You can omit type hinting for them.

Dimosphen1 (changes requested):

app/main.py
```diff
@@ -1,3 +1,23 @@
-def cache(func):
-    # Write your code here
-    pass
+from functools import wraps
+from typing import Callable, Any
+
+
+def cache(func: Callable) -> Callable:
+    caching_result = []
```
Dimosphen1, line 6: Use `dict` for caching results, you could store `args` as key and calculation result as value. `dict` is a more preferred data structure here.

app/main.py
```diff
@@ -1,3 +1,23 @@
...
+
+    @wraps(func)
+    def wrapper(*args) -> Any:
+        for item in caching_result:
+            if args in item:
```
Dimosphen1, line 11: You could limit this expression to `if args in caching_result` in case you would use `dict`.

Dimosphen1 (changes requested):

app/main.py
```diff
@@ -1,3 +1,20 @@
...
-    pass
+from functools import wraps
+from typing import Callable
+
+
+def cache(func: Callable) -> Callable:
+    caching_result = {}
+
+    @wraps(func)
+    def wrapper(*args) -> dict:
```
Dimosphen1, line 9: You should return `Any` for `wrapper` because the result of calculations should not always be dict.

Dimosphen1 (changes requested):
A minor change was requested, try to re-run checks for this PR after the new commit.

PR #681, opened 2023-09-04
Solution

DanuloM:


Polyakiv-Andrey (changes requested):
flake8 not passed

DanuloM:
image

app/main.py
```diff
@@ -1,3 +1,14 @@
-def cache(func):
-    # Write your code here
-    pass
+from typing import Callable
+
+
+def cache(func: Callable) -> Callable:
+    list_of_args = {}
+
+    def inner(*args, **kwargs) -> Callable:
```
Dimosphen1, line 7: You do not return `Callable` in the `inner`, it could be `Any` because you cannot foresee the result of the function execution you return here.

app/main.py
```diff
@@ -1,3 +1,14 @@
-def cache(func):
-    # Write your code here
-    pass
+from typing import Callable
+
+
+def cache(func: Callable) -> Callable:
+    list_of_args = {}
```
Dimosphen1, line 5: Use another naming for the `list_of_args` variable, because you store a dict here, not list as naming indicates. You could use `inner_cache` for example.

app/main.py
Dimosphen1, line 7: You could use a decorator from `functools` for `inner` function to preserve the naming and docstring of the decorated function.

Dimosphen1 (changes requested):

PR #687, opened 2023-09-06
Solution

Julia15031988:


app/main.py
```diff
@@ -1,3 +1,18 @@
...
+    def wrapper(*args, **kwargs) -> dict:
+        key = (args, tuple(kwargs.items()),)
+        if key in results:
+            print("Getting from cache")
+            return results[key]
+        else:
+            result = func(*args, **kwargs)
+            results[key] = result
+            print("Calculating new result")
+            return result
```
Grayder0152, line 16: DRY, use only one return

Grayder0152 (changes requested):

app/main.py
```diff
@@ -1,3 +1,18 @@
...
+    def wrapper(*args, **kwargs) -> dict:
+        key = (args, tuple(kwargs.items()),)
+        if key in results:
+            print("Getting from cache")
+            result = results[key]
+        else:
+            result = func(*args, **kwargs)
+            results[key] = result
+            print("Calculating new result")
+        return result
```
```suggestion
    def wrapper(*args, **kwargs) -> dict:
        key = (args, tuple(kwargs.items()),)
        if key in results:
            print("Getting from cache")
        else:
            results[key] = func(*args, **kwargs)
            print("Calculating new result")
        return results[key]
```

app/main.py
```diff
@@ -1,3 +1,18 @@
-def cache(func):
-    # Write your code here
-    pass
+from typing import Callable
+
+
```
vsmutok, line 3: use functools.wraps

app/main.py
```diff
@@ -1,3 +1,18 @@
...
+def cache(func: Callable) -> Callable:
+    results = {}
```
vsmutok, line 5:
```suggestion
    cache_results = {}
```

vsmutok (changes requested):

ArsenPidhoretskyi (approved):
Well done!

PR #707, opened 2023-09-19
Solution

kravetb:


YukihiroSM (changes requested):
Fix flake, please.

app/main.py
```diff
@@ -1,3 +1,16 @@
...
-    pass
+from typing import Callable
+from functools import wraps
+
+
+def cache(func: Callable) -> Callable:
+    result: dict = {}
+
+    @wraps(func)
+    def inner(*args) -> int:
```
YukihiroSM, line 9: add kwargs as well, please.

app/main.py
```diff
@@ -1,3 +1,16 @@
...
+        if args not in result.keys():
+            result[args] = func(*args)
```
YukihiroSM, line 11: add kwargs as well, please.

YukihiroSM (changes requested):

app/main.py
kravetb: I add kwargs, but I dont't know is it right

app/main.py
```diff
@@ -1,3 +1,16 @@
...
+    result: dict = {}
+
+    @wraps(func)
+    def inner(*args, **kwargs) -> int:
+        if args not in result.keys():
+            result[args] = func(*args, **kwargs)
+            print("Calculating new result")
+        else:
+            print("Getting from cache")
+        return result[args]
```
YukihiroSM: as a key in result you should be using not only args, because in this case - we are ignoring kwargs, so your cache will not work properly. Please, fix it. Try to use args and kwargs as key. Consider, that kwargs is a dict, so it can't be directly used as key.

YukihiroSM (changes requested):

app/main.py
kravetb: Can you check my changes?

Bohdan-Salamakha (approved):

PR #712, opened 2023-09-21
Solution

Damodar-kumar:


app/main.py
```diff
@@ -1,6 +1,23 @@
 from typing import Callable
 
+results_list = []
```
ArsenPidhoretskyi: It must be created inside cache function
After fix function

ArsenPidhoretskyi: YOu must create dict into function and that store all arguments into it as key, and value - result of function

ArsenPidhoretskyi (changes requested):

Yurnerroo (approved):

PR #719, opened 2023-09-25
decorator of completed runs

Viesich:


app/main.py
```diff
@@ -2,5 +2,16 @@
...
-    # Write your code here
-    pass
+    cache_dict = {}
+
+    def inner(*args, **kwargs) -> dict:
+        key = (args, tuple(kwargs.items()))
+        if key in cache_dict:
+            print("Getting from cache")
+            return cache_dict[key]
+        else:
```

pavlejviki (changes requested):

app/main.py
Viesich: Sure! Thanks)

Polyakiv-Andrey (approved):
Good job!

PR #720, opened 2023-09-25
'Solution'

yul1valko:


app/main.py
```diff
@@ -1,6 +1,15 @@
-from typing import Callable
+from typing import Callable, Any
 
 
 def cache(func: Callable) -> Callable:
-    # Write your code here
-    pass
+    archive_functions = {}
```
Dimosphen1: You could store only args as a key for the inner cache. Rename your variable name and reformat the code to encounter that.

app/main.py
```diff
@@ -1,6 +1,15 @@
...
+
+    def wrapper(*args, **kwargs) -> Any:
```
Dimosphen1, line 9: You could use a decorator from `functools` for the `wrapper` to preserve the decorated function name and docstring.

Dimosphen1 (changes requested):

yul1valko:
Have implemented required changes

pavlejviki (approved):
Well done!

PR #723, opened 2023-09-25
Solution

a-kulaev:


app/main.py
```diff
@@ -2,5 +2,20 @@
 
 
 def cache(func: Callable) -> Callable:
-    # Write your code here
-    pass
+    cache_dict = {}
+
+    def wrapper(*args, **kwargs) -> Callable:
```
Dimosphen1, line 7: You could use a decorator from `functools` for the `wrapper` to preserve the decorated function name and docstring

app/main.py
```diff
@@ -2,5 +2,20 @@
...
+            return cache_dict[key]
+
+        result = func(*args, **kwargs)
+        cache_dict[key] = result
+
+        print("Calculating new result")
+        return result
+
+    return wrapper
+
```
Dimosphen1, line 21: Resolve flake8 error.

app/main.py
Dimosphen1, line 7: The return type for wrapper should be `Any` because the function calculation result is not known in advance.

Dimosphen1 (changes requested):

PR #724, opened 2023-09-26
Solution

D1st3f:
Added solution!

### DIFF
--- a/app/main.py
+++ b/app/main.py
@@ -5,9 +5,8 @@ def cache(func: Callable) -> Callable:
     cached_info = {}
 
     def inner(*args, **kwargs) -> int:
-        # Щоб унеможливити зміну ключа кешу, довелося перетворити
-        # ключові аргументи у tuple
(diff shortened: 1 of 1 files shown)


Familenko (approved):

app/main.py
```diff
@@ -2,5 +2,17 @@
 
 
 def cache(func: Callable) -> Callable:
-    # Write your code here
-    pass
+    cached_info = {}
+
+    def inner(*args, **kwargs) -> int:
+        # Щоб унеможливити зміну ключа кешу, довелося перетворити
+        # ключові аргументи у tuple
```
LiudmylaKulinchenko: Could try this one! But you should remove your comments and just create additional variable for sorted kwargs tuple to make it more readable

LiudmylaKulinchenko (changes requested):

app/main.py
D1st3f: Done!

fsocie7y (approved):
well dove ^_^

PR #726, opened 2023-09-27
solution

Familenko:


vshvanska (approved):

app/main.py
```diff
@@ -1,6 +1,26 @@
...
 
 def cache(func: Callable) -> Callable:
-    # Write your code here
-    pass
+    cached = []
+    arg_cache = []
+
+    @wraps(func)
+    def wrapper(*args) -> Any:
+        nonlocal cached, arg_cache
```
fsocie7y: this statement is redundant, as dict is mutable type

app/main.py
```diff
@@ -1,6 +1,26 @@
...
+
+        if args not in arg_cache:
+            print("Calculating new result")
+            result = func(*args)
+            cached.append(result)
+            arg_cache.append(args)
+
+            return result
+
+        else:
```

app/main.py
```diff
@@ -1,6 +1,26 @@
-from typing import Callable
+from typing import Callable, Any
+from functools import wraps
 
 
 def cache(func: Callable) -> Callable:
-    # Write your code here
-    pass
+    cached = []
```
fsocie7y: your approach is incorrect: you should store arguments and the result of the function with these arguments and return this result without calling function when arguments are the same.

fsocie7y (changes requested):

app/main.py
Familenko: I want to use list becouse it is more comfortable to operate list as a row in DataFrame in case if it will be reciered to store as structured info. And both of the list have same input index for ich time use. I saw that a lot of people use dic in this task, but my way also give same output and additionaly more comfortable for Data Scierntist job (in case if it will be required to analize cash)

app/main.py
Familenko: I do not use dict

app/main.py
LiudmylaKulinchenko: You missed this one

app/main.py
LiudmylaKulinchenko: You should use a dict here. Under the hood dicts in python use hash tables. You will go through this in Dict Advanced topic but shortly: it works faster than list. You will retrieve values with much less efforts
Using a dictionary provides a more concise and clear representation of the data
It's sutable if you will have a dict instead of two lists for Data Analysis. Data scientists often work with dictionaries or other data types, different from the python lists (numpy arrays for example)

app/main.py
LiudmylaKulinchenko: Lists are also mutable, so mind this comment

LiudmylaKulinchenko (changes requested):

app/main.py
Familenko: I used this decorator when train a machine learning model. During the experiment I needed to go through different variants of hyperparameters and correlate these values with the values already obtained. In case of failure to find the optimal parameters, I could use the saved cache values to return to the place where I had started on a non-straight path. In this case, the order of the experiments was also important to me (which could be taken into account in the cache only when using a list)

The task conditions do not say what exactly the caching will be used for, if for the log, then you are right and the speed of information retrieval matters and you should use a dictionary. But if the task is similar to the one I performed with this decorator, then the order of caching is more important than the fact that the result will be 0.0001 seconds slower.

So, since the priority of performance or usability was not explicitly stated, why is my option so unpleasant? And as you say "Data scientists often work with numpy arrays» wich is vary easy to get from Python list, but not dict

fsocie7y (approved):
well done 👍

PR #727, opened 2023-09-28
Solution

vshvanska:


DenPrislipskyi (approved):

app/main.py
```diff
@@ -1,6 +1,18 @@
...
-    # Write your code here
-    pass
+    caching_results = {}
+
+    def wrapper(*args) -> Any:
+        cach_key = args
+        if cach_key in caching_results.keys():
+            print("Getting from cache")
+            return caching_results[args]
+        else:
```
fsocie7y: better try without else
its best practice avoid else operator if it possible

app/main.py
```diff
@@ -1,6 +1,18 @@
...
 def cache(func: Callable) -> Callable:
-    # Write your code here
-    pass
+    caching_results = {}
+
+    def wrapper(*args) -> Any:
+        cach_key = args
+        if cach_key in caching_results.keys():
+            print("Getting from cache")
+            return caching_results[args]
```
fsocie7y, line 10: i think its better use cach_key as key for dic
in similar case too

fsocie7y (changes requested):

app/main.py
```diff
@@ -5,14 +5,12 @@ def cache(func: Callable) -> Callable:
     caching_results = {}
 
     def wrapper(*args) -> Any:
-        cach_key = args
-        if cach_key in caching_results.keys():
+        if args in caching_results.keys():
```

fsocie7y (changes requested):

fsocie7y (approved):
horay, well done ❤️

PR #728, opened 2023-09-28
Solution decorator

DenPrislipskyi:


MiskoRuslan (approved):

MiskoRuslan:
gj man ;)

mykolamateichuk (approved):

fedorov-s-od (approved):

fedorov-s-od:
I'm not sure, that *args is a Callable object (line 7). But overall, this is a nice and concise solution, excellent job! Thanks to your solution, I've found some issues with my own one)

app/main.py
```diff
@@ -2,5 +2,14 @@
 
 
 def cache(func: Callable) -> Callable:
-    # Write your code here
-    pass
+    cache_dict = {}
+
+    def wrapper(*args: Callable) -> Callable:
```
fsocie7y: in this case better to use other type annotation to function return
u return not callable, u return result from func()

app/main.py
```diff
@@ -2,5 +2,14 @@
...
+        if args not in cache_dict:
+            cache_dict[args] = func(*args)
+            print("Calculating new result")
+        else:
```
fsocie7y, line 11: u can do it without else

fsocie7y (changes requested):

app/main.py
```diff
@@ -1,6 +1,15 @@
...
-    pass
+    cache_dict = {}
+
+    def wrapper(*args) -> Any:
+        if args not in cache_dict:
+            cache_dict[args] = func(*args)
+            print("Calculating new result")
+            return cache_dict[args]
+        print("Getting from cache")
+        return cache_dict[args]
```
LiudmylaKulinchenko, line 13: try to rewrite your decorator not to have `return` duplicated

LiudmylaKulinchenko (changes requested):

LiudmylaKulinchenko (commented):
Don't mark comments as resolved so we could easily see them while reviewing a work the next time

LiudmylaKulinchenko (approved):
Well done! 🌼🌺

PR #730, opened 2023-09-28
Cache decorator solution

Lyutillis:


app/main.py
```diff
@@ -2,5 +2,13 @@
 
 
 def cache(func: Callable) -> Callable:
-    # Write your code here
-    pass
+    stored_runs = {}
+
+    def wrapper(*args) -> int:
```
mykolamateichuk: You can't always guarantee that the function you run will return `int` type.

mykolamateichuk (changes requested):

Shatrovskyi:
In my opinion, for more readable code it's nice to use one return with the ternary operator in wrapper().

Shatrovskyi (approved):

app/main.py
```diff
@@ -1,6 +1,14 @@
...
-    # Write your code here
-    pass
+    stored_runs = {}
+
+    def wrapper(*args) -> Any:
+        if args in stored_runs:
+            print("Getting from cache")
+            return stored_runs[args]
+        print("Calculating new result")
+        stored_runs.update({args: func(*args)})
```
LiudmylaKulinchenko: Use `stored_runs[args] = func(*args)`, your approach introduces unnecessary complexity without any significant advantage. Also it will be more pythonic

app/main.py
```diff
@@ -1,6 +1,14 @@
...
 
 def cache(func: Callable) -> Callable:
-    # Write your code here
-    pass
+    stored_runs = {}
+
+    def wrapper(*args) -> Any:
+        if args in stored_runs:
+            print("Getting from cache")
+            return stored_runs[args]
```
LiudmylaKulinchenko: Rewrite without having return duplicated

LiudmylaKulinchenko (changes requested):

LiudmylaKulinchenko (approved):
Great! ☀

PR #731, opened 2023-09-28
Solution

fedorov-s-od:
IMHO, Decorators is a very hard topic to comprehend... ) Solving this task, I thought It was over my head

### DIFF
--- a/app/main.py
+++ b/app/main.py
@@ -2,5 +2,19 @@ from typing import Callable
 
 
 def cache(func: Callable) -> Callable:
-    # Write your code here
-    pass
+    cache_dict = {}
+
+    def inner(*args, **kwargs) -> None:
+        key = args
+
+        if key in cache_dict:
+            print("Getting from cache")
+            return cache_dict[key]
+        else:
+            print("Calculating new result")
+            result = func(*args, **kwargs)
+            cache_dict[key] = result
+
+            return result
+
+    return inner
--- a/app/main.py
+++ b/app/main.py
@@ -4,17 +4,15 @@ from typing import Callable
 def cache(func: Callable) -> Callable:
     cache_dict = {}
 
-    def inner(*args, **kwargs) -> None:
-        key = args
+    def inner(*args, **kwargs) -> [list, int]:
 
-        if key in cache_dict:
+        if args in cache_dict:
             print("Getting from cache")
-            return cache_dict[key]
-        else:
-            print("Calculating new result")
-            result = func(*args, **kwargs)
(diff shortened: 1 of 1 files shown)


Naz-iv (approved):
Looking good, well done. Maybe no need to create extra variable in line 8 since you can just write <if args in cache_dict>, everything else looks good

fedorov-s-od:
> Looking good, well done. Maybe no need to create extra variable in line 8 since you can just write , everything else looks good

Thank you! Yeah, when I realized that creating a variable in line 8 was redundant it was already too late ))) The pull request had been done )

app/main.py
```diff
@@ -2,5 +2,19 @@
 
 
 def cache(func: Callable) -> Callable:
-    # Write your code here
-    pass
+    cache_dict = {}
+
+    def inner(*args, **kwargs) -> None:
+        key = args
```
MiskoRuslan: indeed, on line 8 you can do without "key"

app/main.py
```diff
@@ -2,5 +2,19 @@
...
+
+        if key in cache_dict:
+            print("Getting from cache")
+            return cache_dict[key]
+        else:
```
MiskoRuslan, line 11: you can try to do without "else"

MiskoRuslan (changes requested):
Everything seems great, there are a few edits left

app/main.py
fedorov-s-od: Yeah, it was a silly thing ))) Goin' to correct ;)

app/main.py
```diff
@@ -2,5 +2,17 @@
...
+    def inner(*args, **kwargs) -> [list, int]:
+
+        if args in cache_dict:
+            print("Getting from cache")
+            return cache_dict[args]
+
+        print("Calculating new result")
+        cache_dict[args] = func(*args, **kwargs)
+
+        return cache_dict[args]
```
LiudmylaKulinchenko, line 15: try to rewrite your decorator not to have `return` duplicated

app/main.py
```diff
@@ -2,5 +2,17 @@
 
 
 def cache(func: Callable) -> Callable:
-    # Write your code here
-    pass
+    cache_dict = {}
+
+    def inner(*args, **kwargs) -> [list, int]:
```
LiudmylaKulinchenko: inner can return not only that types, it's better to use Any

LiudmylaKulinchenko (changes requested):

fsocie7y (approved):
well done ^_^

PR #734, opened 2023-09-29
SOL

neagarr:


app/main.py
```diff
@@ -1,6 +1,17 @@
...
+
+    @wraps(func)
+    def wrapper(*args, **kwargs) -> Any:
+        if f"{args}, {kwargs}" in database.keys():
+            print("Getting from cache")
+            return database[f"{args}, {kwargs}"]
+        else:
+            print("Calculating new result")
+            database[f"{args}, {kwargs}"] = func(*args, **kwargs)
+            return database[f"{args}, {kwargs}"]
```
ArsenPidhoretskyi: return the same in both cases
Move it after if-else statement

ArsenPidhoretskyi (changes requested):

app/main.py
neagarr: ok

ArsenPidhoretskyi (approved):
Well done!

PR #737, opened 2023-09-29
Solution

Naz-iv:


app/main.py
```diff
@@ -2,5 +2,16 @@
...
-    # Write your code here
-    pass
+
+    cached_results_storage = {}
+
+    def inner(*args, **kvargs) -> int:
+        if args in cached_results_storage:
+            print("Getting from cache")
+            return cached_results_storage.get(args)
+        else:
```
sakhaline: This "else" seems redundant. Try to rewrite without it.

Naz-iv: Why exactly is it redundant? Since I want two different codes to execute depending on the result of a condition.

sakhaline: The code that is written in the block "else" can be taken out of the conditional block and simply written below at the same level as the "if" statement. This will not affect the result but will make your code better!

Naz-iv: I agree that it's good code style to not use else, if not needed. but in my case it's needed. Because I don't want my program to print "Calculating new result" when 0 positional arguments are passes to function.

sakhaline: If you take the code out of "else" block, it will still be executed only when the block of code in "if" has not been executed. The function will complete execution if it enters "if" due to "return".

roffi37 (commented):
I agree with @sakhaline , else is not needed, just write the code from the else block on the same level as if.

Wonsky1 (changes requested):
Rewrite the code without "else"
Other changes seems good, GJ

Naz-iv:
I'll try to explain one more time.... Else block is need. If my function didn't get any positional argument block if-else won't be executed and function won't give error for test with 0 arguments. If I take away block else, function will always try to execute second part of the code, this will cause Error. So else is there because I don't want this block of code to be executed every time... I only won't it executed when function received positional argument. You can fetch my code and try it yourself.

sakhaline:
I`ve tried to run your code without else, it passed all the tests!

app/main.py
```diff
@@ -2,5 +2,16 @@
...
+            print("Calculating new result")
+            result = func(*args, **kvargs)
+            cached_results_storage[args] = result
+            return result
```
LiudmylaKulinchenko: avoid using additional variable here

app/main.py
```diff
@@ -2,5 +2,16 @@
...
 def cache(func: Callable) -> Callable:
-    # Write your code here
-    pass
+
+    cached_results_storage = {}
+
+    def inner(*args, **kvargs) -> int:
+        if args in cached_results_storage:
+            print("Getting from cache")
+            return cached_results_storage.get(args)
```
LiudmylaKulinchenko: U don't need to use `get` here. You can be sure that there will be such key, you did this check in if statement

app/main.py
```diff
@@ -2,5 +2,16 @@
 
 
 def cache(func: Callable) -> Callable:
-    # Write your code here
-    pass
+
+    cached_results_storage = {}
+
+    def inner(*args, **kvargs) -> int:
```
LiudmylaKulinchenko: inner can return not only integers, it's better to use Any

LiudmylaKulinchenko: `kwargs`*

app/main.py

LiudmylaKulinchenko (changes requested):

sakhaline (approved):

app/main.py
```diff
@@ -1,6 +1,23 @@
...
+    @functools.wraps(func)
+    def inner(*args, **kwargs) -> any:
+        cache_key = func.__name__, args, tuple(kwargs.items())
+
+        if cache_key not in cached_results_storage:
+            func_res = func(*args, **kwargs)
+            cached_results_storage[cache_key] = func_res
+
+            print("Calculating new result")
+            return func_res
```
LiudmylaKulinchenko: avoid using additional variable here, you can skip assigning `func_res` here

LiudmylaKulinchenko (changes requested):

app/main.py
Naz-iv: Not sure how since I need to write this calculation to cached_results_storage. If I calculate it directly in return I won't be able to extract it

Tituha (approved):

app/main.py
LiudmylaKulinchenko: You have `return cached_results_storage[cache_key]` in 21 line. It won't be a problem to do the same here

app/main.py
```diff
@@ -1,6 +1,23 @@
 from typing import Callable
+import functools
 
 
 def cache(func: Callable) -> Callable:
-    # Write your code here
-    pass
+
+    cached_results_storage = {}
```
LiudmylaKulinchenko: make a shorter variable name. Remove `storage` from the variable name won't make it less descriptive

LiudmylaKulinchenko (changes requested):

app/main.py
Naz-iv: Thanks for advice with return statement.... somehow I didn't realize I can do it this way. Fixed now

LiudmylaKulinchenko (approved):
Good job! 🌼

PR #738, opened 2023-09-29
Implement py-cache-decorator

SevKrok:


Sel-Fisher (approved):
i like it👍

app/main.py
```diff
@@ -2,5 +2,14 @@
...
 
 def cache(func: Callable) -> Callable:
-    # Write your code here
-    pass
+    cached_results = {}
+
+    def wrapper(*args) -> int:
+        if args in cached_results:
+            print("Getting from cache")
+        else:
```
fsocie7y: better try without else, use return in if instead

app/main.py
```diff
@@ -2,5 +2,14 @@
...
+            cached_results[args] = func(*args)
+            print("Calculating new result")
+        return cached_results[args]
```
fsocie7y, line 13: then here do caching and return func result

fsocie7y (changes requested):

skoval7 (approved):

skoval7:
everything ok

LiudmylaKulinchenko (approved):
Well done 😋

PR #739, opened 2023-09-29
Solution

Shatrovskyi:


kostomeister (approved):
Nice one!

app/main.py
```diff
@@ -2,5 +2,16 @@
 
 
 def cache(func: Callable) -> Callable:
-    # Write your code here
-    pass
+    cached = {}
+
+    def inner(*args, **kwargs) -> None:
```
fsocie7y: it's better to use Any for return type annotation in this case

app/main.py
```diff
@@ -2,5 +2,16 @@
...
+        if args in cached.keys():
```
in similar case too

fsocie7y (changes requested):

app/main.py
Shatrovskyi: Got it.

app/main.py
Shatrovskyi: Got it! Thanks.

app/main.py
```diff
@@ -1,6 +1,17 @@
...
+
+    def inner(*args, **kwargs) -> Any:
+        if args in cached:
+            print("Getting from cache")
+        else:
+            new_cache = func(*args)
+            cached[args] = new_cache
+            print("Calculating new result")
+
+        return func(*args) if args not in cached.keys() else cached[args]
```
fsocie7y: .keys() again)))

fsocie7y: and may be better use just one if else statment

fsocie7y (changes requested):

app/main.py
Shatrovskyi: If  I replace one if statement, inner() is going to have two returns, isn't it? It's not good as well.

Shatrovskyi: I've found out how to deal with that. Now It looks better.

LiudmylaKulinchenko (approved):
Well done! 🐇

PR #740, opened 2023-09-29
Solution

sakhaline:


YuliiaSaliuta (approved):

RabenkoYevhenii (approved):

app/main.py
```diff
@@ -1,6 +1,22 @@
 from typing import Callable
+import functools
 
 
 def cache(func: Callable) -> Callable:
-    # Write your code here
-    pass
+    @functools.wraps(func)
+    def wraper(*args, **kwargs) -> int:
```
LiudmylaKulinchenko: inner can return not only integer, it's better to use Any

LiudmylaKulinchenko: `wrapper`*

app/main.py
```diff
@@ -1,6 +1,22 @@
...
+            func_res = func(*args, **kwargs)
+            wraper.execution_cache[cache_key] = func_res
+
+            print("Calculating new result")
+            return func_res
+
+        print("Getting from cache")
+        return wraper.execution_cache[cache_key]
+
+    wraper.execution_cache = {}
```
LiudmylaKulinchenko: no need to use it as a func attribute if u store function name in the key
It takes more memory

app/main.py
```diff
@@ -1,6 +1,22 @@
...
+    @functools.wraps(func)
+    def wraper(*args, **kwargs) -> int:
+        cache_key = func.__name__, args, tuple(kwargs.items())
+
+        if cache_key not in wraper.execution_cache:
+            func_res = func(*args, **kwargs)
+            wraper.execution_cache[cache_key] = func_res
+
+            print("Calculating new result")
+            return func_res
```
LiudmylaKulinchenko: try to avoid using additional variable here

LiudmylaKulinchenko (changes requested):

RVChornyy (approved):

fsocie7y (approved):
well done ^_^

PR #741, opened 2023-09-29
Solution

Sel-Fisher:


kostomeister (approved):
Wow such an amazing code, good jobs

app/main.py
```diff
@@ -2,5 +2,13 @@
 
 
 def cache(func: Callable) -> Callable:
-    # Write your code here
-    pass
+    data = {}
+
+    def inner(*args) -> int:
```
LiudmylaKulinchenko: inner can return not only int, it's better to use Any

app/main.py
```diff
@@ -2,5 +2,13 @@
 
 
 def cache(func: Callable) -> Callable:
-    # Write your code here
-    pass
+    data = {}
```
LiudmylaKulinchenko: use descriptive names

LiudmylaKulinchenko (changes requested):

kostomeister (approved):
good jobs!

app/main.py
```diff
@@ -2,5 +2,13 @@
 
 
 def cache(func: Callable) -> Callable:
-    # Write your code here
-    pass
+    saved_cache = {}
+
+    def inner(*args) -> any:
```
wQuelS: `Any` from typing, not `any`

wQuelS (changes requested):

fsocie7y (approved):
well done 🚀

PR #742, opened 2023-09-29
Solution

kostomeister:


Sel-Fisher (approved):
code like a burger from blackstar👍

app/main.py
```diff
@@ -2,5 +2,16 @@
 
 
 def cache(func: Callable) -> Callable:
-    # Write your code here
-    pass
+    stored_values = {}
+
+    def wrapper(*args) -> int | list:
```
LiudmylaKulinchenko: inner can return not only tuple, it's better to use Any

LiudmylaKulinchenko (changes requested):

LiudmylaKulinchenko (approved):
Well done! 🎈

PR #743, opened 2023-09-29
Develop

MykytaPD09:


app/main.py
```diff
@@ -2,5 +2,19 @@
...
+    from functools import wraps
+    cached_results = {}
+
+    @wraps(func)
+    def wrapper(*args, **kwargs) -> int:
+        key = (args, frozenset(kwargs.items()))
+        if key in cached_results:
+            print("Getting from cache")
+            return cached_results[key]
+        else:
```
viktoria-rybenchuk: remove redundant `else`

app/main.py
```diff
@@ -2,5 +2,19 @@
 
 
 def cache(func: Callable) -> Callable:
-    # Write your code here
-    pass
+    from functools import wraps
```
viktoria-rybenchuk: why do you make this import here?

viktoria-rybenchuk (changes requested):

app/main.py
Dimosphen1: The redundant `else` should be removed, code will not execute after the `return` statement anyway.

app/main.py
```diff
@@ -1,6 +1,20 @@
...
+from functools import wraps
 
 
 def cache(func: Callable) -> Callable:
-    # Write your code here
-    pass
+    cached_results = {}
+
+    @wraps(func)
+    def wrapper(*args, **kwargs) -> int:
```
Dimosphen1: The return type could be `Any` because the output of the wrapped function is not known in advance to the decorator.

app/main.py
```diff
@@ -1,6 +1,20 @@
 from typing import Callable
+from functools import wraps
```
Dimosphen1, line 1: Imports should be done in alphabetical order here.

Dimosphen1 (changes requested):

app/main.py
MykytaPD09: якщо я видаляю еlse то не проходять тести

pavlejviki (approved):
Well done!

PR #745, opened 2023-09-29
The decorator "cache" is created, tested and ready to work

Moonralex:
Decorator for caching results of functions with immutable arguments.

The `cache` function is designed to automatically cache the results of functions with immutable (unchangeable) arguments.
Each function decorated with this decorator gets its own cache to store results.
In this decorator, the cache is reset with each new test suite.

Args:
    func (Callable): The function to be decorated.

Returns:
    Callable: The modified function that caches results based on arguments.

Examples:
    @cache
    def long_time_func(a: int, b: int, c: int) -> int:
        return (a ** b ** c) % (a * c)
    result = long_time_func(2, 3, 4)  # Computation is performed and result is cached
    result = long_time_func(2, 3, 4)  # Result is retrieved from cache, computation is not repeated

Note:
    This decorator supports caching results for functions with arguments that are immutable types,
    such as int, float, str, tuple, bool, and NoneType. The results are stored in memory
    and used for subsequent calls to the function with the same arguments.

### DIFF
--- a/app/main.py
+++ b/app/main.py
@@ -2,38 +2,6 @@ from typing import Callable, Any
 
 
 def cache(func: Callable) -> Callable:
-    """
-    Decorator for caching results of functions with immutable arguments.
-
-    The `cache` function is designed to automatically cache results of
-    functions with immutable (unchangeable) arguments. Each function
-    decorated with this decorator gets its own cache to store results.
-    In this decorator, the cache is reset with each new test suite.
-
-    Args:
-        func (Callable): The function to be decorated.
-
-    Returns:
-        Callable: The modified function that caches results based on
-        arguments.
-
-    Examples:
-        @cache
-        def long_time_func(a: int, b: int, c: int) -> int:
-            return (a ** b ** c) % (a * c)
-        result = long_time_func(2, 3, 4)  # Computation is performed
-                                          # and result is cached
-        result = long_time_func(2, 3, 4)  # Result is retrieved from
-                                          # cache, computation is not
-                                          # repeated
-
-    Note:
-        This decorator supports caching results for functions with
-        arguments that are immutable types, such as int, float, str,
-        tuple, bool, and NoneType. The results are stored in memory
-        and used for subsequent calls to the function with the same
-        arguments.
-    """
     cached_data = {}
     immutable_type = (int, float, str, tuple, bool, type(None))
 
@@ -46,12 +14,10 @@ def cache(func: Callable) -> Callable:
             if func_key in cached_data[func.__name__]:
                 print("Getting from cache")
                 return cached_data[func.__name__][func_key]
-            else:
-                print("Calculating new result")
-                func_result = func(*args)
-                cached_data[func.__name__][func_key] = func_result
-                return func_result
-        else:
             print("Calculating new result")
-            return func(*args)
+            func_result = func(*args)
+            cached_data[func.__name__][func_key] = func_result
+            return func_result
+        print("Calculating new result")
+        return func(*args)
     return argument_based_cacher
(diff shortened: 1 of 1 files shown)


app/main.py
```diff
@@ -1,6 +1,57 @@
...
+    def argument_based_cacher(*args) -> Any:
+        nonlocal cached_data
+        if all([isinstance(variable, immutable_type) for variable in args]):
+            func_key = tuple(args)
+            if func.__name__ not in cached_data.keys():
+                cached_data[func.__name__] = {}
+            if func_key in cached_data[func.__name__]:
+                print("Getting from cache")
+                return cached_data[func.__name__][func_key]
+            else:
```
Bohdan-Salamakha: remove redundant 'else'

app/main.py
```diff
@@ -1,6 +1,57 @@
...
+                print("Calculating new result")
+                func_result = func(*args)
+                cached_data[func.__name__][func_key] = func_result
+                return func_result
+        else:
```
Bohdan-Salamakha: remove redundant 'else'

app/main.py
```diff
@@ -1,6 +1,57 @@
...
+                                          # cache, computation is not
+                                          # repeated
+
+    Note:
+        This decorator supports caching results for functions with
+        arguments that are immutable types, such as int, float, str,
+        tuple, bool, and NoneType. The results are stored in memory
+        and used for subsequent calls to the function with the same
+        arguments.
+    """
```
Bohdan-Salamakha: docstring is super redundant here
remove it

Bohdan-Salamakha (changes requested):

app/main.py
```diff
@@ -1,6 +1,23 @@
...
 
 
 def cache(func: Callable) -> Callable:
-    # Write your code here
-    pass
+    cached_data = {}
+    immutable_type = (int, float, str, tuple, bool, type(None))
+
+    def argument_based_cacher(*args) -> Any:
+        nonlocal cached_data
```
pavlejviki: you're modifying the dictionary that cached_data points to (by adding keys and values). Since you're not re-assigning the variable, you don't need nonlocal.

app/main.py
```diff
@@ -1,6 +1,23 @@
...
+        if all([isinstance(variable, immutable_type) for variable in args]):
+            func_key = tuple(args)
```
pavlejviki: args is already a tuple, and there's no need to convert it again using tuple(args).

app/main.py
```diff
@@ -1,6 +1,23 @@
...
+            if func.__name__ not in cached_data.keys():
+                cached_data[func.__name__] = {}
+            if func_key in cached_data[func.__name__]:
+                print("Getting from cache")
+                return cached_data[func.__name__][func_key]
+            print("Calculating new result")
+            func_result = func(*args)
+            cached_data[func.__name__][func_key] = func_result
+            return func_result
+        print("Calculating new result")
```
pavlejviki, line 13: Decorator cache is supposed to decorate only functions that take immutable arguments.

pavlejviki (changes requested):

app/main.py
```diff
@@ -1,6 +1,17 @@
-from typing import Callable
+from typing import Callable, Any
 
 
 def cache(func: Callable) -> Callable:
-    # Write your code here
-    pass
+    cached_data = {}
+
+    def argument_based_cacher(*args) -> Any:
```
Dimosphen1, line 7: You could use decorator from `functools` to preserve the wrapped function naming and docstring.

Dimosphen1 (approved):
LGTM, one comment is left for consideration.

PR #746, opened 2023-09-29
Solution

MiskoRuslan:
I feel a little problem on the seventh line :))
But let`s try

### DIFF
--- a/app/main.py
+++ b/app/main.py
@@ -9,7 +9,7 @@ def cache(func: Callable) -> Callable:
         if args not in cache_dict:
             cache_dict[args] = func(*args)
             print("Calculating new result")
-            return cache_dict[args]
-        print("Getting from cache")
+        else:
+            print("Getting from cache")
         return cache_dict[args]
     return inner
--- a/app/main.py
+++ b/app/main.py
@@ -1,10 +1,11 @@
 from typing import Callable
+from typing import Any
 
 
 def cache(func: Callable) -> Callable:
     cache_dict = {}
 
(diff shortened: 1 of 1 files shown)


app/main.py
```diff
@@ -2,5 +2,13 @@
 
 
 def cache(func: Callable) -> Callable:
-    # Write your code here
-    pass
+    cache_dict = {}
+
+    def inner(*args: tuple) -> [list, int]:
```
LiudmylaKulinchenko: Yes, inner can return not only these types, it's better to use Any

LiudmylaKulinchenko: No need to add type annotations to *args. You can use, for example, `*args: int` if you want to indicate that func expects integer arguments

LiudmylaKulinchenko (changes requested):

app/main.py
```diff
@@ -1,6 +1,15 @@
...
 def cache(func: Callable) -> Callable:
-    # Write your code here
-    pass
+    cache_dict = {}
+
+    def inner(*args: int) -> Any:
+        if args not in cache_dict:
+            cache_dict[args] = func(*args)
+            print("Calculating new result")
+            return cache_dict[args]
```

fsocie7y (changes requested):

DenPrislipskyi (approved):

DenPrislipskyi (approved):

Oxbay (approved):

LiudmylaKulinchenko (approved):
Nice work! 🎈

PR #748, opened 2023-09-30
Solution

roffi37:


RVChornyy (changes requested):
Seems like "inner" output description should be other than "Callable". Because "inner" returns the result of func output packed in dictionary.  So, in my oppinion, "Any" is more preferable for "inner".
Also from task requirement: "Decorator `cache` creating for 
decorating only functions that take **immutable** arguments.". So, **kwargs as "inner" arguments are not acceptable.

katyakoziar (approved):
Good job!

app/main.py
```diff
@@ -2,5 +2,13 @@
...
 
 def cache(func: Callable) -> Callable:
-    # Write your code here
-    pass
+    cached_result = {}
+
+    def inner(*args, **kwargs) -> Callable:
+        if args in cached_result:
+            print("Getting from cache")
+            return cached_result[args]
```

fsocie7y (changes requested):

app/main.py
```diff
@@ -7,8 +7,8 @@ def cache(func: Callable) -> Callable:
     def inner(*args, **kwargs) -> Callable:
```
fsocie7y: inner return not callable object, better use here Any

fsocie7y (changes requested):

OlehShumov (approved):

fsocie7y (approved):
well done! 🥇

PR #750, opened 2023-09-30
Solution: created decorator "Cache"

petrykivd:


app/main.py
```diff
@@ -2,5 +2,14 @@
 
 
 def cache(func: Callable) -> Callable:
-    # Write your code here
-    pass
+    cache_dict = {}
+
+    def wrapper(*args) -> int:
```
LiudmylaKulinchenko: inner can return not only integer, it's better to use Any

LiudmylaKulinchenko (changes requested):

app/main.py
```diff
@@ -1,6 +1,15 @@
...
-    # Write your code here
-    pass
+    cache_dict = {}
+
+    def wrapper(*args) -> Any:
+        if args in cache_dict:
+            print("Getting from cache")
+        else:
+            new_result = func(*args)
+            cache_dict[args] = new_result
```
LiudmylaKulinchenko: Avoid using additional variable, it is not needed here

LiudmylaKulinchenko (changes requested):

LiudmylaKulinchenko (approved):
Well done 🚀

PR #752, opened 2023-09-30
Solution

skoval7:


app/main.py
```diff
@@ -2,5 +2,27 @@
 
 
 def cache(func: Callable) -> Callable:
-    # Write your code here
-    pass
+
+    cache_dict = {}
+
+    def wrapper(*args, **kwargs) -> None:
```
LiudmylaKulinchenko: Does it return None?

app/main.py
```diff
@@ -2,5 +2,27 @@
...
+
+
+@cache
+def long_time_func(arg1: int, arg2: int, arg3: int) -> int:
+    return (arg1 ** arg2 ** arg3) % (arg1 * arg3)
+
+
+@cache
+def long_time_func_2(n_tuple: tuple, power: int) -> int:
+    return [number ** power for number in n_tuple]
```
LiudmylaKulinchenko: Remove test funcs

app/main.py
```diff
@@ -2,5 +2,27 @@
...
+    def wrapper(*args, **kwargs) -> None:
+        key = (args, frozenset(kwargs.items()))
+        if key in cache_dict:
+            print("Getting from cache")
+            return cache_dict[key]
+        else:
+            print("Calculating new result")
+            result = func(*args, **kwargs)
+            cache_dict[key] = result
+            return result
```
LiudmylaKulinchenko: Avoid using additional variable here

LiudmylaKulinchenko (changes requested):

app/main.py
```diff
@@ -1,6 +1,28 @@
...
+            print("Getting from cache")
+            return cache_dict[key]
+        else:
+            print("Calculating new result")
+            cache_dict[key] = func(*args, **kwargs)
+            return cache_dict[key]
+    return wrapper
+
+
+# I wrote a new function because I deleted the old one
```
LiudmylaKulinchenko: You don't need these functions here. Remove tests from the main file, you already have tests for the decorator, they are in the different module. Don't mix tests and main logic

app/main.py
```diff
@@ -1,6 +1,28 @@
...
+
+    def wrapper(*args, **kwargs) -> Any:
+        key = (args, frozenset(kwargs.items()))
+        if key in cache_dict:
+            print("Getting from cache")
+            return cache_dict[key]
+        else:
+            print("Calculating new result")
+            cache_dict[key] = func(*args, **kwargs)
+            return cache_dict[key]
```
LiudmylaKulinchenko: Move return out of condition so it won't be duplicated

LiudmylaKulinchenko (changes requested):

LiudmylaKulinchenko (approved):
Good job! 🌼

PR #754, opened 2023-10-01
Solution

katyakoziar:


app/main.py
```diff
@@ -1,6 +1,16 @@
-from typing import Callable
+from typing import Callable, Any
 
 
 def cache(func: Callable) -> Callable:
-    # Write your code here
-    pass
+    results_of_function_runs = {}
```
LiudmylaKulinchenko: use shorter name

app/main.py
```diff
@@ -1,6 +1,16 @@
...
+
+    def inner(*args) -> Any:
+        if args not in results_of_function_runs:
+            result = func(*args)
+            results_of_function_runs[args] = result
+            print("Calculating new result")
+            return result
```
LiudmylaKulinchenko: try to avoid using additional `result` variable here

app/main.py
```diff
@@ -1,6 +1,16 @@
...
+        else:
+            print("Getting from cache")
+            return results_of_function_runs[args]
```
LiudmylaKulinchenko: also try not to have `return` duplicated

LiudmylaKulinchenko (changes requested):

fsocie7y (approved):
GJ ❤️

PR #756, opened 2023-10-01
Solution

Wonsky1:
Implemented cache decorator

### DIFF
--- a/app/main.py
+++ b/app/main.py
@@ -2,5 +2,17 @@ from typing import Callable
 
 
 def cache(func: Callable) -> Callable:
-    # Write your code here
-    pass
+    cached_values = {}
+
+    def inner(*args) -> int:
+        cached_keys = args
+        if cached_keys not in cached_values:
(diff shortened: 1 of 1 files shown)


roffi37 (approved):

app/main.py
```diff
@@ -2,5 +2,16 @@
...
-    # Write your code here
-    pass
+    cached_results = {}
+
+    def inner(*args) -> int:
+        if args not in cached_results:
+            print("Calculating new result")
+            result = func(*args)
+            cached_results[args] = result
+            return result
```

fsocie7y (changes requested):

app/main.py
Wonsky1: @fsocie7y like the push i added last?

Wonsky1: Do i have to get rid of "else" block somehow, or it is fine?

fsocie7y (approved):
well done 🥇

PR #757, opened 2023-10-01
Solution

OlehShumov:


kostomeister (approved):
Superb

app/main.py
```diff
@@ -2,5 +2,17 @@
 
 
 def cache(func: Callable) -> Callable:
-    # Write your code here
-    pass
+    cache_results = {}
+
+    def wrapper(*args, **kwargs) -> None:
```
Andrii-Silenko:
```suggestion
    def wrapper(*args, **kwargs) -> Any:
```

Andrii-Silenko: You can import also Any from typing

Andrii-Silenko (approved):

app/main.py
```diff
@@ -2,5 +2,17 @@
...
+        arg_key = (args, frozenset(kwargs.items()))
+        if arg_key in cache_results:
+            print("Getting from cache")
+            return cache_results[arg_key]
```
try to rewrite your decorator not to have `return` duplicated

fsocie7y (changes requested):

app/main.py
fsocie7y: yep, better to use Any in this case

app/main.py
```diff
@@ -2,5 +2,17 @@
 
 
 def cache(func: Callable) -> Callable:
-    # Write your code here
-    pass
+    cache_results = {}
+
+    def wrapper(*args, **kwargs) -> None:
+        arg_key = (args, frozenset(kwargs.items()))
```
fsocie7y: i think its unnessesary, we can do just smthn like this 'if args in cache_dict:'
wbu?

fsocie7y (changes requested):

app/main.py
fsocie7y: sory, its my miss 
this action from u was ok

fsocie7y (approved):
gj 👍

PR #759, opened 2023-10-03
Solution

Andrii-Silenko:


app/main.py
```diff
@@ -1,6 +1,21 @@
...
-    # Write your code here
-    pass
+    previous_args = {}
+
+    def inner(*args) -> Any:
+        if all([
+            isinstance(arg, (tuple, str, int, float, bool))
+            for arg in args
+        ]):
+            if args in previous_args.keys():
```

app/main.py
```diff
@@ -1,6 +1,21 @@
...
+                print("Getting from cache")
+                return previous_args[args]
```

fsocie7y (changes requested):

fsocie7y (approved):
well done ^_^

PR #760, opened 2023-10-04
Solution

paashkovaaa:


app/main.py
```diff
@@ -1,6 +1,14 @@
...
-    # Write your code here
-    pass
+    cache_archive = {}
+
+    def inner(*args, **kwargs) -> Any:
+        if args in cache_archive:
+            print("Getting from cache")
+        else:
+            cache_archive[args] = func(*args, **kwargs)
+            print("Calculating new result")
```
YukihiroSM: add checking by kwargs as well or just remove kwargs, because in this case kwargs are ignored.

YukihiroSM (changes requested):

YukihiroSM (approved):

PR #761, opened 2023-10-04
'Solution'

Olala2104:


vasylhnatiuk (changes requested):
try to use only one `return`

Olala2104:
Any

app/main.py
```diff
@@ -1,6 +1,18 @@
...
+    def wrapper(*args: Any, **kwargs: Any) -> Any:
+        key = (args, frozenset(kwargs.items()))
+        if key in cached_results:
+            print("Getting from cache")
+            return cached_results[key]
+        else:
+            result = func(*args, **kwargs)
+            cached_results[key] = result
+            print("Calculating new result")
+            return result
```
```suggestion
    def wrapper(*args: Any, **kwargs: Any) -> Any:
        key = (args, frozenset(kwargs.items()))
        if key in cached_results:
            print("Getting from cache")
        else:
            cached_results[key] = func(*args, **kwargs)
            print("Calculating new result")
        return cached_results[key]
```

vsmutok (changes requested):

PR #762, opened 2023-10-04

Aeros72:


app/main.py
```diff
@@ -1,6 +1,14 @@
-from typing import Callable
+from typing import Callable, Any
 
 
 def cache(func: Callable) -> Callable:
-    # Write your code here
-    pass
+    params = {}
+
```
Polyakiv-Andrey, line 7: use here `from functools import wraps`

Polyakiv-Andrey (changes requested):

app/main.py
Aeros72: fixed

Dimosphen1 (approved):
LGTM, well done!

PR #765, opened 2023-10-05
Add solution

IlyaGrynyshyn:


app/main.py
```diff
@@ -2,5 +2,14 @@
 
 
 def cache(func: Callable) -> Callable:
-    # Write your code here
-    pass
+    cache_memory = {}
+
+    def wrapper(*args) -> dict:
```
Dimosphen1: Change your return type annotation for `wrapper`, it could return `Any` because the function output is not known in advance.

Dimosphen1: Consider using a decorator for `wrapper` from `functools` to preserve the wrapped function naming and docstring.

Dimosphen1 (changes requested):

app/main.py
```diff
@@ -1,6 +1,17 @@
-from typing import Callable
+from typing import Callable, Any
+import functools
```
Dimosphen1: `import` should precede `from ... import`, place your import in the right order.

app/main.py
```diff
@@ -1,6 +1,17 @@
...
 
 
 def cache(func: Callable) -> Callable:
-    # Write your code here
-    pass
+    functools.wraps(func)
```
Dimosphen1: Place a `wraps` decorator at the right place as [provided in the documentation](https://docs.python.org/uk/3/library/functools.html#functools.wraps).

Dimosphen1 (changes requested):

app/main.py
Dimosphen1: Place your imports in the right order according to the previous comment.

Dimosphen1 (changes requested):

viktoria-rybenchuk (approved):
Good job!

PR #767, opened 2023-10-07
Solution

ShcherbynaSv:


app/main.py
```diff
@@ -1,6 +1,16 @@
...
 def cache(func: Callable) -> Callable:
-    # Write your code here
-    pass
+    args_dict = {}
+
+    def saver(*args) -> Any:
+        if args in args_dict:
+            print("Getting from cache")
+            return args_dict[args]
+        else:
```

pavlejviki (changes requested):

app/main.py
```diff
@@ -1,6 +1,17 @@
 from typing import Callable
+from typing import Any
```
Dimosphen1: There is no need to duplicate `from typing import`, import `Any` together with `Callable` on the first line.

app/main.py
```diff
@@ -1,6 +1,17 @@
...
 
 
 def cache(func: Callable) -> Callable:
-    # Write your code here
-    pass
+    args_dict = {}
+
+    def saver(*args) -> Any:
```
Dimosphen1, line 9: You could use decorator from `functools` for `saver` to preserve the wrapped function name and docstring.

Dimosphen1 (changes requested):

Dimosphen1 (approved):
LGTM, well done!

PR #769, opened 2023-10-08
Task done

aleksandr-paslawski:


app/main.py
```diff
@@ -1,6 +1,19 @@
...
 
 
+cache_history = dict()
+
+
 def cache(func: Callable) -> Callable:
-    # Write your code here
-    pass
+    def wrapper(*args) -> str:
+        func_name = f"{func}({args})"
```
ArsenPidhoretskyi: This is redundant

app/main.py
```diff
@@ -1,6 +1,19 @@
 from typing import Callable
 
 
+cache_history = dict()
```
ArsenPidhoretskyi: This variable must be inside decorator, to create separate state for each function

ArsenPidhoretskyi (changes requested):

app/main.py
```diff
@@ -2,5 +2,17 @@
...
 
 def cache(func: Callable) -> Callable:
-    # Write your code here
-    pass
+    cache_history = dict()
+
+    def wrapper(*args) -> str:
+        # func_name = f"{func}({args})"
+
+        if f"{func}({args})" in cache_history:
```
```suggestion
        if args in cache_history:
```
Why do you need func?
Remove it from all places

ArsenPidhoretskyi (changes requested):

app/main.py
aleksandr-paslawski: I added the function name to the cache because I moved the cache to outside the decorator. In this case arguments could be repeated for different functions.

I corrected it as you recommended.
What else can I add to make the code better?

app/main.py
```diff
@@ -2,5 +2,17 @@
 
 
 def cache(func: Callable) -> Callable:
-    # Write your code here
-    pass
+    cache_history = dict()
+
+    def wrapper(*args) -> str:
+        # func_name = f"{func}({args})"
```
Yurnerroo: remove comment

app/main.py
```diff
@@ -2,5 +2,17 @@
 
 
 def cache(func: Callable) -> Callable:
-    # Write your code here
-    pass
+    cache_history = dict()
+
+    def wrapper(*args) -> str:
```
Yurnerroo:
```suggestion
    def wrapper(*args) -> Any:
```

Yurnerroo (changes requested):

Yurnerroo (approved):

PR #771, opened 2023-10-09
Solution

CrazyHilly:


app/main.py
```diff
@@ -2,5 +2,16 @@
...
-    pass
+    results = {}
+
+    def inner(*args) -> None:
+        nonlocal results
+        if args in results:
+            print("Getting from cache")
+        else:
+            print("Calculating new result")
+            results.update({args: func(*args)})
```
```suggestion
            results.update[args] = func(*args)
```
No need to use .update if change only one key

ArsenPidhoretskyi (changes requested):

app/main.py
CrazyHilly: thanks!

ArsenPidhoretskyi (approved):
Well done!

PR #772, opened 2023-10-09
Solution

yanamushka:


app/main.py
```diff
@@ -1,3 +1,16 @@
...
+    def wrapper(*args) -> Callable:
+        cache_key = (args)
+        if cache_key in cache_result.keys():
+            print("Getting from cache")
+        else:
+            cache_result[cache_key] = func(*args)
+            print("Calculating new result")
+        return cache_result[cache_key]
+    return wrapper
     pass
```
ArsenPidhoretskyi: Pass in this case not needed

app/main.py
```diff
@@ -1,3 +1,16 @@
-def cache(func):
-    # Write your code here
+from typing import Callable
+
+
+def cache(func: Callable) -> Callable:
+    cache_result = {}
+
+    def wrapper(*args) -> Callable:
+        cache_key = (args)
```
ArsenPidhoretskyi: Why you need this line?

ArsenPidhoretskyi (changes requested):

app/main.py
yanamushka: done

app/main.py
yanamushka: you are right, I don't need this one. Fixed

vasylhnatiuk (approved):

PR #773, opened 2023-10-10
Solution

BirovShandor:


app/main.py
```diff
@@ -2,5 +2,30 @@
...
+    results = {}
+
+    def inner(*args, **kwargs) -> None:
+
+        key = (args, frozenset(kwargs.items()))
+
+        if key in results:
+            print("Getting from cache")
+            return results[key]
+        else:
```
EdAlekseiev: This `else`-statement  is redundant

app/main.py
```diff
@@ -2,5 +2,30 @@
...
+            print("Calculating new result")
+            result = func(*args, **kwargs)
+            results[key] = result
+            return result
```
```suggestion
        print("Calculating new result")
        result = func(*args, **kwargs)
        results[key] = result
        return result
```

EdAlekseiev (changes requested):

EdAlekseiev (approved):
good job

PR #775, opened 2023-10-13
'Solution'

Sparix:
Can i use cashed[args] = result = func(*args)  in 10-11 line 
instead result = func(*args)
             cached[args] = result ?

### DIFF
--- a/app/main.py
+++ b/app/main.py
@@ -1,6 +1,19 @@
-from typing import Callable
+from typing import Callable, Any
 
 
 def cache(func: Callable) -> Callable:
-    # Write your code here
-    pass
+    cached = {}
+
+    def inner(*args: Any) -> Any:
+        if args not in cached:
+            print("Calculating new result")
+            result = func(*args)
+            cached[args] = result
+            return result
+        else:
+            print("Getting from cache")
+            for key, value in cached.items():
+                if key == args:
+                    return value
+
+    return inner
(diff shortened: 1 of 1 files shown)


app/main.py
```diff
@@ -1,6 +1,19 @@
...
+
+    def inner(*args: Any) -> Any:
+        if args not in cached:
+            print("Calculating new result")
+            result = func(*args)
+            cached[args] = result
+            return result
+        else:
+            print("Getting from cache")
+            for key, value in cached.items():
```
Polyakiv-Andrey: you can do it easier just get args from cached

Polyakiv-Andrey (changes requested):

app/main.py
```diff
@@ -1,6 +1,17 @@
...
-    pass
+    cached = {}
+
+    def inner(*args: Any) -> int:
+        if args not in cached:
+            print("Calculating new result")
+            result = func(*args)
+            cached[args] = result
+            return result
+        else:
```
viktoria-rybenchuk: redundant else

app/main.py
```diff
@@ -1,6 +1,17 @@
-from typing import Callable
+from typing import Callable, Any
 
 
 def cache(func: Callable) -> Callable:
-    # Write your code here
-    pass
+    cached = {}
+
+    def inner(*args: Any) -> int:
```
viktoria-rybenchuk: does this function return int?

viktoria-rybenchuk (changes requested):

viktoria-rybenchuk (approved):
Good job!

PR #778, opened 2023-10-14
Solution

IvanMozhar:


app/main.py
```diff
@@ -2,5 +2,16 @@
 
 
 def cache(func: Callable) -> Callable:
-    # Write your code here
-    pass
+    storage_dictionary = {}
+
+    def inner_function(*args, **kwargs) -> Callable:
```
vsmutok: don't forget about `functools.wraps`

app/main.py
```diff
@@ -2,5 +2,16 @@
...
+        key = (args, tuple(kwargs.items()))
+        if key in storage_dictionary:
+            print("Getting from cache")
+            return storage_dictionary[key]
+        else:
+            result = func(*args, **kwargs)
+            storage_dictionary[key] = result
+            print("Calculating new result")
+            return result
```
```suggestion
        if key in storage_dictionary:
            print("Getting from cache")
        else:
            storage_dictionary[key] = func(*args, **kwargs)
            print("Calculating new result")
        return storage_dictionary[key]
```

app/main.py
vsmutok: `Callable` wrong annotation here

vsmutok (changes requested):

vsmutok (approved):
Well Done 👍

PR #779, opened 2023-10-14
solution

boris-devs:


app/main.py
```diff
@@ -1,6 +1,32 @@
...
+    results_cache = {}
+
+    def wrapper(*args, **kwargs) -> dict:
+        args_tuple = args + tuple(kwargs.items())
+        if args_tuple in results_cache:
+            print("Getting from cache")
+            return results_cache[args_tuple]
+        else:
+            result = func(*args, **kwargs)
+            results_cache[args_tuple] = result
```

ArsenPidhoretskyi: And then move return after if-else statement

ArsenPidhoretskyi (changes requested):

ArsenPidhoretskyi (approved):
Well done!

PR #781, opened 2023-10-18
solution

Andrey1104:


app/main.py
```diff
@@ -2,5 +2,18 @@
...
-    pass
+    cache_dict = {}
+
+    def wrapper(*args, **kwargs) -> Callable:
+
+        cache_key = args + tuple(sorted(kwargs.items()))
+        if cache_key in cache_dict:
+            print("Getting from cache")
+            return cache_dict[cache_key]
+        else:
```
Dimosphen1: There is no need in the `else`, statement, code will not execute after `return` anyway. Remove `else`.

app/main.py
```diff
@@ -2,5 +2,18 @@
 
 
 def cache(func: Callable) -> Callable:
-    # Write your code here
-    pass
+    cache_dict = {}
+
+    def wrapper(*args, **kwargs) -> Callable:
```
Dimosphen1: The outcome of the wrapper is not always `Callable`, it could be `Any` and depends on the wrapped function logic. Change your return type annotation.

Dimosphen1 (changes requested):

Polyakiv-Andrey (approved):
Good job!

PR #791, opened 2023-10-23
'Solution'

Anton-Konyk:


app/main.py
```diff
@@ -1,6 +1,18 @@
...
 def cache(func: Callable) -> Callable:
-    # Write your code here
-    pass
+    result_dict = {}
+
+    def inner(*args) -> Any:
+        for key, value in result_dict.items():
+            if key == args:
+                print("Getting from cache")
+                return value
```
EdAlekseiev: I'm not sure, that is the best solution. I think, that you must check that key exists in dict and return value

```suggestion
        if args in result_dict:
            print("Getting from cache")
            return result_dict[args]
```

EdAlekseiev (changes requested):

Anton-Konyk:
You are right, thanks. It will be faster.

Anton-Konyk:
Please see the last version.

Anton-Konyk:
New version 30102023

EdAlekseiev (approved):
Good job

PR #793, opened 2023-10-24
solution_1

OleksandrHubar:


app/main.py
```diff
@@ -2,5 +2,16 @@
...
-    # Write your code here
-    pass
+    collected_dictionary = {}
+
+    def wrapper(*args, **kwargs) -> dict:
+        dict_key = (func, args, frozenset(kwargs.items()))
+        if dict_key in collected_dictionary:
+            print("Getting from cache")
+            return collected_dictionary[dict_key]
+        else:
```

app/main.py
```diff
@@ -2,5 +2,16 @@
 
 
 def cache(func: Callable) -> Callable:
-    # Write your code here
-    pass
+    collected_dictionary = {}
```
pavlejviki: variable name is confusing. maybe use cached_results or cache

pavlejviki (changes requested):

app/main.py
OleksandrHubar: @pavlejviki could you please check my update "changed_solution_1"

app/main.py
```diff
@@ -2,5 +2,17 @@
 
 
 def cache(func: Callable) -> Callable:
-    # Write your code here
-    pass
+    cached_results = {}
+
+    def wrapper(*args, **kwargs) -> dict:
```
viktoria-rybenchuk: does this function return `dict`?

viktoria-rybenchuk (changes requested):

app/main.py
OleksandrHubar: seems doesn't
I checked "Readme" and it looks like "int"

app/main.py
```diff
@@ -2,5 +2,17 @@
 
 
 def cache(func: Callable) -> Callable:
-    # Write your code here
-    pass
+    cached_results = {}
+
+    def wrapper(*args, **kwargs) -> int:
```
Dimosphen1: You should indicate that `wrapper` returns `Any` because we do not know in advance which function the cache decorator will be applied to.

Dimosphen1: Add a decorator from `functools` for `wrapper` to preserve the wrapped function name and docstring.

Dimosphen1 (changes requested):

pavlejviki (approved):
Good job!

PR #795, opened 2023-10-25
Solution

Marinel444:


app/main.py
```diff
@@ -2,5 +2,20 @@
...
 
 def cache(func: Callable) -> Callable:
-    # Write your code here
-    pass
+    result_dict = {}
+
+    def inner(*args) -> int:
+        func_name = func.__name__
+        args_str = str(args)
+        if func_name not in result_dict:
```
Polyakiv-Andrey: to check if a function is in the cache, just check if args is in the cache

Polyakiv-Andrey (changes requested):

Polyakiv-Andrey (approved):
Good job!

PR #799, opened 2023-10-30
Solution to the Cache decorator problem

Ananiev-Vitalii:


app/main.py
```diff
@@ -1,6 +1,21 @@
...
+    @wraps(func)
+    def wrapper(*args, **kwargs):
+        key = (args, tuple(kwargs.items()))
+
+        if key in cache_store:
+            print("Getting from cache")
+            return cache_store[key]
+
+        result = func(*args, **kwargs)
+        cache_store[key] = result
```

app/main.py
```diff
@@ -1,6 +1,21 @@
...
+from functools import wraps
 
 
 def cache(func: Callable) -> Callable:
-    # Write your code here
-    pass
+    cache_store = {}
+
+    @wraps(func)
+    def wrapper(*args, **kwargs):
```
ArsenPidhoretskyi: What about return type annotations?

ArsenPidhoretskyi (changes requested):

PR #803, opened 2023-10-31
Solution

Maxboond:


app/main.py
```diff
@@ -1,6 +1,18 @@
-from typing import Callable
+from typing import Callable, Any
 
 
-def cache(func: Callable) -> Callable:
-    # Write your code here
-    pass
+def cache(func: Callable) -> Any:
```
Dimosphen1: The `cache` function will return the `wrapper` which is callable. Change your type annotation here.

app/main.py
```diff
@@ -1,6 +1,18 @@
...
+    cache_dict = {}
+
+    def wrapper(*args, **kwargs) -> Any:
+        key = (args, frozenset(kwargs.items()))
+        if key in cache_dict:
+            print("Getting from cache")
+            return cache_dict[key]
+        else:
```
Dimosphen1: There is no need for `else`, the code will not execute after the `return` statement on line 11 anyway.

app/main.py
```diff
@@ -1,6 +1,18 @@
...
+from typing import Callable, Any
 
 
-def cache(func: Callable) -> Callable:
-    # Write your code here
-    pass
+def cache(func: Callable) -> Any:
+    cache_dict = {}
+
+    def wrapper(*args, **kwargs) -> Any:
```

Dimosphen1 (changes requested):

app/main.py
```diff
@@ -1,6 +1,18 @@
...
+        key = (args, frozenset(kwargs.items()))
+        if key in cache_dict:
+            print("Getting from cache")
+            return cache_dict[key]
+        else:
+            print("Calculating new result")
+            result = func(*args, **kwargs)
+            cache_dict[key] = result
+            return result
```
viktoria-rybenchuk: you can simplify the code by using one `return` statement

viktoria-rybenchuk (changes requested):

app/main.py
Dimosphen1: This comment is not resolved, the `else` statement is still present. Remove it from your code.

app/main.py
Dimosphen1: The decorator from `functools` is not used, with removing `else` add it to the `wrapper`.

Dimosphen1 (changes requested):

app/main.py
```diff
@@ -1,6 +1,16 @@
...
-    pass
+    cache_dict = {}
+
+    def wrapper(*args, **kwargs) -> Any:
+        key = (args, frozenset(kwargs.items()))
+        if key in cache_dict:
+            print("Getting from cache")
+            return cache_dict[key]
+        else:
+            print("Calculating new result")
```
Dimosphen1: There is no need for `else`, the code will not execute after the return statement on line 11 anyway. Delete `else` from your code.

app/main.py
```diff
@@ -1,6 +1,16 @@
-from typing import Callable
+from typing import Callable, Any
 
 
 def cache(func: Callable) -> Callable:
-    # Write your code here
-    pass
+    cache_dict = {}
+
+    def wrapper(*args, **kwargs) -> Any:
```
Dimosphen1, line 9: Add a decorator from `functools` to preserve the wrapped function name and docstring.

Dimosphen1 (changes requested):
Resolve comments and update code upon the next review request.

viktoria-rybenchuk (approved):
Good job!

PR #806, opened 2023-11-01
Solution

Molodoirockstar:


ArsenPidhoretskyi (changes requested):
Totally wrong and hard realization.
First of all why data is global argument?
Why data is list, you can use dict

app/main.py
```diff
@@ -1,6 +1,18 @@
...
 def cache(func: Callable) -> Callable:
-    # Write your code here
-    pass
+    data = {}
+
+    def inner(*args: Any) -> Callable:
+        key = (func.__name__, args)
+        if key in data:
+            print("Getting from cache")
+            return data[key]
```
vasylhnatiuk: Try to avoid this return

vasylhnatiuk (changes requested):

app/main.py
```diff
@@ -1,6 +1,18 @@
...
+from typing import Callable, Any
 
 
 def cache(func: Callable) -> Callable:
-    # Write your code here
-    pass
+    data = {}
+
+    def inner(*args: Any) -> Callable:
+        key = (func.__name__, args)
```
ArsenPidhoretskyi: No need to store func__name__ as key, remove it

ArsenPidhoretskyi (changes requested):

app/main.py
Molodoirockstar: If they are two different functions but their arguments are the same, how can we distinguish the result for them? Or I don't understand something?

ArsenPidhoretskyi: Is data will one for all function really? As for me we will create for each function new data dict

ArsenPidhoretskyi (changes requested):

app/main.py
Molodoirockstar: thanks for the explanation)

app/main.py
```diff
@@ -1,6 +1,18 @@
...
-    pass
+    data = {}
+
+    def inner(*args: Any) -> Callable:
+        key = (args)
+        if key in data:
+            print("Getting from cache")
+            result = data[key]
+        else:
+            result = func(*args)
```
ArsenPidhoretskyi: data[key] = func(*args)

And then return will be: return data[key]

ArsenPidhoretskyi (changes requested):

Yurnerroo (approved):
Good job!

PR #807, opened 2023-11-03
cache function created

itpsychologist:


app/main.py
```diff
@@ -2,5 +2,22 @@
...
 def cache(func: Callable) -> Callable:
-    # Write your code here
-    pass
+    cache = []
+
+    def wrapper(*args) -> None:
+        for data in cache:
+            if data["arguments"] == args:
+                print("Getting from cache")
+                return data.get("result")
```
pavlejviki: you can simplify the structure of the cache.

cache can be a dictionary with args as the key and the result of the function as its value.

pavlejviki (changes requested):

app/main.py
itpsychologist: Yes, I understand, but how to be when the argument is a tuple? can it be a dictionary key?

itpsychologist: Got it, I was wrong...

app/main.py
```diff
@@ -2,5 +2,18 @@
...
 def cache(func: Callable) -> Callable:
-    # Write your code here
-    pass
+    cache_dict = {}
+
+    def wrapper(*args) -> None:
+        for cache_key in cache_dict.keys():
+            if cache_key == args:
+                print("Getting from cache")
+                return cache_dict.get(cache_key)
```
viktoria-rybenchuk: You don't need a for loop to check if `args` is a key in the `cache_dict`

viktoria-rybenchuk (changes requested):

app/main.py
```diff
@@ -2,5 +2,17 @@
 
 
 def cache(func: Callable) -> Callable:
-    # Write your code here
-    pass
+    cache_dict = {}
+
+    def wrapper(*args) -> None:
```
vsmutok: add functools.wraps

vsmutok: None wrong  type  annotation

vsmutok (changes requested):

viktoria-rybenchuk (approved):
Good job!

PR #809, opened 2023-11-03
Cache decorator Solution

CANR-95:


app/main.py
```diff
@@ -1,6 +1,23 @@
...
 
 def cache(func: Callable) -> Callable:
-    # Write your code here
-    pass
+    func_cache = {}
+
+    @functools.wraps(func)
+    def wrapper(*args) -> any:
+
+        if args in func_cache.get(func.__name__, ()):
```
EdAlekseiev: `func_cache` is unique for each function, you don't need to use `func.__name__`

EdAlekseiev (changes requested):

CANR-95 (commented):
New Solution

app/main.py
```diff
@@ -1,6 +1,20 @@
...
+    @functools.wraps(func)
+    def wrapper(*args) -> any:
+
+        if args in func_cache:
+            print("Getting from cache")
+            return func_cache[args]
+
+        print("Calculating new result")
+        result = func(*args)
+        func_cache[args] = result
```
ArsenPidhoretskyi: write this in one line

ArsenPidhoretskyi (changes requested):

NikOneZ1 (approved):
Well done!

PR #810, opened 2023-11-05
Solution

Taran21V:


app/main.py
```diff
@@ -1,6 +1,16 @@
-from typing import Callable
+from typing import Callable, Any
 
 
 def cache(func: Callable) -> Callable:
-    # Write your code here
-    pass
+    history_def = {}
```
ArsenPidhoretskyi, line 5: Why do you need _def as variable name?

app/main.py
```diff
@@ -1,6 +1,16 @@
...
+
+    def wrapper(*args: Any) -> Callable:
+        if args not in history_def:
+            print("Calculating new result")
+            history_def[args] = func(*args)
+            return history_def[args]
+
+        print("Getting from cache")
+        return history_def[args]
```
ArsenPidhoretskyi, line 14:
```suggestion
        else:
             print("Getting from cache")
        return history_def[args]
```
In this case better to use if-else because return the same

ArsenPidhoretskyi (changes requested):

PR #812, opened 2023-11-07
decorator Cache created

vasyl-main-dev:


app/main.py
```diff
@@ -1,6 +1,16 @@
...
-    pass
+    results = {}
+
+    def inner(*args) -> Any:
+        if args in results:
+            print("Getting from cache")
+            return results[args]
+        print("Calculating new result")
+        temp_res = func(*args)
+        results[args] = temp_res
```
ArsenPidhoretskyi: Write this inline. And use if-else statement(return after it)

ArsenPidhoretskyi (changes requested):

app/main.py
vasyl-main-dev: Sorry, i don't understand. What should i write and how? 
If i write         
 ```
    temp_res = func(*args)
    results[args] = temp_res
```
before if statement i will can't check args in results. 
Also i use return for finishing the function, so i don't need else statement.

Do you mean that i should write 
```
    results[args] = func(*args)
    return results[args]
```
?

Comment (EdAlekseiev) [reply app/main.py]: Yes, I think, that you should write this something like this.
```suggestion
        results[args] = func(*args)
```

EdAlekseiev (changes requested):

EdAlekseiev (approved):
Good job

PR #813, opened 2023-11-07
Solution

gertsovenko:


app/main.py
```diff
@@ -2,5 +2,16 @@
 
 
 def cache(func: Callable) -> Callable:
-    # Write your code here
-    pass
+    items_dic = {}
+
+    def wrapper(*args) -> int:
+        nonlocal items_dic
```
pavlejviki: using nonlocal is redundant in this case

app/main.py
```diff
@@ -2,5 +2,16 @@
...
+        if args not in items_dic.keys():
```
pavlejviki: calling keys() method is redundant,` in ` will call it anyway under the hood

app/main.py
```diff
@@ -2,5 +2,16 @@
...
+            print("Calculating new result")
+            res = func(*args)
+            items_dic[args] = res
+            return res
+        else:
```

pavlejviki (changes requested):

app/main.py
```diff
@@ -2,5 +2,14 @@
...
 
 def cache(func: Callable) -> Callable:
-    # Write your code here
-    pass
+    items_dic = {}
+
+    def wrapper(*args) -> int:
+        if args not in items_dic:
+            print("Calculating new result")
+            res = func(*args)
```
Dimosphen1: Do not use `res` variable name, be more explicit, there is no need for a shortened variable name.

app/main.py
```diff
@@ -2,5 +2,14 @@
 
 
 def cache(func: Callable) -> Callable:
-    # Write your code here
-    pass
+    items_dic = {}
+
+    def wrapper(*args) -> int:
```
Dimosphen1: Add a decorator from the `functools` to preserve the wrapped function name and doc string.

Dimosphen1: Indicate that `wrapper` could return the `Any` type in the return type annotation. The function result potentially is not limited to only integers.

app/main.py
```diff
@@ -2,5 +2,14 @@
 
 
 def cache(func: Callable) -> Callable:
-    # Write your code here
-    pass
+    items_dic = {}
```
Dimosphen1: Rename this variable to `cached_items`, do not add the underlying type without the need.

Dimosphen1 (changes requested):

app/main.py
```diff
@@ -1,6 +1,17 @@
-from typing import Callable
+from typing import Callable, Any
+from functools import wraps
```
Dimosphen1, line 1: Place your imports in alphabetical order next time.

Dimosphen1 (approved):

PR #819, opened 2023-11-12
solve py-cache-decorator

zdmytryszyn:


app/main.py
```diff
@@ -2,5 +2,13 @@
 
 
 def cache(func: Callable) -> Callable:
-    # Write your code here
-    pass
+    stored_results = {}
+
+    def inner(*args) -> Callable:
```
viktoria-rybenchuk: The type annotation for the return type of `inner` should be `Any` instead of `Callable`

viktoria-rybenchuk (changes requested):

pavlejviki (commented):
Good job!

pavlejviki (approved):

PR #820, opened 2023-11-13
Solution

Grotski:


app/main.py
```diff
@@ -1,6 +1,19 @@
...
+
+    @wraps(func)
+    def wrapper(*args, **kwargs) -> Any:
+        cache_key = (args, frozenset(kwargs.items()))
+        if cache_key in cache_memory:
+            print("Getting from cache")
+            return cache_memory[cache_key]
+        else:
+            result = func(*args, **kwargs)
+            cache_memory[cache_key] = result
```
ArsenPidhoretskyi: Write this inline, and after move return from if-else

ArsenPidhoretskyi (changes requested):

app/main.py
Grotski: @ArsenPidhoretskyi Hi, I think I fixed it

EdAlekseiev (approved):
Good job

PR #822, opened 2023-11-15
solution

AlisaSkl:


app/main.py
```diff
@@ -2,5 +2,15 @@
 
 
 def cache(func: Callable) -> Callable:
-    # Write your code here
-    pass
+    arguments_dictionary = {}
+
+    def inner(*args, **kwargs) -> None:
+        arguments = args + tuple(kwargs.items())
```
pavlejviki: better to use (args, frozenset(kwargs.items()))

A tuple is hashable only if all of its elements are hashable. A frozenset is immutable and hashable, regardless of the hashability of its elements.

pavlejviki (changes requested):

app/main.py
```diff
@@ -2,5 +2,15 @@
 
 
 def cache(func: Callable) -> Callable:
-    # Write your code here
-    pass
+    arguments_dictionary = {}
+
+    def inner(*args, **kwargs) -> None:
```
viktoria-rybenchuk: Does this function return `None`?

viktoria-rybenchuk (changes requested):

app/main.py
AlisaSkl: Arguments could be of any type, so I believe it should be Any. I've fixed the code

Polyakiv-Andrey (approved):
Good job!

PR #824, opened 2023-11-17
'Solution'

NPovashevych:


app/main.py
```diff
@@ -2,5 +2,28 @@
...
 
 def cache(func: Callable) -> Callable:
-    # Write your code here
-    pass
+    cache_for_funk = {}
+    cache_for_funk2 = {}
+
+    def inner(*arg) -> Callable:
+        if len(arg) == 3:
+            a, b, c = arg
```
ArsenPidhoretskyi: Why do you do this??????
It is so strange
Your code will not work when function will be:
def func(a):
    pass

def func():
    pass

def func(a, b, c, d):
   pass

Your code must work with all possible data

ArsenPidhoretskyi (changes requested):

app/main.py
NPovashevych: The tests did not pass... And I had to dig into them and "select" the code. The problem was that I was passing key_cache instead of *args to the external function. That is better?

app/main.py
```diff
@@ -1,6 +1,17 @@
-from typing import Callable
+from typing import Callable, Any
 
 
-def cache(func: Callable) -> Callable:
-    # Write your code here
-    pass
+def cache(func: Callable) -> Any:
```
vsmutok: use functools.wraps

app/main.py
```diff
@@ -1,6 +1,17 @@
...
+    cache_for_func = {}
+
+    def inner(*args: Any) -> Any:
```
vsmutok: what happens if the function takes key arguments?

vsmutok: add **kwargs

app/main.py
vsmutok: Wrong  return  type  annotation

vsmutok (changes requested):

app/main.py
```diff
@@ -1,6 +1,19 @@
...
+    @wraps(func)
+    def inner(*args: Any, **kwargs: Any) -> int:
+        key_cache = args if args else kwargs
+        if key_cache in cache_for_func:
+            print("Getting from cache")
+            result = cache_for_func[key_cache]
+        else:
+            print("Calculating new result")
+            result = func(*args, **kwargs)
+            cache_for_func[key_cache] = result
```

ArsenPidhoretskyi (changes requested):

ArsenPidhoretskyi (approved):
Well done!

PR #825, opened 2023-11-18
Solution

pnasonov:


app/main.py
```diff
@@ -1,6 +1,27 @@
 from typing import Callable
+from functools import wraps
 
 
 def cache(func: Callable) -> Callable:
-    # Write your code here
-    pass
+    cached_func_names = []
```
vasylhnatiuk: instead of 3 lists try to use one dict where `args` will be key

vasylhnatiuk (changes requested):

app/main.py
```diff
@@ -1,6 +1,21 @@
...
-    # Write your code here
-    pass
+    cached_funcs = {}
+
+    @wraps(func)
+    def wrapper(*args) -> int:
+        for key_arg in cached_funcs:
+            if args == key_arg:
+                print("Getting from cache")
+                return cached_funcs[args]
```
vsmutok: try writing this without using a loop

vsmutok (changes requested):

app/main.py
vsmutok: try to use `if args  in cached_funcs`

vsmutok (changes requested):

vsmutok (approved):
Good Job 👍

PR #826, opened 2023-11-19
Solution

Vlad1slavo:


app/main.py
```diff
@@ -1,6 +1,18 @@
-from typing import Callable
+from typing import Callable, Any
 
 
 def cache(func: Callable) -> Callable:
-    # Write your code here
+    cache_dict = {}
+
+    def wrapper(*args: Any) -> Any:
```
vasylhnatiuk, line 7: try to use only one `return` in def wrapper

vasylhnatiuk (changes requested):

app/main.py
```diff
@@ -1,6 +1,18 @@
...
+        if args in cache_dict:
+            print("Getting from cache")
+            result = cache_dict[args]
+        else:
+            print("Calculating new result")
+            result = func(*args)
+            cache_dict[args] = result
```
ArsenPidhoretskyi, line 14: Write this inline

app/main.py
```diff
@@ -1,6 +1,18 @@
...
+        return result
```
```suggestion
        return cache_dict[args]
```

app/main.py
```diff
@@ -1,6 +1,18 @@
...
 
 
 def cache(func: Callable) -> Callable:
-    # Write your code here
+    cache_dict = {}
+
+    def wrapper(*args: Any) -> Any:
+        if args in cache_dict:
+            print("Getting from cache")
+            result = cache_dict[args]
```
ArsenPidhoretskyi, line 10: This line redundant

ArsenPidhoretskyi (changes requested):

PR #827, opened 2023-11-19
'Solution'

anastasiia-yavorska:


app/main.py
```diff
@@ -2,5 +2,18 @@
 
 
 def cache(func: Callable) -> Callable:
-    # Write your code here
-    pass
+
+    results = []
+
```
vsmutok, line 8: add `functools.wraps`

app/main.py
```diff
@@ -2,5 +2,18 @@
 
 
 def cache(func: Callable) -> Callable:
-    # Write your code here
-    pass
+
+    results = []
```
```suggestion
    cache_dictionary = {}
```
better to use dict

app/main.py
```diff
@@ -2,5 +2,18 @@
...
+
+    def wrapper(*args) -> any:
+        for result in results:
+            if result["func"] == func and result["args"] == args:
+                print("Getting from cache")
+                return result["output"]
```
vsmutok: no need to use loop

vsmutok: just write args as the key in the dictionary, and the value is the result of the function

vsmutok (changes requested):

app/main.py
```diff
@@ -1,6 +1,19 @@
...
 
 def cache(func: Callable) -> Callable:
-    # Write your code here
-    pass
+
+    cache_dictionary = {}
+
+    @wraps(func)
+    def wrapper(*args) -> any:
+        if args in cache_dictionary.keys():
```
```suggestion
        if args in cache_dictionary:
```

`keys()` by default

vsmutok (changes requested):

app/main.py
```diff
@@ -1,6 +1,19 @@
...
 
 
 def cache(func: Callable) -> Callable:
-    # Write your code here
-    pass
+
+    cache_dictionary = {}
+
+    @wraps(func)
+    def wrapper(*args) -> any:
```
Dimosphen1, line 10: It would be better to use `Any` from `typing` for type annotation instead of `any` since it is not deprecated.

Dimosphen1 (approved):

PR #828, opened 2023-11-20
Solution

Roman-Sokolov-V:


app/main.py
```diff
@@ -2,5 +2,19 @@
...
+
+        print("Calculating new result")
+        result = func(*args)
+        cash.append([func, args, result])
+        return result
+
+    return inner
+
+
+cash = [[None, None, None]]
```
pavlejviki: cache  is a global variable, which is not a good practice as it can lead to conflicts and unintended side-effects. It is better to encapsulate the cache within the decorator itself.

pavlejviki: cache is implemented as a list of lists, which is inefficient for lookups. A dictionary would be more appropriate

pavlejviki (changes requested):

app/main.py
```diff
@@ -2,5 +2,16 @@
...
+
+    def inner(*args) -> int:
+        if cash.get(f"({func},{args})") is None:
+            result = func(*args)
+            cash[f"({func},{args})"] = result
+            print("Calculating new result")
+            return result
+
+        print("Getting from cache")
+        return cash.get(f"({func},{args})")
```
Dimosphen1: You could avoid multiple conversions to a string of the function with its arguments, allocate a variable at the beginning of the inner logic:
```suggestion
        cached_key = (func, args)
        if cash.get(cached_key) is None:
            result = func(*args)
            cash[cached_key] = result
            print("Calculating new result")
            return result

        print("Getting from cache")
        return cash.get(cached_key)
```

app/main.py
```diff
@@ -2,5 +2,16 @@
 
 
 def cache(func: Callable) -> Callable:
-    # Write your code here
-    pass
+    cash = {}
```
Dimosphen1: `cash` stands for money in coins and notes, this name does not apply to this context. Replace this name with another, e.g. `cache_storage`.

app/main.py
```diff
@@ -2,5 +2,16 @@
...
+
+    def inner(*args) -> int:
```
Dimosphen1: Return type annotation is not correct because this decorator may be applied to functions that return values other than integers. Replace `int` with `Any`.

Dimosphen1: Apply a decorator from `functools` to the `inner` to preserve the wrapped function name and docstring.

Dimosphen1 (changes requested):

app/main.py
```diff
@@ -1,6 +1,20 @@
...
 
 
 def cache(func: Callable) -> Callable:
-    # Write your code here
-    pass
+    cache_storage = {}
+
+    @wraps(func)
+    def inner(*args) -> Any:
+        cached_key = (func, args)
```
ArsenPidhoretskyi, line 10: Why do you need store func?

app/main.py
```diff
@@ -1,6 +1,20 @@
...
+        if cache_storage.get(cached_key) is None:
+            result = func(*args)
+            cache_storage[cached_key] = result
```

ArsenPidhoretskyi (changes requested):

app/main.py
Roman-Sokolov-V: there is no "store func". If it's about @wraps(func), I've inserted as requested by Dimosphen1
"Apply a decorator from functools to the inner to preserve the wrapped function name and docstring."

Dimosphen1: @Roman-Sokolov-V @ArsenPidhoretskyi asked about the need to store `func` inside of `cached_key` rather than about using `wraps` decorator. Since storing function inside of the cache is not obligatory in this task, it was a point of this question. However, we had a discussion during a Q&A session where we decided to store `func` explicitly to prevent cases when two different functions may be decorated using `cache` and have the same arguments but different logic so that they would not contradict one another.

Dimosphen1 (approved):
LGTM, well done!

PR #829, opened 2023-11-21
Solution

ViaMal:


app/main.py
```diff
@@ -1,6 +1,17 @@
...
+
+    def inner(*args, **kwargs) -> Any:
+        key = (args, frozenset(kwargs.items()))
+        if key not in cache_dict:
+            cache_dict[key] = func(*args, **kwargs)
+            print("Calculating new result")
+            return cache_dict[key]
+        else:
+            print("Getting from cache")
+            return cache_dict[key]
```
ArsenPidhoretskyi: return the same in if and else statement move it after if-else

ArsenPidhoretskyi (changes requested):

app/main.py
ViaMal: @ArsenPidhoretskyi done!

ArsenPidhoretskyi (approved):
Well done!

PR #830, opened 2023-11-22
make decorator with two lists

angelzagniy:
(I can`t understand how to create ditionary with values of function and results)

### DIFF
--- a/app/main.py
+++ b/app/main.py
@@ -2,5 +2,21 @@ from typing import Callable
 
 
 def cache(func: Callable) -> Callable:
-    # Write your code here
-    pass
+    list_of_functions = []
+    list_of_results = []
+
+    def wrapper(*args: object, **kwargs: object) -> object:
+        if [func.__name__, args, kwargs] in list_of_functions:
+            print("Getting from cache")
+            return list_of_results[
+                list_of_functions.index([func.__name__, args, kwargs])
+            ]
+        else:
+            print("Calculating new result")
+            list_of_functions.append([func.__name__, args, kwargs])
+            list_of_results.append(func(*args, **kwargs))
+            return list_of_results[
+                list_of_functions.index([func.__name__, args, kwargs])
+            ]
+
+    return wrapper
--- a/app/main.py
+++ b/app/main.py
@@ -1,21 +1,20 @@
 from typing import Callable
+from functools import wraps
 
 
 def cache(func: Callable) -> Callable:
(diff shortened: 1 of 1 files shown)


app/main.py
```diff
@@ -2,5 +2,21 @@
 
 
 def cache(func: Callable) -> Callable:
-    # Write your code here
-    pass
+    list_of_functions = []
+    list_of_results = []
+
+    def wrapper(*args: object, **kwargs: object) -> object:
```
Polyakiv-Andrey, line 9: Implement this decarator without list_of_functions

Polyakiv-Andrey (changes requested):

app/main.py
```diff
@@ -2,5 +2,20 @@
...
+            print("Getting from cache")
+        else:
+            print("Calculating new result")
+            dict_of_functions[
+                (func.__name__, args, frozenset(kwargs.items()))
+            ] = func(*args, **kwargs)
+
+        return dict_of_functions[
+            (func.__name__, args, frozenset(kwargs.items()))
+        ]
```
```suggestion
        key = (func.__name__, args, frozenset(kwargs.items())
        if key in dict_of_functions:
            print("Getting from cache")
        else:
            print("Calculating new result")
            dict_of_functions[key] = func(*args, **kwargs)

        return dict_of_functions[kye]
```

app/main.py
```diff
@@ -2,5 +2,20 @@
 
 
 def cache(func: Callable) -> Callable:
-    # Write your code here
-    pass
+    dict_of_functions = {}
+
+    def wrapper(*args: object, **kwargs: object) -> object:
```
vsmutok, line 9: use functools.wraps

vsmutok (changes requested):

app/main.py
```diff
@@ -1,6 +1,20 @@
...
 
 
 def cache(func: Callable) -> Callable:
-    # Write your code here
-    pass
+    dict_of_functions = {}
+
+    @wraps(func)
+    def wrapper(*args: object, **kwargs: object) -> object:
+        key = (func.__name__, args, frozenset(kwargs.items()))
```
ArsenPidhoretskyi: Why do you need to store func.__name__ ? It is redundant for each function dict_of_functions will be created new

ArsenPidhoretskyi (changes requested):

ArsenPidhoretskyi (approved):
Well done!

PR #831, opened 2023-11-23
Solution

OlgaRozd:


app/main.py
```diff
@@ -1,6 +1,16 @@
-from typing import Callable
+from typing import Callable, Any
 
 
 def cache(func: Callable) -> Callable:
-    # Write your code here
-    pass
+    result_dic_args = {}
```
pavlejviki:
```suggestion
    cache = {}
```

app/main.py
```diff
@@ -1,6 +1,16 @@
...
+
+    def wrapper(*args) -> Any:
+        if args in result_dic_args:
+            print("Getting from cache")
+            return result_dic_args[args]
+        else:
```

pavlejviki (changes requested):

app/main.py
OlgaRozd: fixed

app/main.py
OlgaRozd: fixed

Polyakiv-Andrey (approved):
Good job!

PR #836, opened 2023-12-02
Solution

dissom:


vasylhnatiuk (changes requested):
try to use only one return in def wrapper

app/main.py
```diff
@@ -2,5 +2,16 @@
 
 
 def cache(func: Callable) -> Callable:
-    # Write your code here
-    pass
+    cache_dict = {}
+
+    def wrapper(*args, **kwargs) -> None:
```
vsmutok, line 9: add functools.wraps

app/main.py
```diff
@@ -2,5 +2,16 @@
...
+        key = (frozenset(args), frozenset(kwargs.items()))
+        if key not in cache_dict:
+            print("Calculating new result")
+            result = func(*args, **kwargs)
+            cache_dict[key] = result
+        else:
+            print("Getting from cache")
+            result = cache_dict[key]
+        return result
```
vsmutok, line 18:
```suggestion
            cache_dict[key] = func(*args, **kwargs)
        else:
            print("Getting from cache")
        return cache_dict[key]
```

vsmutok (changes requested):

vsmutok (approved):
well done!

PR #837, opened 2023-12-03
Develop

Vasavik-ua:


app/main.py
```diff
@@ -2,5 +2,15 @@
...
-    pass
+    store_results = {}
+
+    def inner(*args) -> str:
+        if args in store_results.keys():
+            print("Getting from cache")
+            return store_results[args]
+        else:
+            func_result = func(*args)
+            store_results.update({args: func_result})
```

```suggestion
            func_result = func(*args)
            store_results[args] = func_result
```

ArsenPidhoretskyi (changes requested):

app/main.py
```diff
@@ -2,5 +2,13 @@
...
 def cache(func: Callable) -> Callable:
-    # Write your code here
-    pass
+    store_results = {}
+
+    def inner(*args) -> str:
+        if args in store_results.keys():
+            print("Getting from cache")
+        else:
+            store_results.update({args: func(*args)})
```
ArsenPidhoretskyi: .update is redundant here
```suggestion
            store_results[args] = func(*args)
```
Simplify and use brackets

ArsenPidhoretskyi (changes requested):

ArsenPidhoretskyi (approved):
Well done!

PR #841, opened 2023-12-05
Solution

rinotokun:


app/main.py
```diff
@@ -1,6 +1,16 @@
...
+from typing import Callable, Any
 
 
 def cache(func: Callable) -> Callable:
-    # Write your code here
-    pass
+    cache_dict = {}
+
+    def wrapper(*args) -> Any:
+        key = ", ".join(map(str, [*args]))
```
Dimosphen1, line 8: You could use `args` as key directly.

app/main.py
```diff
@@ -1,6 +1,16 @@
...
+        cache_value = cache_dict.get(key)
+        if cache_value is not None:
+            print("Getting from cache")
+            return cache_value
```
Dimosphen1, line 12: You could shorten this code with the walrus operator:
```suggestion
        if cache_value := cache_dict.get(key):
            print("Getting from cache")
            return cache_value
```

app/main.py
```diff
@@ -1,6 +1,16 @@
-from typing import Callable
+from typing import Callable, Any
 
 
 def cache(func: Callable) -> Callable:
-    # Write your code here
-    pass
+    cache_dict = {}
+
+    def wrapper(*args) -> Any:
```
Dimosphen1, line 7: You could use the `wraps` decorator from `functools` to preserve the wrapped function name and docstring.

Dimosphen1 (approved):
LGTM, several comments were left for consideration.

PR #845, opened 2023-12-06
Solution

xrkxdd:


app/main.py
```diff
@@ -2,5 +2,15 @@
 
 
 def cache(func: Callable) -> Callable:
-    # Write your code here
-    pass
+
+    storage = {}
+
+    def wrapper(*args) -> Callable:
```
vsmutok: You could use the `wraps` decorator from `functools` to preserve the wrapped function name and docstring.

vsmutok: what if function take key arguments?
```suggestion
    def wrapper(*args, **kwargs) -> Ant:
```

vsmutok (changes requested):

app/main.py
```diff
@@ -1,6 +1,14 @@
-from typing import Callable
+from typing import Callable, Any
 
 
 def cache(func: Callable) -> Callable:
-    # Write your code here
-    pass
+    result = {}
+
+    def wrapper(*args) -> Any:
```
```suggestion
    def wrapper(*args, **kwargs) -> Any:
```
Please note that the function can also accept key arguments

vsmutok (changes requested):

app/main.py
xrkxdd: If it's not correct, could you please provide more details, as I don't understand

app/main.py
```diff
@@ -1,6 +1,14 @@
...
 
 def cache(func: Callable) -> Callable:
-    # Write your code here
-    pass
+    result = {}
+
+    def wrapper(*args, **kwargs) -> Any:
+        if args not in result:
+            print("Calculating new result")
+            result[args] = func(*args)
```
vsmutok: why did you forget  pass `**kwargs` to func

app/main.py
```diff
@@ -1,6 +1,14 @@
...
+from typing import Callable, Any
 
 
 def cache(func: Callable) -> Callable:
-    # Write your code here
-    pass
+    result = {}
+
+    def wrapper(*args, **kwargs) -> Any:
+        if args not in result:
```
vsmutok: you should also check here and  `kwargs`

vsmutok (changes requested):

vasylhnatiuk (changes requested):
Try to use only in inner function

app/main.py
```diff
@@ -2,15 +2,19 @@
...
+
         if str(args) in cash_list:
             print("Getting from cache")
             return cash_list[str(args)]
         else:
             result = func(*args)
             cash_list[str(args)] = result
             print("Calculating new result")
+
+            inner.cash_list = cash_list
```
vasylhnatiuk: what is it?

app/main.py
```diff
@@ -2,5 +2,19 @@
...
+            print("Getting from cache")
+            return cash_list[str(args)]
+        else:
+            result = func(*args)
+            cash_list[str(args)] = result
+            print("Calculating new result")
+
+            inner.cash_list = cash_list
+
+            return result
```
```suggestion
        return cash_list[str(args)]
```

app/main.py
```diff
@@ -2,5 +2,19 @@
...
-    pass
+    def inner(*args) -> int:
+        cash_list = getattr(inner, "cash_list", {})
+
+        if str(args) in cash_list:
+            print("Getting from cache")
+            return cash_list[str(args)]
+        else:
+            result = func(*args)
+            cash_list[str(args)] = result
```
```suggestion
            cash_list[str(args)] = func(*args)
```

app/main.py
```diff
@@ -2,5 +2,19 @@
...
 
 def cache(func: Callable) -> Callable:
-    # Write your code here
-    pass
+    def inner(*args) -> int:
+        cash_list = getattr(inner, "cash_list", {})
+
+        if str(args) in cash_list:
+            print("Getting from cache")
+            return cash_list[str(args)]
```
vasylhnatiuk: delete this and all of useless lines

app/main.py
vasylhnatiuk: use only this `return` in `inner()`

vasylhnatiuk (changes requested):

vasylhnatiuk (approved):

PR #847, opened 2023-12-06
solution

PythonZem:


Vladyslav-Pliashechnyk (approved):

app/main.py
```diff
@@ -2,5 +2,15 @@
 
 
 def cache(func: Callable) -> Callable:
-    # Write your code here
-    pass
+    cash_list = {}
+
+    def inner(*args) -> int:
+        if str(args) in cash_list:
```
Serhii-Leonenko: for what purpose do you call `str()` method here?)

app/main.py
```diff
@@ -2,5 +2,15 @@
...
+            print("Getting from cache")
+            return cash_list[str(args)]
```
Serhii-Leonenko: try to remove this return and leave only one return in this func

Serhii-Leonenko (changes requested):

app/main.py
PythonZem: Done!

app/main.py
PythonZem: Removed this method :)

Y-Havryliv (approved):
gj!

PR #849, opened 2023-12-07
Solution

StepanSTY:


app/main.py
```diff
@@ -2,5 +2,18 @@
...
+    cache_dict = {}
+
+    def wrapper(*args, **kwargs) -> dict:
+        key = (args, frozenset(kwargs.items()))
+        if key in cache_dict:
+            print("Getting from cache")
+            return cache_dict[key]
+        else:
+            result = func(*args, **kwargs)
+            cache_dict[key] = result
```

ArsenPidhoretskyi (changes requested):

PR #854, opened 2023-12-08
Solution

DanuloM:


app/main.py
```diff
@@ -2,5 +2,13 @@
 
 
 def cache(func: Callable) -> Callable:
-    # Write your code here
-    pass
+    results = {}
+
+    def inner(*args, **kwargs) -> int:
```
pavlejviki: add functools.wraps

pavlejviki: Any will be a more appropriate return type annotation

pavlejviki (changes requested):

app/main.py
```diff
@@ -1,6 +1,16 @@
-from typing import Callable
+from typing import Callable, Any
+from functools import wraps
```
Dimosphen1, line 2: Order your imports in alphabetical order in the future.

Dimosphen1 (approved):
LGTM, well done! One comment was left for consideration.

PR #857, opened 2023-12-11
Solution

Acik33:


app/main.py
```diff
@@ -2,5 +2,20 @@
 
 
 def cache(func: Callable) -> Callable:
-    # Write your code here
-    pass
+
+    result_history = {}
+    input_history = []
```
ArsenPidhoretskyi: This is redundant

app/main.py
```diff
@@ -2,5 +2,20 @@
...
+
+    def wrapper(*args) -> Callable:
+        arguments = ", ".join(str(arg) for arg in args)
```
```suggestion
```
THis line is totally redundant

app/main.py
```diff
@@ -2,5 +2,20 @@
...
+
+        if arguments not in input_history:
+            result = func(*args)
+            input_history.append(arguments)
```
ArsenPidhoretskyi: This line is redundant

app/main.py
```diff
@@ -2,5 +2,20 @@
...
-    # Write your code here
-    pass
+
+    result_history = {}
+    input_history = []
+
+    def wrapper(*args) -> Callable:
+        arguments = ", ".join(str(arg) for arg in args)
+
+        if arguments not in input_history:
```
```suggestion
        if arguments not in result_history:
```

app/main.py
```diff
@@ -2,5 +2,20 @@
...
+            result = func(*args)
+            input_history.append(arguments)
+            result_history[arguments] = result
```
ArsenPidhoretskyi: 13 and 15 line write in one line
and move return after if-else statement

ArsenPidhoretskyi (changes requested):

app/main.py
Acik33: Thank's, it seems i need to review *args and **kwargs properties

Acik33:
Fixed solution with less redundancies was commited

vasylhnatiuk (approved):

PR #860, opened 2023-12-12
Solution

Zonda001:


app/main.py
```diff
@@ -1,6 +1,28 @@
...
+            result = func(*args, **kwargs)
+            results_cache[key] = result
+            print("Calculating new result")
+            return result
+
+    return wrapper
+
+
+@cache
+def long_time_func(time_a: int, time_b: int, time_c: int) -> int:
```
vasylhnatiuk: why do you need this here?

app/main.py
```diff
@@ -1,6 +1,28 @@
...
+    return (time_a ** time_b ** time_c) % (time_a * time_c)
+
+
+@cache
+def long_time_func_2(n_tuple: tuple, power: int) -> int:
```
vasylhnatiuk: why do you need this here?

app/main.py
```diff
@@ -1,6 +1,28 @@
...
 def cache(func: Callable) -> Callable:
-    # Write your code here
-    pass
+    results_cache = {}
+
+    def wrapper(*args: Any, **kwargs: Any) -> Any:
+        key = (args, frozenset(kwargs.items()))
+        if key in results_cache:
+            print("Getting from cache")
+            return results_cache[key]
```
vasylhnatiuk: try to use only one return in  `wrapper`

vasylhnatiuk (changes requested):
try to use only one return in

app/main.py
Zonda001: to call the cache function

app/main.py
vasylhnatiuk: do it

vasylhnatiuk (changes requested):

app/main.py
```diff
@@ -1,6 +1,16 @@
...
-    pass
+    results_cache = {}
+
+    def wrapper(*args: Any, **kwargs: Any) -> Any:
+        key = (args, frozenset(kwargs.items()))
+        if key in results_cache:
+            print("Getting from cache")
+        else:
+            result = func(*args, **kwargs)
+            results_cache[key] = result
```
vasylhnatiuk: this line is useless

vasylhnatiuk (changes requested):

app/main.py
```diff
@@ -1,6 +1,15 @@
-from typing import Callable
+from typing import Callable, Any
 
 
 def cache(func: Callable) -> Callable:
-    # Write your code here
-    pass
+    results_cache = {}
+
+    def wrapper(*args: Any, **kwargs: Any) -> Any:
```
vsmutok: add functools.wraps

vsmutok (changes requested):

vsmutok (approved):
good Job!

PR #864, opened 2023-12-16
decision of cache-decorator

Yevhen23tk:


app/main.py
```diff
@@ -2,5 +2,18 @@
...
+
+    def wrapper(*args, **kwargs) -> None:
+        key = (args, frozenset(kwargs.items()))
+
+        if key in all_result:
+            print("Getting from cache")
+            return all_result[key]
+        else:
+            result = func(*args, **kwargs)
+            all_result[key] = result
```

ArsenPidhoretskyi: And then move return after if-else statement

ArsenPidhoretskyi (changes requested):

app/main.py
Yevhen23tk: ok, accepted

ArsenPidhoretskyi (approved):
The same code in if-else please move return after it

ArsenPidhoretskyi (changes requested):
---

ArsenPidhoretskyi (approved):
Well done!

PR #866, opened 2023-12-17
Solution

annaminaieva:


app/main.py
```diff
@@ -1,6 +1,20 @@
...
 
 
 def cache(func: Callable) -> Callable:
-    # Write your code here
-    pass
+    cache_dict = dict()
+
+    @functools.wraps(func)
+    def cache_wrapper(*args, **kwargs) -> Callable:
+        func_args = args + tuple(f"{k}={v}" for k, v in kwargs.items())
```
ArsenPidhoretskyi: It is so strange to convert kwargs to string

app/main.py
```diff
@@ -1,6 +1,20 @@
...
+
+        if func_args in cache_dict:
+            print("Getting from cache")
+            return cache_dict[func_args]
```
ArsenPidhoretskyi: Move return after if-else

ArsenPidhoretskyi (changes requested):

app/main.py
```diff
@@ -1,6 +1,20 @@
...
+import functools
 
 
 def cache(func: Callable) -> Callable:
-    # Write your code here
-    pass
+    cache_dict = dict()
+
+    @functools.wraps(func)
+    def cache_wrapper(*args, **kwargs) -> Callable:
```
Dimosphen1: The wrapped function does not always return `Callable`, replace the return type annotation with `Any` to have an actual representation.

Dimosphen1 (changes requested):
One change in the type annotation was requested.

viktoria-rybenchuk (approved):
Good job!

PR #867, opened 2023-12-17
'Solution'

JohnnyWalker010:


app/main.py
```diff
@@ -2,5 +2,15 @@
...
 def cache(func: Callable) -> Callable:
-    # Write your code here
-    pass
+    storage = {}
+
+    def wrapper(*args) -> Callable:
+        if args in storage:
+            print("Getting from cache")
+            return storage[args]
+        else:
```
Serhii-Leonenko: In general ok, but this `else` is redundant

Serhii-Leonenko (changes requested):

app/main.py
JohnnyWalker010: I've corrected my code, but somewhy changes were not sent to GH... Trying to understand why

JohnnyWalker010: Seems to me now everything is ok, but I implemented changes after requesting review) But anyway remarks from your side are clear for me. Thanks!

aLEKS-e3 (approved):

NEkropo1 (approved):
Good job!

PR #868, opened 2023-12-17
Solution

MariiaSydoruk:


app/main.py
```diff
@@ -2,5 +2,17 @@
 
 
 def cache(func: Callable) -> Callable:
-    # Write your code here
-    pass
+    funcs = {}
+
+    def wrapper(*args, **kwargs) -> Callable:
```
Dimosphen1: The wrapped function does not always return `Callable`, replace the return type annotation with `Any` to have an actual representation.

app/main.py
```diff
@@ -2,5 +2,17 @@
 
 
 def cache(func: Callable) -> Callable:
-    # Write your code here
-    pass
+    funcs = {}
```
Dimosphen1: `funcs` variable name is not representative enough, replace it with a name that would be more applicable, e.g. `cached_results`.

Dimosphen1 (changes requested):

app/main.py
```diff
@@ -1,6 +1,20 @@
...
+    @wraps(func)
+    def wrapper(*args, **kwargs) -> any:
+        key = (args, frozenset(kwargs.items()))
+        if key not in cached_results:
+            print("Calculating new result")
+            result = func(*args, **kwargs)
+            cached_results[key] = result
+        else:
+            print("Getting from cache")
+            result = cached_results[key]
```
viktoria-rybenchuk: no need save value in result, you can directly return ` cached_results[key]`

viktoria-rybenchuk (changes requested):

app/main.py
```diff
@@ -1,6 +1,18 @@
...
+from functools import wraps
 
 
 def cache(func: Callable) -> Callable:
-    # Write your code here
-    pass
+    cached_results = {}
+
+    @wraps(func)
+    def wrapper(*args, **kwargs) -> any:
```
Dimosphen1, line 9: It is recommended to import `Any` from the `typing` module since it is not deprecated.

Dimosphen1 (approved):

PR #871, opened 2023-12-18
Solution

liudmylaperchevska:


app/main.py
```diff
@@ -1,6 +1,17 @@
-from typing import Callable
+from typing import Callable, Any
 
 
 def cache(func: Callable) -> Callable:
-    # Write your code here
-    pass
+    cache_results = {}
+
+    def wrapper(*args: (int, float, bool, str, bytes, tuple)) -> Any:
```
Dimosphen1, line 7: There is no need to use type annotation for `args` and `kwargs`, remove it from the function signature.

Dimosphen1, line 7: You could add a decorator from `functools` to preserve the wrapped function name and docstring.

app/main.py
```diff
@@ -1,6 +1,17 @@
...
+        key = f"{func.__name__}: {args}"
+        if key in cache_results.keys():
```
Dimosphen1, line 9: There is no need to indicate that iteration should happen over keys, it is a default way of iteration in dict:
```suggestion
        if key in cache_results:
```

app/main.py
```diff
@@ -1,6 +1,17 @@
...
+from typing import Callable, Any
 
 
 def cache(func: Callable) -> Callable:
-    # Write your code here
-    pass
+    cache_results = {}
+
+    def wrapper(*args: (int, float, bool, str, bytes, tuple)) -> Any:
+        key = f"{func.__name__}: {args}"
```
Dimosphen1, line 8: You could keep the whole function as an object to prevent the potential name collision:
```suggestion
        key = (func, args)
```

Dimosphen1 (changes requested):

PR #872, opened 2023-12-19
19_12_23

SerheiZhus:


app/main.py
```diff
@@ -2,5 +2,16 @@
...
+    result_dict = {}
+
+    def inner(*args, **kwargs) -> Callable:
+
+        if args in result_dict:
+            print("Getting from cache")
+            return result_dict[args]
+        else:
+            result = func(*args, **kwargs)
+            result_dict[args] = result
```
ArsenPidhoretskyi: Write this in one line and move return after if-else block

ArsenPidhoretskyi (changes requested):

vasylhnatiuk (approved):

PR #886, opened 2024-01-06
solution

TIgor22:


app/main.py
```diff
@@ -2,5 +2,19 @@
...
+
+        if key in cached_results:
+            print("Getting from cache")
+            return cached_results[key]
+
+        else:
+            result = func(*args)
+            cached_results[key] = result
+            print("Calculating new result")
+            return result
```
```suggestion
        return cached_results[key]
```

app/main.py
```diff
@@ -2,5 +2,19 @@
...
+    def wrapper(*args) -> dict:
+        key = (args)
+
+        if key in cached_results:
+            print("Getting from cache")
+            return cached_results[key]
+
+        else:
+            result = func(*args)
+            cached_results[key] = result
```
```suggestion
            cached_results[key] = func(*args)
```

vasylhnatiuk (changes requested):
use only one return in `def wrapper(*args) -> dict:`

vasylhnatiuk (approved):

PR #889, opened 2024-01-08
Solution

MaksymNikulyak:
In this task, decorators demonstrate their amazing versatility!

### DIFF
--- a/app/main.py
+++ b/app/main.py
@@ -1,6 +1,17 @@
-from typing import Callable
+from typing import Callable, Union
 
 
 def cache(func: Callable) -> Callable:
-    # Write your code here
-    pass
+    history = {}
+
+    def inner(*args: Union[int, tuple]) -> None:
+        # (mx: The use of `|` for type annotations, was added in Python 3.10)
+        if args in history:
+            print("Getting from cache")
+            return history[args]
+        else:
+            print("Calculating new result")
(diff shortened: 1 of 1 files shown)


app/main.py
```diff
@@ -1,6 +1,17 @@
-from typing import Callable
+from typing import Callable, Union
 
 
 def cache(func: Callable) -> Callable:
-    # Write your code here
-    pass
+    history = {}
+
+    def inner(*args: Union[int, tuple]) -> None:
```
```suggestion
    def inner(*args: tuple) -> None:
```

app/main.py
```diff
@@ -1,6 +1,17 @@
...
+        # (mx: The use of `|` for type annotations, was added in Python 3.10)
+        if args in history:
+            print("Getting from cache")
+            return history[args]
+        else:
+            print("Calculating new result")
+            result = func(*args)
+            history[args] = result
+            return result
```
```suggestion
        print("Calculating new result")
        result = func(*args)
        history[args] = result
        return result
```

EdAlekseiev (changes requested):

MaksymNikulyak:
Fixed as per your suggestions. (Correspondingly, also removed the no-longer-necessary `Union` from the import statement.)

EdAlekseiev (approved):
Good job

PR #891, opened 2024-01-08
solution

marta-pylypchuk:


app/main.py
```diff
@@ -1,6 +1,18 @@
...
+from typing import Callable, Any
 
 
 def cache(func: Callable) -> Callable:
-    # Write your code here
-    pass
+    result_cache = {}
+
+    def wrapper(*args, **kwargs) -> Any:
+        key = args + tuple(kwargs.items())
```
kabust, line 8: You can use only args as a dict key. It's not necessary to combine args and kwargs in our case.

kabust (changes requested):

app/main.py
marta-pylypchuk: Just wanted to ensure the cache captures all possible ways as solely using args might miss some variations. Btw, I'm aware that the tests can pass without including kwargs :)

kabust: Gotcha, make sense

Serhii-Leonenko (approved):
Good job!

PR #892, opened 2024-01-09
Solution

Anna728560:


app/main.py
```diff
@@ -1,6 +1,24 @@
-from typing import Callable
+from typing import Callable, Any
 
 
 def cache(func: Callable) -> Callable:
-    # Write your code here
-    pass
+    def wrapper(*args, **kwargs) -> Any:
+        func_name = func.__name__
+        key_args = args
```
marta-pylypchuk: You have kwargs in def wrapper but current implementation only considers positional arguments as keys for caching, without keyword arguments. To include both args and kwargs in the key you can use
`key = args + tuple(kwargs.items())` instead of `key_args = args`

Alternatively, you can just consider removing **kwargs from the function signature :)

app/main.py
```diff
@@ -1,6 +1,24 @@
...
+
+        if func_name not in wrapper.results:
+            wrapper.results[func_name] = {}
+
+        if key_args not in wrapper.results[func_name]:
```
marta-pylypchuk, line 12: replace all instances of key_args with **key**

marta-pylypchuk (changes requested):

app/main.py
```diff
@@ -1,6 +1,24 @@
-from typing import Callable
+from typing import Callable, Any
 
 
 def cache(func: Callable) -> Callable:
-    # Write your code here
-    pass
+    def wrapper(*args, **kwargs) -> Any:
+        func_name = func.__name__
```
Serhii-Leonenko: no need to store func_name in the wrapper.results, because for every function that will be decorated this variable will be its  own

Serhii-Leonenko (changes requested):

app/main.py
```diff
@@ -1,6 +1,20 @@
...
 def cache(func: Callable) -> Callable:
-    # Write your code here
-    pass
+    def wrapper(*args, **kwargs) -> Any:
+        key = args + tuple(kwargs.items())
+
+        if key not in wrapper.results:
+            value_result = func(*args, **kwargs)
+            new_dict = {key: value_result}
+            wrapper.results.update(new_dict)
```
Serhii-Leonenko: simplify these rows, set the key instead of update

app/main.py
```diff
@@ -1,6 +1,20 @@
...
+            print("Calculating new result")
+
+        else:
+            print("Getting from cache")
+
+        return wrapper.results[key]
+
+    wrapper.results = {}
```
Serhii-Leonenko: also, you can close the variable `results` before the wrapper, and it won't be the wrapper's attribute.

here is example 

image

Serhii-Leonenko (changes requested):

app/main.py
```diff
@@ -1,6 +1,20 @@
...
+from typing import Callable, Any
 
 
 def cache(func: Callable) -> Callable:
-    # Write your code here
-    pass
+    results = {}
+
+    def wrapper(*args, **kwargs) -> Any:
+        key = args + tuple(kwargs.items())
```
NEkropo1: better to use frozenset here:
if you check tuple({"a"=12, "b"=14}.items()) with tuple({"b"=14, "a"=12}.items()) it would be False, which is not expected in our case.

NEkropo1 (changes requested):

Y-Havryliv (approved):

PR #896, opened 2024-01-09
Solution

kabust:


app/main.py
```diff
@@ -2,5 +2,14 @@
 
 
 def cache(func: Callable) -> Callable:
-    # Write your code here
-    pass
+    data = {}
+
+    def wrapper(*args, **kwargs) -> int:
```
juliastetsko: return type should be Any

app/main.py
```diff
@@ -2,5 +2,14 @@
...
+        if args in data:
```
juliastetsko: named parameters **kwargs should be also included in data

juliastetsko (changes requested):
Fix return type and **kwargs

Serhii-Leonenko (approved):
Well done!

PR #897, opened 2024-01-09
Solution

tskopenko:


motivated2 (approved):
Good job!

app/main.py
```diff
@@ -1,6 +1,19 @@
...
-    # Write your code here
-    pass
+    cache_storage = {}
+
+    @wraps(func)
+    def inner(*args) -> Any:
+        if args in cache_storage:
+            print("Getting from cache")
+            return cache_storage[args]
+        else:
```
Serhii-Leonenko: this `else` is redundant here because you have `return` before it

Serhii-Leonenko (changes requested):

app/main.py
tskopenko: Fair enough

Serhii-Leonenko (approved):
Well done!

PR #898, opened 2024-01-09
'Solution'

Nick-r2c:


app/main.py
```diff
@@ -1,12 +1,12 @@
 from functools import wraps
-from typing import Callable, Tuple, Union
+from typing import Any, Callable, Tuple, Union
 
 
 def cache(func: Callable) -> Callable:
     results = {}
 
     @wraps(func)
```
PythonZem, line 8: I think that the 8th line is a rudiment. What is it for?

PythonZem (commented):
I think that the 8th line is a rudiment. What is it for?

PythonZem (approved):
I think that the 8th line is a rudiment. What is it for?

app/main.py
Serhii-Leonenko: @PythonZem https://docs.python.org/3/library/functools.html#functools.wraps

app/main.py
```diff
@@ -1,6 +1,20 @@
...
-    pass
+    results = {}
+
+    @wraps(func)
+    def wrapper(*args: Union[int, float, str, bool, Tuple]) -> Any:
+        if args in results:
+            print("Getting from cache")
+
+            return results[args]
+        else:
```
Serhii-Leonenko: else is redundant here because you have return before it

Serhii-Leonenko (changes requested):

Serhii-Leonenko (approved):
Well done!

PR #902, opened 2024-01-09
Solution

Dobryk12:


MaxymChyncha (approved):
Good job!

app/main.py
```diff
@@ -2,5 +2,14 @@
 
 
 def cache(func: Callable) -> Callable:
-    # Write your code here
-    pass
+    caches = {}
+
+    def inner(*args) -> dict:
+        if args in caches.keys():
```
NEkropo1: you don't need explicitly mention keys here, since checking a dict will anyway check its keys

NEkropo1 (changes requested):

app/main.py
```diff
@@ -2,5 +2,14 @@
...
 def cache(func: Callable) -> Callable:
-    # Write your code here
-    pass
+    caches = {}
+
+    def inner(*args) -> dict:
+        if args in caches:
+            print("Getting from cache")
+            return caches[args]
+        else:
```
Serhii-Leonenko, line 11: else is redundant here, remove it

Serhii-Leonenko (approved):

PR #907, opened 2024-01-10
Solution

u123dev:


app/main.py
```diff
@@ -1,6 +1,16 @@
...
+    def wrapper(*args) -> Any:
+        if args in cache_dict :
+            print("Getting from cache")
+            return cache_dict[args]
+        else :
+            print("Calculating new result")
+            result = func(*args)
+            cache_dict[args] = result
+        return result
+    return wrapper
```
YukihiroSM:
```suggestion
    def wrapper(*args) -> Any:
        if args in cache_dict :
            print("Getting from cache")
            return cache_dict[args]
        print("Calculating new result")
        result = func(*args)
        cache_dict[args] = result
        return result
    return wrapper
```

YukihiroSM (changes requested):

app/main.py
YukihiroSM: This else is not needed because you have a return inside the if statement

EdAlekseiev (approved):
Good job

PR #910, opened 2024-01-10
'Solution'

sofiazholud:


app/main.py
```diff
@@ -2,5 +2,15 @@
...
+
+    def wrapper(*args) -> int:
+        if args in cache_dict:
+            print("Getting from cache")
+            return cache_dict[args]
+        else:
+            print("Calculating new result")
+            result = func(*args)
+            cache_dict[args] = result
+            return result
```
YukihiroSM:
```suggestion
        if args in cache_dict:
            print("Getting from cache")
            return cache_dict[args]
        print("Calculating new result")
        result = func(*args)
        cache_dict[args] = result
        return result
```

This else in not needed, because you have return inside if statement

YukihiroSM (changes requested):

EdAlekseiev (approved):
Good job

PR #913, opened 2024-01-11
solution

KovbasinskyV:


app/main.py
```diff
@@ -2,5 +2,15 @@
 
 
 def cache(func: Callable) -> Callable:
-    # Write your code here
-    pass
+    store = {}
+
+    def inner(*args) -> (int, float):
```
EdAlekseiev, line 7: This function can return any value
```suggestion
    def inner(*args) -> Any:
```

EdAlekseiev (approved):
Good job

PR #918, opened 2024-01-15
Solution

volry:


app/main.py
```diff
@@ -2,5 +2,18 @@
 
 
 def cache(func: Callable) -> Callable:
-    # Write your code here
-    pass
+    cache_dict = {}
+
+    def wrapper(*args) -> Callable:
```
NazarSheremet: Good to add an annotation, that this is a wrapped function
```suggestion
    @wraps(func)
    def wrapper(*args) -> Callable:
```

app/main.py
```diff
@@ -2,5 +2,18 @@
```
```suggestion
from functools import wraps
```

NazarSheremet (approved):
Good job! Need to add an annotation

volry:
Solution

PR #920, opened 2024-01-15
Solution

cth-usq:


app/main.py
```diff
@@ -2,5 +2,18 @@
...
 def cache(func: Callable) -> Callable:
-    # Write your code here
-    pass
+
+    cache_dict = {}
+
+    def wrapper(*args) -> int:
+        if args in cache_dict:
+            print("Getting from cache")
+            return cache_dict[args]
```
NazarSheremet: There is no need for two separate returns in if and else, one can be made after the condition.

app/main.py
```diff
@@ -2,5 +2,18 @@
 
 
 def cache(func: Callable) -> Callable:
-    # Write your code here
-    pass
+
+    cache_dict = {}
+
+    def wrapper(*args) -> int:
```
NazarSheremet: Incorrect annotation, function returns not int and good to add an annotation, that this is a wrapped function. 
```suggestion
    @wraps(func)
    def wrapper(*args) -> Callable:
```

app/main.py
```diff
@@ -2,5 +2,18 @@
```
NazarSheremet, line 3:
```suggestion
from functools import wraps
```

NazarSheremet (changes requested):

app/main.py
cth-usq: fixed it

app/main.py
cth-usq: fixed it

NazarSheremet (approved):
Changes are correct. Good job!

PR #922, opened 2024-01-16
py-cache-decorator

Vik-By:


app/main.py
```diff
@@ -1,6 +1,16 @@
-from typing import Callable
+from typing import Callable, Any
 
 
 def cache(func: Callable) -> Callable:
-    # Write your code here
-    pass
+    results = {}
+
+    def repeat_function(*args) -> Any:
```
vasylhnatiuk: strange name for inner function

app/main.py
```diff
@@ -1,6 +1,16 @@
...
+        if args in results:
+            print("Getting from cache")
+            return results[args]
+        else:
+            print("Calculating new result")
+            new_result = func(*args)
+            results[args] = new_result
+            return new_result
```
vasylhnatiuk: you can use only one return in `repeat_function`. do it

vasylhnatiuk (changes requested):

vasylhnatiuk (approved):
Good job

PR #926, opened 2024-01-20
Solution

VasylRybalka:


app/main.py
```diff
@@ -1,6 +1,19 @@
...
+    @wraps(func)
+    def wrapper(*args) -> Any:
+        if args in results_cache:
+            print("Getting from cache")
+            return results_cache[args]
+        else:
+            print("Calculating new result")
+            result = func(*args)
+            results_cache[args] = result
+            return result
```
```suggestion
        print("Calculating new result")
        result = func(*args)
        results_cache[args] = result
        return result
```

EdAlekseiev (changes requested):

EdAlekseiev (changes requested):
v1

app/main.py
VasylRybalka: I done and updated code. Thanks

PR #929, opened 2024-01-20
Solution

kapitoshk4:


vasylhnatiuk (changes requested):
try to use only one return in `wrapper`

EdAlekseiev (approved):
Good job

PR #930, opened 2024-01-22
Solution

vburkalo:


app/main.py
```diff
@@ -1,6 +1,19 @@
...
+        key = (args, frozenset(kwargs.items()))
+
+        if key in cache_store:
+            print("Getting from cache")
+            return cache_store[key]
+        else:
+            result = func(*args, **kwargs)
+            cache_store[key] = result
+            print("Calculating new result")
+            return result
```
EdAlekseiev: Rewrite this function with using `return` statement one time

EdAlekseiev (changes requested):

app/main.py
vburkalo: Okay, boss!

EdAlekseiev (approved):
Good job

PR #939, opened 2024-01-24
create cache decorator

ghost:


app/main.py
```diff
@@ -1,6 +1,19 @@
...
-    # Write your code here
-    pass
+    cache_dict = {}
+
+    def wrapper(*args_1: tuple) -> Any:
+        def save_result_in_cache(*args_2: tuple) -> bool:
+            if args_2 not in cache_dict.keys():
+                cache_dict[args_2] = func(*args_2)
+                return True
+            return False
```
ArsenPidhoretskyi: This function make code more harder

app/main.py
```diff
@@ -1,6 +1,19 @@
...
+        was_calculating = save_result_in_cache(*args_1)
+        print(
+            "Calculating new result" if
+            was_calculating else "Getting from cache"
+        )
+        return cache_dict[*args_1]
```
```suggestion
        return cache_dict[args_1]
```

ArsenPidhoretskyi (changes requested):

app/main.py
ghost: thanks,  it's dizzy task

ArsenPidhoretskyi (approved):
Well done!

PR #944, opened 2024-01-25
Solution

ErikGrinov:


app/main.py
```diff
@@ -1,6 +1,22 @@
...
+            print("Getting from cache")
+            for i in range(len(cache_values)):
+                if key == cache_values[i]:
+                    return result_values[i]
+        else:
+            result = func(*args, **kwargs)
+            cache_values.append(key)
+            print("Calculating new result")
+            result_values.append(result)
+            return result
```
ArsenPidhoretskyi: This code so hard
Try to simplify this you should have a dict to store key args and kwargs as key, and function result as value

ArsenPidhoretskyi (changes requested):

ErikGrinov (commented):
fix hard code

app/main.py
```diff
@@ -1,6 +1,16 @@
-from typing import Callable
+from typing import Callable, Any
 
 
 def cache(func: Callable) -> Callable:
-    # Write your code here
-    pass
+    cache_dict = {}
+
+    def wrapper(*args, **kwargs) -> Any:
```
vasylhnatiuk, line 7: use on;y one return in wrapper()

vasylhnatiuk (changes requested):

vasylhnatiuk (approved):

PR #947, opened 2024-01-27
Solution py-cache-decorator

MilArtem78:


app/main.py
```diff
@@ -2,5 +2,26 @@
...
-    # Write your code here
-    pass
+    result_dict = {}
+
+    def wrapper(*args, **kwargs) -> None:
+        name_func = func.__name__
+        if name_func in result_dict:
+            if args in [arg[0] for arg in result_dict[name_func]]:
+                print("Getting from cache")
+                result_num = [num for num in result_dict[name_func]
```
vasylhnatiuk: why do you need it?

app/main.py
```diff
@@ -2,5 +2,26 @@
...
+                              if num[0] == args][0][-1]
+                return result_num
+            else:
+                print("Calculating new result")
+                result_func = func(*args, **kwargs)
+                result_dict[name_func] = (result_dict.get(name_func, [])
+                                          + [[args] + [result_func]])
+                return result_func
+        else:
```
vasylhnatiuk: don't duplicate yourself

app/main.py
```diff
@@ -2,5 +2,26 @@
 
 
 def cache(func: Callable) -> Callable:
-    # Write your code here
-    pass
+    result_dict = {}
+
+    def wrapper(*args, **kwargs) -> None:
```
vasylhnatiuk: you can use only args - 

        if args in caching_result:

vasylhnatiuk (changes requested):

app/main.py
MilArtem78: Save the results in a dictionary  **result_dict** {'name_func':[(arg),result_num]}.
line 9 - checking the key in the dictionary
line 10 - checking in a dictionary argument with that function name
line 12 - take the result_num of the execution of the function without starting the re-execution of the function

app/main.py
MilArtem78: Ok

app/main.py
MilArtem78: Ok

app/main.py
ArsenPidhoretskyi: In this case, the main question, is result_dict will be the same for all functions? Did you test it?

ArsenPidhoretskyi (changes requested):

app/main.py
MilArtem78: Sorry, the dictionary will not be for all functions, but for each. name_func no need to save.

app/main.py
```diff
@@ -2,5 +2,14 @@
 
 
 def cache(func: Callable) -> Callable:
-    # Write your code here
-    pass
+    caching_result = {}
+
+    def wrapper(*args) -> Callable:
```

vsmutok, line 9:
```suggestion
    def wrapper(*args, **kwargs) -> Callable:
```

app/main.py
```diff
@@ -2,5 +2,14 @@
...
+        if args in caching_result:
```
vsmutok, line 10: also you must check kwargs

app/main.py
```diff
@@ -2,5 +2,14 @@
...
+            print("Getting from cache")
+            return caching_result[args][0]
+        print("Calculating new result")
+        result_func = func(*args)
+        caching_result[args] = (caching_result.get(args, []) + [result_func])
+        return result_func
```
```suggestion
            return caching_result[args]
        print("Calculating new result")
        caching_result[args] = func(*args, **kwargs)
        return  caching_result[args
```

vsmutok (changes requested):

app/main.py
MilArtem78: Sorry, but https://github.com/mate-academy/py-cache-decorator/pull/947#discussion_r1468450859

app/main.py
MilArtem78: Sorry, but https://github.com/mate-academy/py-cache-decorator/pull/947#discussion_r1468450859

MilArtem78: Ok

MilArtem78: Sorry, Decorator cache creating for decorating only functions that take immutable arguments.

app/main.py
MilArtem78: Sorry, Decorator cache creating for decorating only functions that take immutable arguments.

vsmutok (approved):

PR #948, opened 2024-01-28
solution

vdzhyranov:


app/main.py
```diff
@@ -2,5 +2,14 @@
 
 
 def cache(func: Callable) -> Callable:
-    # Write your code here
-    pass
+    cache_storage = {}
+
+    def wrapper(*args) -> Callable:
```
NazarSheremet, line 9:
```suggestion
     @wraps(func)
    def wrapper(*args) -> Callable:
```
Good to add annotation, that this is a wrapped function.

NazarSheremet (changes requested):
Need to make minor changes

NazarSheremet (approved):
Good job!

PR #954, opened 2024-02-01
'Solution'

shramkomax:


app/main.py
```diff
@@ -2,5 +2,27 @@
...
+
+
+@cache
+def long_time_func(aa: int, bb: int, cc: int) -> int:
+    return (aa ** bb ** cc) % (aa * cc)
+
+
+@cache
+def long_time_func_2(n_tuple: tuple, power: int) -> int:
+    return [number ** power for number in n_tuple]
```
NazarSheremet: No need to add these functions to the task solution.

app/main.py
```diff
@@ -2,5 +2,27 @@
 
 
 def cache(func: Callable) -> Callable:
-    # Write your code here
-    pass
+    results_cache = {}
+
+    def wrapper(*args, **kwargs) -> Callable:
```
NazarSheremet, line 9: Good to add an annotation, that this is a wrapped function.
```suggestion
    @wraps(func)
    def wrapper(*args, **kwargs) -> Callable:
```

NazarSheremet (changes requested):
Minor changes

NazarSheremet (approved):
Good job!

PR #961, opened 2024-02-04
Implement cache-decorator

anrozhna:


app/main.py
```diff
@@ -1,6 +1,18 @@
 from typing import Callable
 
 
+cash = []
+
+
 def cache(func: Callable) -> Callable:
-    # Write your code here
-    pass
+    def wrapper(*args) -> int:
```
viktoria-rybenchuk: decorator should contains `kwargs` as well

app/main.py
```diff
@@ -1,6 +1,18 @@
 from typing import Callable
 
 
+cash = []
```
viktoria-rybenchuk: cash shouldn't be the global variable. Save this variable in the decorator and  better define cash as dict for better memory performance

app/main.py
```diff
@@ -1,6 +1,18 @@
...
+
+
 def cache(func: Callable) -> Callable:
-    # Write your code here
-    pass
+    def wrapper(*args) -> int:
+        for item in cash:
+            if item[0] == func and item[1] == args:
+                print("Getting from cache")
+                return item[2]
```
viktoria-rybenchuk: change is logic according to the cash as dict

viktoria-rybenchuk (changes requested):

app/main.py
```diff
@@ -2,5 +2,16 @@
 
 
 def cache(func: Callable) -> Callable:
-    # Write your code here
-    pass
+    cash = {}
+
+    def wrapper(*args, **kwargs) -> int:
```
Dimosphen1: You could use a decorator from functools to preserve the wrapped function name and docstring.

Dimosphen1: It is not guaranteed that `wrapper` would return `int`, it could return `Any`. Indicate this in the return type annotation.

app/main.py
```diff
@@ -2,5 +2,16 @@
...
+        for item in cash:
+            if f"{func}, {args}" in cash:
+                print("Getting from cache")
+                return cash[f"{func}, {args}"]
```
Dimosphen1: There is no need to perform iteration, just check the presence of a particular key in dict and return the corresponding value accordingly.

app/main.py
```diff
@@ -2,5 +2,16 @@
...
+        else:
```
Dimosphen1: There is no need for `else`, the code will not execute after the return anyway. Remove it after removing the loop.

Dimosphen1 (changes requested):

pavlejviki (approved):
Good job!

PR #965, opened 2024-02-06
'Solution'

ArtomUmantsev:


app/main.py
```diff
@@ -1,6 +1,19 @@
...
+    def inner(*args, **kwargs) -> Any:
+        key = (args, frozenset(kwargs.items()))
+        if key in cache_info:
+            print("Getting from cache")
+            return cache_info[key]
+        else:
+            result = func(*args, **kwargs)
+            cache_info[key] = result
+            print("Calculating new result")
+            return result
```
EdAlekseiev, line 18: Rewrite this function with using one `return`

EdAlekseiev (changes requested):

PR #966, opened 2024-02-06
add Solution

Dr-Dron:


app/main.py
```diff
@@ -1,6 +1,18 @@
-from typing import Callable
+from typing import Callable, Any
 
 
 def cache(func: Callable) -> Callable:
-    # Write your code here
-    pass
+    dict_cache = {}
+
+    def wrapper(*args, **kwargs) -> Any:
```

app/main.py
```diff
@@ -1,6 +1,18 @@
...
+        key = (args, tuple(kwargs.items()))
+        if key in dict_cache:
+            print("Getting from cache")
+            return dict_cache[key]
+        else:
```
Dimosphen1: There is no need for `else`, the code will not execute after the `return` anyway. Remove it.

Dimosphen1 (changes requested):

Polyakiv-Andrey (approved):
Good job!

PR #968, opened 2024-02-09
Wrote cache decorator

include031:


app/main.py
```diff
@@ -1,6 +1,22 @@
...
+from functools import wraps
 
 
 def cache(func: Callable) -> Callable:
-    # Write your code here
-    pass
+    cache_dict = {}
+
+    @wraps(func)
+    def inner(*args) -> None:
```
Dimosphen1: `inner` does not return `None`, it could return `Any`. Change your return type annotation.

Dimosphen1: Add `kwargs` to the `inner` signature as well.

app/main.py
```diff
@@ -1,6 +1,22 @@
...
+        if args in cache_dict:
+            print("Getting from cache")
+            return cache_dict[args]
+        print("Calculating new result")
+        add = True
+        for arg in args:
+            if isinstance(arg, (list, set, dict)):
+                add = False
+        if add:
+            cache_dict[args] = func(*args)
```
Dimosphen1: There is no need for these operations, you could write the result of the function execution to the `cache_dict` and return the value by key afterward, there is no need to overcomplicate the logic.

Dimosphen1 (changes requested):

pavlejviki (approved):
Looks good)

PR #972, opened 2024-02-10
Solution

d-kaliuta:


app/main.py
```diff
@@ -1,6 +1,17 @@
...
+
+    def wrapper(*args) -> Any:
+        key = args
+        if key in saved_results:
+            print("Getting from cache")
+            return saved_results[key]
+        else:
+            print("Calculating new result")
+            result = func(*args)
+            saved_results[key] = result
```
And also move return after if-else statement

ArsenPidhoretskyi (changes requested):

YukihiroSM (approved):
Good job!

PR #973, opened 2024-02-10
'Solution'

Meyertal:


app/main.py
```diff
@@ -2,5 +2,15 @@
...
 
 def cache(func: Callable) -> Callable:
-    # Write your code here
-    pass
+    data = {}
+
+    def wrapper(*args) -> Callable:
+        if args in data:
+            print("Getting from cache")
+            return data[args]
```
vasylhnatiuk: use only one return in      def wrapper(*args) -> Callable:

vasylhnatiuk (changes requested):

vasylhnatiuk (approved):

PR #976, opened 2024-02-13
Solution

twonumberfortyfives:


app/main.py
```diff
@@ -2,5 +2,16 @@
...
 def cache(func: Callable) -> Callable:
-    # Write your code here
-    pass
+    cache_dict = {}
+
+    def wrapper(*args, **kwargs) -> None:
+        if args in cache_dict:
+            print("Getting from cache")
+            return cache_dict[args]
+        else:
```
Dimosphen1: There is no need for an `else` statement since the code will not execute after `return` anyway. Remove `else`.

app/main.py
```diff
@@ -2,5 +2,16 @@
 
 
 def cache(func: Callable) -> Callable:
-    # Write your code here
-    pass
+    cache_dict = {}
+
+    def wrapper(*args, **kwargs) -> None:
```

Dimosphen1: `wrapper` is not returning `None` always, it could return `Any`. Change your type annotation to encounter that.

Dimosphen1 (changes requested):

viktoria-rybenchuk (approved):
GJ!

PR #977, opened 2024-02-15
'Solution'

KravchenkoSerhiii:


app/main.py
```diff
@@ -1,6 +1,18 @@
-from typing import Callable
+from typing import Callable, Any
 
 
 def cache(func: Callable) -> Callable:
-    # Write your code here
-    pass
+    cache_storage = {}
+
+    def wrapper(*args, **kwargs) -> Any:
```

app/main.py
```diff
@@ -1,6 +1,18 @@
...
+        key = (args, frozenset(kwargs.items()))
+        if key in cache_storage:
+            print("Getting from cache")
+            return cache_storage[key]
+        else:
```
Dimosphen1: There is no need for `else`, the code will not execute after the `return` statement anyway. Remove redundant `else`

Dimosphen1 (changes requested):

app/main.py
```diff
@@ -2,5 +2,17 @@
 
 
 def cache(func: Callable) -> Callable:
-    # Write your code here
-    pass
+    storage = {}
+
+    def wrapper(*args, **kwargs) -> Callable:
```
pavlejviki: consider using @functools.wraps to follow best practises when working with decorators

pavlejviki (changes requested):

app/main.py
```diff
@@ -1,6 +1,18 @@
...
+import functools
 
 
 def cache(func: Callable) -> Callable:
-    # Write your code here
-    pass
+    storage = {}
+
+    @functools.wraps(func)
+    def wrapper(*args, **kwargs) -> Callable:
```
Dimosphen1: The `wrapper` does not return `Callable`, it could return `Any` depending on the logic of the wrapped function. Change your return type annotation to encounter that.

Dimosphen1 (changes requested):

viktoria-rybenchuk (approved):
Good job!

PR #979, opened 2024-02-15
Solution

rinarice:


app/main.py
```diff
@@ -2,5 +2,15 @@
...
+    cached_results = {}
+
+    def wrapper(*args) -> None:
+        if args not in cached_results:
+            cached_results[args] = func(*args)
+            print("Calculating new result")
+            return cached_results[args]
+        else:
+            print("Getting from cache")
+            return cached_results[args]
```
vasylhnatiuk: use only one return in  def wrapper(*args)

vasylhnatiuk (changes requested):

vasylhnatiuk (approved):
Well done!

PR #980, opened 2024-02-16
Solution of task

Maria145:


app/main.py
```diff
@@ -2,5 +2,16 @@
 
 
 def cache(func: Callable) -> Callable:
-    # Write your code here
-    pass
+    results = {}
+
+    def wrapper(*args) -> str:
```
Dimosphen1: `wrapper` does not always return `str`, it could return `Any`. Indicate this in the return type annotation.

Dimosphen1: You could apply a decorator from `functools` to preserve the wrapped function name and docstring.

app/main.py
```diff
@@ -2,5 +2,16 @@
...
+        if args in results:
+            print("Getting from cache")
+            return results[args]
+        else:
```
Dimosphen1: There is no need for `else` since the code will not execute after the `return` statement anyway. Remove redundant `else`.

Dimosphen1 (changes requested):

app/main.py
```diff
@@ -1,6 +1,17 @@
-from typing import Callable
+from typing import Callable, Any
 
 
 def cache(func: Callable) -> Callable:
-    # Write your code here
-    pass
+    results = {}
+
+    def wrapper(*args) -> Any:
```
viktoria-rybenchuk, line 9: This comment `You could apply a decorator from functools to preserve the wrapped function name and docstring.` was not resolved

viktoria-rybenchuk (changes requested):

viktoria-rybenchuk (approved):
Well done!

PR #982, opened 2024-02-16
Solution

ILyakhova:


app/main.py
```diff
@@ -2,5 +2,16 @@
 
 
 def cache(func: Callable) -> Callable:
-    # Write your code here
-    pass
+    results_value_cashe = {}
+
+    def wrapper(*args, **kwargs) -> Callable:
```
NazarSheremet: Good to add an annotation, that this is a wrapped function.
```suggestion
    @wraps(func)
    def wrapper(*args) -> Callable:
```

NazarSheremet (changes requested):
Need minor changes

app/main.py
ILyakhova: changes were created

NazarSheremet (approved):
Good job!

PR #983, opened 2024-02-17
Solution

jyjuk:


app/main.py
```diff
@@ -1,6 +1,19 @@
...
+    @wraps(func)
+    def inner(*args) -> any:
+        if args in cached_data:
+            print("Getting from cache")
+            return cached_data[args]
+        else:
+            print("Calculating new result")
+            res = func(*args)
+            cached_data[args] = res
+            return res
```
vasylhnatiuk: try to use only one return

vasylhnatiuk (changes requested):

app/main.py
```diff
@@ -1,6 +1,18 @@
...
+
+    @wraps(func)
+    def inner(*args) -> any:
+        if args in cached_data:
+            print("Getting from cache")
+            return cached_data[args]
+        else:
+            print("Calculating new result")
+            cached_data[args] = func(*args)
+            return cached_data[args]
```
viktoria-rybenchuk: You can make this just with one return

viktoria-rybenchuk (changes requested):

viktoria-rybenchuk (approved):
LGTM!

PR #984, opened 2024-02-19
feature: implement solution

Githi54:


app/main.py
```diff
@@ -1,6 +1,20 @@
...
 
 
 def cache(func: Callable) -> Callable:
-    # Write your code here
-    pass
+    cached_results = {}
+
+    def wrapper(*args, **kwargs) -> Any:  # any because unknown return type
+        cache_key = f"args={args} kwargs={kwargs}"
+        nonlocal cached_results
```
LiudmylaKulinchenko: this statement is redundant, as dict is mutable type

app/main.py
```diff
@@ -1,6 +1,20 @@
...
+        if cache_key in cached_results:
+            print("Getting from cache")
+            return cached_results[cache_key]
+
+        result = func(*args, **kwargs)
+        cached_results[cache_key] = result
```

LiudmylaKulinchenko (changes requested):

LiudmylaKulinchenko (approved):
Good job! ☀😉

PR #985, opened 2024-02-20
Solution

skeletronys:


app/main.py
```diff
@@ -1,6 +1,19 @@
...
+from functools import wraps
 
 
 def cache(func: Callable) -> Callable:
-    # Write your code here
-    pass
+    cache_dict = {}
+
+    @wraps(func)
+    def wrapper(*args, **kwargs) -> None:
```
Dimosphen1: `wrapper` does not return only `None`, it could return `Any`. Indicate this in the return type annotation

app/main.py
```diff
@@ -1,6 +1,19 @@
...
+        key = (args, frozenset(kwargs.items()))
+        if key in cache_dict:
+            print("Getting from cache")
+            return cache_dict[key]
+        else:
```
Dimosphen1: There is no need for the `else`, the code will not execute after `return` anyway. Remove redundant `else`

app/main.py
```diff
@@ -1,6 +1,19 @@
 from typing import Callable
+from functools import wraps
```
Dimosphen1: Order your imports in alphabetical order according to PEP-8

Dimosphen1 (changes requested):

app/main.py
viktoria-rybenchuk: Not fixed

viktoria-rybenchuk (changes requested):
Not all problems were resolved.

skeletronys:
> Not all problems were resolved.

Sorry, but I seem to have fixed everything, tell me exactly what I didn't fix

app/main.py
```diff
@@ -1,6 +1,18 @@
-from typing import Callable
+from functools import wraps
+from typing import Callable, Any
 
 
-def cache(func: Callable) -> Callable:
-    # Write your code here
-    pass
+def cache(func: Callable) -> Callable[..., Any]:
```
Dimosphen1, line 5: There is no need to specify that the wrapper could return any argument, this annotation is enough  
```suggestion
def cache(func: Callable) -> Callable:
```

Dimosphen1 (approved):
LGTM, well done! One comment is left for consideration

PR #986, opened 2024-02-21
Solution

Yaschenko1:


app/main.py
```diff
@@ -1,6 +1,18 @@
...
+        if key in wrapper.cache_storage:
+            print("Getting from cache")
+            return wrapper.cache_storage[key]
+
+        print("Calculating new result")
+        result = func(*args, **kwargs)
+        wrapper.cache_storage[key] = result
+        return result
+
+    wrapper.cache_storage = {}
```
viktoria-rybenchuk: By defining `cache_storage = {}` above the wrapper function, you ensure that the cache storage is initialized only once when the cache decorator is called

viktoria-rybenchuk (changes requested):

viktoria-rybenchuk (approved):
Well done!

PR #987, opened 2024-02-21
Solution

kostiukmkalne:
by the way
I changed line num: 22-23 because my flake8 didn`t pass

Thanks!

### DIFF
--- a/app/main.py
+++ b/app/main.py
@@ -1,6 +1,36 @@
-from typing import Callable
+from typing import Callable, Any
 
 
 def cache(func: Callable) -> Callable:
-    # Write your code here
-    pass
+    cached_results = {}
+
+    def wrapper(*args, **kwargs) -> Any:
+        args_tuple = tuple(args)
+        if args_tuple in cached_results:
+            print("Getting from cache")
+            return cached_results[args_tuple]
+        else:
+            print("Calculating new result")
+            result = func(*args, **kwargs)
+            cached_results[args_tuple] = result
+            return result
+
+    return wrapper
+
+
+@cache
+def long_time_func(a_one: int, b_one: int, c_one: int) -> int:
+    return (a_one ** b_one ** c_one) % (a_one * c_one)
+
+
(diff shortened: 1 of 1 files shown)


kostiukmkalne:
> by the way I changed line num: 22-23 because my flake8 didn`t pass
> 
> Thanks!

I mean
I changed the names of the values

it was like that:

```
def long_time_func(a: int, b: int, c: int) -> int:
    return (a ** b ** c) % (a * c)
```

I did:

```
def long_time_func(a_one: int, b_one: int, c_one: int) -> int:
    return (a_one ** b_one ** c_one) % (a_one * c_one)

```

NazarSheremet:
I understand, but need to add this function `long_time_func()` and `long_time_func2()`

NazarSheremet (changes requested):
Good job, `cache() ` is correct, but no need to add  `long_time_func()` and` long_time_func_2() ` functions to solution

kostiukmkalne:
> Good job, `cache() ` is correct, but no need to add `long_time_func()` and`long_time_func_2()` functions to solution

I deleted my redundant code!
is it better now?

app/main.py
```diff
@@ -1,6 +1,28 @@
...
+
+
+@cache
+def long_time_func(a_one: int, b_one: int, c_one: int) -> int:
+    return (a_one ** b_one ** c_one) % (a_one * c_one)
+
+
+@cache
+def long_time_func_2(n_tuple: tuple, power: int) -> int:
+    return [number ** power for number in n_tuple]
```
YuriiSmolii: No need to include this code at all
```suggestion

```

YuriiSmolii (changes requested):
Changes requested.

app/main.py
kostiukmkalne: > No need to include this code at all

Understand!

Thanks!

YuriiSmolii (approved):
LGTM now

PR #989, opened 2024-02-22
solution

xtsvirko:


app/main.py
```diff
@@ -1,6 +1,21 @@
...
 
 
 def cache(func: Callable) -> Callable:
-    # Write your code here
-    pass
+
+    cache_dict = {}
+
+    @wraps(func)
+    def inner(*args) -> Any:
```
vsmutok: What if the function takes key arguments?
Also what if decorator is applied to two different functions that take the same arguments?

vsmutok (changes requested):

app/main.py
```diff
@@ -1,6 +1,23 @@
...
+
+        key = (args, tuple(kwargs.items()))
+
+        if key in cache_dict:
+            print("Getting from cache")
+            return cache_dict[key]
+
+        print("Calculating new result")
+        result = func(*args, **kwargs)
+        cache_dict[key] = result
```
```suggestion
        cache_dict[key] = func(*args, **kwargs)
```

app/main.py
```diff
@@ -1,6 +1,23 @@
...
+    def inner(*args, **kwargs) -> Any:
+
+        key = (args, tuple(kwargs.items()))
+
+        if key in cache_dict:
+            print("Getting from cache")
+            return cache_dict[key]
+
+        print("Calculating new result")
+        result = func(*args, **kwargs)
```
vasylhnatiuk: delete it

vasylhnatiuk (changes requested):

vasylhnatiuk (approved):

PR #990, opened 2024-02-23
Solution cache task

Alex-Plugin:


Alex-Plugin:
My Flake8 test in PyCharm is Ok

app/main.py
```diff
@@ -1,6 +1,18 @@
...
+from functools import wraps
 
 
 def cache(func: Callable) -> Callable:
-    # Write your code here
-    pass
+    storage_dict = {}
+
+    @wraps(func)
+    def wrapper(*args: tuple) -> Any:
```
vsmutok: what if function take key arguments?
don't forgot about **kwargs

app/main.py
```diff
@@ -1,6 +1,18 @@
...
+        if args in storage_dict:
+            print("Getting from cache")
+            return storage_dict[args]
+        else:
+            result = func(*args)
+            storage_dict[*args] = result
```
```suggestion
            storage_dict[args] = result
```

vsmutok (changes requested):

Alex-Plugin:
Hello! Could you please make a fedback, I've implemeted the new decision!

app/main.py
Alex-Plugin: Hello! Could you please make a fedback, I've implemeted the new decision!

Alex-Plugin: Hello! Could you please make a fedback, I've implemeted the new decision!

vasylhnatiuk (changes requested):

Alex-Plugin:
I have shifted one return in the wrapper itself and another one I've left inside the if-condition

app/main.py
```diff
@@ -1,6 +1,18 @@
...
-    # Write your code here
-    pass
+    storage_dict = {}
+
+    @wraps(func)
+    def wrapper(*args: tuple, **kwargs: dict) -> Any:
+        if args in storage_dict:
+            print("Getting from cache")
+            return storage_dict[args]
+        else:
```
viktoria-rybenchuk: `else` is redundant

app/main.py
```diff
@@ -1,6 +1,18 @@
...
 
 
 def cache(func: Callable) -> Callable:
-    # Write your code here
-    pass
+    storage_dict = {}
+
+    @wraps(func)
+    def wrapper(*args: tuple, **kwargs: dict) -> Any:
+        if args in storage_dict:
```
viktoria-rybenchuk: You can use a frozenset to represent the combination of positional arguments and keyword arguments as an immutable collection that can be used as a key

viktoria-rybenchuk (changes requested):

app/main.py
Alex-Plugin: Thanks! Made all your recomendation!

viktoria-rybenchuk (approved):
Well done!

PR #995, opened 2024-02-25
Solution

Viacheslav2024:


app/main.py
```diff
@@ -2,5 +2,18 @@
...
+    results = dict()
+
+    def wrapper_cache(*args, **kwargs) -> int:
+        if args in results:
+            print("Getting from cache")
+
+            return results[args]
+
+        result = func(*args, **kwargs)
+        results[args] = result
```
```suggestion
        results[args] = func(*args, **kwargs)
```

app/main.py
```diff
@@ -2,5 +2,18 @@
...
+        print("Calculating new result")
+
+        return result
```
```suggestion
        return results[args]
```

vasylhnatiuk (changes requested):

app/main.py
Viacheslav2024: Yes, of course! Thanks!

EdAlekseiev (approved):
Good job

PR #996, opened 2024-02-25
Solution

sserkov84:


app/main.py
```diff
@@ -1,6 +1,22 @@
 from typing import Callable
+import functools
+
+
+cached = {}
```
YuriiSmolii: This variable should live inside `cache` function block

app/main.py
```diff
@@ -1,6 +1,22 @@
...
 
 
 def cache(func: Callable) -> Callable:
-    # Write your code here
-    pass
+    @functools.wraps(func)
+    def wrapper(*args) -> int:
+        global cached
+        key = str(id(func)) + "(" + f"{args}" + ")"
```
YuriiSmolii: There is ready to use function for this purpose `(args, frozenset(kwargs.items()))`

app/main.py
```diff
@@ -1,6 +1,22 @@
...
+
+
+cached = {}
 
 
 def cache(func: Callable) -> Callable:
-    # Write your code here
-    pass
+    @functools.wraps(func)
+    def wrapper(*args) -> int:
```
YuriiSmolii: wrapper should take **kwargs as well

app/main.py
```diff
@@ -1,6 +1,22 @@
...
+        key = str(id(func)) + "(" + f"{args}" + ")"
+        # print(key)
+        if key not in cached:
+            print("Calculating new result")
+            result = func(*args)
+            cached[key] = result
+        else:
+            print("Getting from cache")
+            result = cached[key]
+        return result
```
YuriiSmolii: Simplify this code.
In any case do `return cached[key]`

Write this in one line:
```
result = func(*args)
cached[key] = result
```

YuriiSmolii (changes requested):
Changes requested

app/main.py
```diff
@@ -1,6 +1,18 @@
...
 
 def cache(func: Callable) -> Callable:
-    # Write your code here
-    pass
+    cached = {}
+
+    @functools.wraps(func)
+    def wrapper(*args, **kwargs) -> int:
+        key = (str(id(func)) + f"({(args, frozenset(kwargs.items()))})")
+        # print(key)
```
NazarSheremet, line 11: Remove this comment

NazarSheremet (approved):
Approved, but remove unnecessary comment

PR #1000, opened 2024-02-27
'Solution'

Bludger404:


app/main.py
```diff
@@ -1,6 +1,18 @@
...
-    # Write your code here
-    pass
+    cached_results = {}
+
+    def wrapper(*args, **kwargs) -> Any:
+        key = (args, frozenset(kwargs.items()))
+        if key in cached_results:
+            print("Getting from cache")
+            return cached_results[key]
+        else:
```
Dimosphen1: here is no need for `else` since the code will not execute after the `return` statement anyway. Please, remove the redundant `else`

app/main.py
```diff
@@ -1,6 +1,18 @@
-from typing import Callable
+from typing import Callable, Any
 
 
 def cache(func: Callable) -> Callable:
-    # Write your code here
-    pass
+    cached_results = {}
+
+    def wrapper(*args, **kwargs) -> Any:
```
Dimosphen1, line 9: You could use a decorator from `functools` to preserve the wrapped function name and docstring

Dimosphen1 (changes requested):

PR #1006, opened 2024-03-01
'Solution'

AnnaLub:


EdAlekseiev (changes requested):
Fix flake8

app/main.py
```diff
@@ -2,5 +2,16 @@
 
 
 def cache(func: Callable) -> Callable:
-    # Write your code here
-    pass
+    cache_result = {}
+
+    def wrapper(*args, **kwargs) -> str:
```
NikOneZ1, line 9: It s a good practice to use @wraps decorator when defining wrapper

app/main.py
```diff
@@ -2,5 +2,16 @@
...
+        arguments = args + tuple(kwargs.items())
+        if arguments not in cache_result:
+            print("Calculating new result")
+            cache_result[arguments] = func(*args, **kwargs)
+            return cache_result[arguments]
+        else:
+            print("Getting from cache")
+            return cache_result[arguments]
```
NikOneZ1: Try to use return only once

NikOneZ1 (changes requested):

app/main.py
vasylhnatiuk: do it

vasylhnatiuk (changes requested):

vasylhnatiuk (approved):

PR #1007, opened 2024-03-02
Solution

kate-chern:


app/main.py
```diff
@@ -2,5 +2,16 @@
 
 
 def cache(func: Callable) -> Callable:
-    # Write your code here
-    pass
+    result = {}
+
+    def wrapper(*args, **kwargs) -> str:
+        key = str(args)
```
viktoria-rybenchuk: ensure  that both positional arguments (args) and keyword arguments (kwargs) are used , use `frozenset`  for this purpose

app/main.py
```diff
@@ -2,5 +2,16 @@
 
 
 def cache(func: Callable) -> Callable:
-    # Write your code here
-    pass
+    result = {}
```
viktoria-rybenchuk, line 6: Using the `@wraps decorator` is a good practice when defining wrapper functions. This decorator is designed to preserve metadata, such as the function name, docstring, and other attributes,

viktoria-rybenchuk (changes requested):

app/main.py
```diff
@@ -1,6 +1,19 @@
...
+        key = (args, frozenset(kwargs.items()))
+        if key not in result:
+            print("Calculating new result")
+            new_result = func(*args, **kwargs)
+            result[key] = new_result
+        else:
+            print("Getting from cache")
+            return result[key]
+        return new_result
+    return wrapper
```
```suggestion
    def wrapper(*args, **kwargs) -> Any:
        key = (args, frozenset(kwargs.items()))
        if key not in result:
            print("Calculating new result")
            result[key] =  func(*args, **kwargs)
        else:
            print("Getting from cache")
        return result[key]
    return wrapper
```

vsmutok (changes requested):

vsmutok (approved):
Good Job!

PR #1008, opened 2024-03-02
'Solution'

mykola-2024:


app/main.py
```diff
@@ -1,6 +1,19 @@
-from typing import Callable
+from typing import Callable, Any
 
 
 def cache(func: Callable) -> Callable:
-    # Write your code here
-    pass
+    cache_dict = {}
+
+    def wrapper(*args: Any, **kwargs: Any) -> Any:
```
vsmutok, line 9: Using the `@wraps` decorator is a good practice when defining wrapper functions. This decorator is designed to preserve metadata, such as the function name, docstring, and other attributes,

app/main.py
```diff
@@ -1,6 +1,19 @@
...
+        key = (func, args, frozenset(kwargs.items()))
```
```suggestion
        key = (func.__name__, args, frozenset(kwargs.items()))
```

app/main.py
```diff
@@ -1,6 +1,19 @@
...
+
+        if key in cache_dict:
+            print("Getting from cache")
+            return cache_dict[key]
+        else:
+            result = func(*args, **kwargs)
+            print("Calculating new result")
+            cache_dict[key] = result
+            return result
```
```suggestion
        if key in cache_dict:
            print("Getting from cache")
        else:
            cache_dict[key] = func(*args, **kwargs)
            print("Calculating new result")
        return cache_dict[key]
```

vsmutok (changes requested):

mykola-2024:
Added new solution

app/main.py
vasylhnatiuk: do it https://docs.python.org/3/library/functools.html
Знімок екрана 2024-03-05 о 13 22 39

vasylhnatiuk (changes requested):

app/main.py
```diff
@@ -1,6 +1,18 @@
...
 
-def cache(func: Callable) -> Callable:
-    # Write your code here
-    pass
+def cache(func: Callable) -> [Callable]:
+    cache_dict = {}
+
+    @functools.wraps(func)
+    def wrapper(*args: Any, **kwargs: Any) -> Any:
+        key = (func.__name__, args, frozenset(kwargs.items()))
```
ArsenPidhoretskyi, line 10: Why do you need store func.__name__
cache_dict will be created new for each function

ArsenPidhoretskyi (changes requested):

PR #1019, opened 2024-03-09
Solution

OttoLindholm:


app/main.py
```diff
@@ -1,6 +1,16 @@
-from typing import Callable
+from typing import Callable, Any
 
 
 def cache(func: Callable) -> Callable:
-    # Write your code here
-    pass
+    buffer = {}
+
+    def inner(*args) -> Any:
```
viktoria-rybenchuk: you should pass `kwargs` as well

viktoria-rybenchuk: use  `@wraps`to  preserve the metadata of the original function when creating a decorator.

viktoria-rybenchuk (changes requested):

ArsenPidhoretskyi (approved):
Well done!

PR #1020, opened 2024-03-09
"Solution"

poplyasita:


app/main.py
```diff
@@ -1,6 +1,21 @@
...
+import functools
 
 
 def cache(func: Callable) -> Callable:
-    # Write your code here
-    pass
+    result_dict = {}
+
+    @functools.wraps(func)
+    def inner(*args: Any) -> Any:
```
vsmutok: what if func takes key arguments?
don't forget about kwargs

vsmutok: Also, What if two different func takes the same arguments?

vsmutok (changes requested):

app/main.py
poplyasita: About second comment: give a hint please)) Where to look for the solution?  All tests are passed with in previous commit. Any other tries lead to failed tests.

app/main.py
```diff
@@ -1,6 +1,20 @@
...
 
 
 def cache(func: Callable) -> Callable:
-    # Write your code here
-    pass
+    result_dict = {}
+
+    @functools.wraps(func)
+    def inner(*args: Any, **kwargs: Any) -> Any:
+        if args in result_dict:
```
viktoria-rybenchuk: you can use 'frozenset' in order test kwargs in cash as well

app/main.py
```diff
@@ -1,6 +1,20 @@
...
+            print("Getting from cache")
+            result = result_dict[args]
+            return result
```
viktoria-rybenchuk: you can avoid this `return `

viktoria-rybenchuk (changes requested):

viktoria-rybenchuk (approved):
GJ!

PR #1022, opened 2024-03-11
Pochtalon. py-cache-decorator

pochtalon:


app/main.py
```diff
@@ -2,5 +2,16 @@
 
 
 def cache(func: Callable) -> Callable:
-    # Write your code here
-    pass
+    cached_results = {}
+
+    def wrapper(*args) -> int:
```
NazarSheremet: Good to add an annotation, that this is a wrapped function. Also, the function can get `**kwargs` as an argument and it returns not int 
```suggestion
     @functools.wraps(func)
    def wrapper_function(*args, **kwargs) -> Callable:
```

NazarSheremet (changes requested):
Need changes

NazarSheremet (approved):
Good job!

PR #1023, opened 2024-03-11
'Solution'

PVS1905:


app/main.py
```diff
@@ -1,6 +1,17 @@
...
+from typing import Callable, Any
 
 
 def cache(func: Callable) -> Callable:
-    # Write your code here
-    pass
+    result = {}
+
+    def wrapper(*args, **kwargs) -> Any:
+        key = (args, frozenset(kwargs.items()), func.__name__)
```
ArsenPidhoretskyi, line 8: Why do you need store func.__name__ 
result will be create new for each function

ArsenPidhoretskyi (changes requested):

app/main.py
PVS1905: to separate functions with the same arguments

vasylhnatiuk (approved):

PR #1026, opened 2024-03-13
Develop

NinqueRaven13:


app/main.py
```diff
@@ -1,6 +1,17 @@
...
 
 
 def cache(func: Callable) -> Callable:
-    # Write your code here
-    pass
+
+    result = {}
+
+    def inner(*args) -> Any:
+        if tuple(args) in result.keys():
```
```suggestion
        if args in result.keys():
```

vasylhnatiuk: don't use tuple. change everywhere

app/main.py
```diff
@@ -1,6 +1,17 @@
...
+            print("Getting from cache")
+            return result[tuple(args)]
```
vasylhnatiuk: try to use only one return in inner()

vasylhnatiuk (changes requested):

PR #1028, opened 2024-03-13
Develop

panicua:
It was hard, had to learn about ``` function.__name__ ``` method to use an additional key due to functions having same inputs but different results.


panicua:
actually the problem wasn't in the key I believe, I just deleted an additional key ``` func.__name__ ``` and it still works

app/main.py
```diff
@@ -1,6 +1,16 @@
-from typing import Callable
+from typing import Callable, Any
 
 
-def cache(func: Callable) -> Callable:
-    # Write your code here
-    pass
+def cache(func: Callable, hash_storage: dict = None) -> Callable:
+    if hash_storage is None:
```
CharmNiovi: "if" is not necessary, if you leave hash_storage = {} it will work

CharmNiovi (changes requested):

CharmNiovi (approved):
GJ

YuriiKindrat (approved):
GJ!

PR #1030, opened 2024-03-13
Solution

saywin:


app/main.py
```diff
@@ -1,6 +1,18 @@
-from typing import Callable
+from typing import Callable, Any
 
 
 def cache(func: Callable) -> Callable:
-    # Write your code here
-    pass
+    cash_result = {}
+
+    def wrapper(*args) -> Any:
```
vsmutok: what if func takes key arguments?
don't forget about kwargs

vsmutok: Also, What if two different func takes the same arguments?

vsmutok: don't forgot about functools.wraps

vsmutok (changes requested):

app/main.py
saywin: > Also, What if two different func takes the same arguments?

Will return results for each function separately

app/main.py
```diff
@@ -1,6 +1,20 @@
...
+
+    @wraps(func)
+    def wrapper(*args, **kwargs) -> Any:
+        cash_key = tuple(args)
+        if cash_key in cash_result:
+            print("Getting from cache")
+            return cash_result[cash_key]
+
+        result = func(*args, **kwargs)
+        cash_result[cash_key] = result
```
```suggestion
        cash_result[cash_key] = func(*args, **kwargs)
```

app/main.py
```diff
@@ -1,6 +1,20 @@
...
+    cash_result = {}
+
+    @wraps(func)
+    def wrapper(*args, **kwargs) -> Any:
+        cash_key = tuple(args)
+        if cash_key in cash_result:
+            print("Getting from cache")
+            return cash_result[cash_key]
+
+        result = func(*args, **kwargs)
```
vasylhnatiuk: delete it

vasylhnatiuk (changes requested):

vasylhnatiuk (approved):

saywin:
Thanks for checking

PR #1031, opened 2024-03-13
Solution

onyevyerov:


app/main.py
```diff
@@ -1,6 +1,18 @@
...
+    @wraps(func)
+    def inner(*args, **kwargs) -> Any:
+        if args not in cashed_value:
+            print("Calculating new result")
+            result = func(*args, **kwargs)
+            cashed_value[args] = result
+            return result
+        else:
+            print("Getting from cache")
+            return cashed_value[args]
```

Dimosphen1 (changes requested):

PR #1037, opened 2024-03-14
Solution

DmytroHlazyrin:


oleksandra-shershen (approved):
GJ!

app/main.py
```diff
@@ -1,6 +1,21 @@
...
+
+        if args not in storage:
+            print("Calculating new result")
+            result_of_func = func(*args)
+            storage[args] = result_of_func
+            return result_of_func
+
+        else:
+            print("Getting from cache")
+            return storage[args]
```
YuriiKindrat: could u use only one return here ?

YuriiKindrat (changes requested):

app/main.py
DmytroHlazyrin: Yes, sure

YuriiKindrat (approved):
GJ!

PR #1045, opened 2024-03-15
'Solution'

Bielichenko:


app/main.py
```diff
@@ -2,5 +2,16 @@
...
+    result_store = {}
+
+    def wrapper(*args: any) -> any:
+        if args in result_store:
+            print("Getting from cache")
+            return result_store[args]
+        else:
+            print("Calculating new result")
+            res = func(*args)
+            result_store[args] = res
```
LiudmylaKulinchenko: try to avoid creating additional variable `res` -> move return outside of a condition to avoid two returns

LiudmylaKulinchenko (changes requested):

LiudmylaKulinchenko (approved):
Well done! ✨

PR #1046, opened 2024-03-15
Solution

SerhiiMar:


vitalinamal (approved):
I would write cache_dict inside def cache
But in general, looks smart and it works )

app/main.py
```diff
@@ -1,6 +1,20 @@
...
-    pass
+    cache_dict = dict()
+
+    @wraps(func)
+    def wrapper(*args) -> Any:
+        name_of_function = func.__name__
+        if name_of_function not in cache_dict:
+            cache_dict[name_of_function] = dict()
+        if args not in cache_dict[name_of_function]:
+            cache_dict[name_of_function][args] = func(*args)
```
LiudmylaKulinchenko: how about to make a tuple from function name and args, and make it a key in the cache_dict? It will help avoid nested dictionaries

LiudmylaKulinchenko (changes requested):

LiudmylaKulinchenko (approved):
Looks good! 🌼🌷

PR #1048, opened 2024-03-16
Nice Solution

SpecializedBaby:


app/main.py
```diff
@@ -1,6 +1,19 @@
...
-    pass
+    cache_store = {}
+
+    @functools.wraps(func)
+    def inner(*args) -> Any:
+        nonlocal cache_store
+        if args not in cache_store:
+            print("Calculating new result")
+            result = func(*args)
+            cache_store[args] = result
```
ArsenPidhoretskyi: Write this in one line and move return after if-else

ArsenPidhoretskyi (changes requested):

app/main.py
SpecializedBaby: Moved all done

app/main.py
```diff
@@ -1,6 +1,18 @@
...
 
 
 def cache(func: Callable) -> Callable:
-    # Write your code here
-    pass
+    cache_store = {}
+
+    @functools.wraps(func)
+    def inner(*args) -> Any:
+        nonlocal cache_store
```
vsmutok: In this case, the `nonlocal` statement is not necessary because `cache_store` is already declared in the enclosing scope of the `cache` function, and it can be modified without using `nonlocal`. You can simply reference `cache_store` without declaring it as `nonlocal` within the `inner` function.

app/main.py
```diff
@@ -1,6 +1,18 @@
...
+import functools
 
 
 def cache(func: Callable) -> Callable:
-    # Write your code here
-    pass
+    cache_store = {}
+
+    @functools.wraps(func)
+    def inner(*args) -> Any:
```
vsmutok: - What if the function takes key arguments?

- Also what if decorator is applied to two different functions that take the same arguments?

vsmutok (changes requested):

app/main.py
```diff
@@ -1,6 +1,18 @@
...
 
 
 def cache(func: Callable) -> Callable:
-    # Write your code here
-    pass
+    cache_store = {}
+
+    @functools.wraps(func)
+    def inner(*args, **kwargs) -> Any:
+        cache_store
```
vsmutok: why you need the line here?

app/main.py
```diff
@@ -1,6 +1,18 @@
...
+        if args not in cache_store:
+            print("Calculating new result")
+            cache_store[args] = func(*args, **kwargs)
```
vsmutok: why are you only storing `args ` as the key?
what about `kwargs`?

app/main.py
```diff
@@ -1,6 +1,18 @@
...
+import functools
 
 
 def cache(func: Callable) -> Callable:
-    # Write your code here
-    pass
+    cache_store = {}
+
+    @functools.wraps(func)
+    def inner(*args, **kwargs) -> Any:
```
vsmutok, line 9: What if decorator is applied to two different functions that take the same arguments?

vsmutok (changes requested):

app/main.py
```diff
@@ -1,6 +1,24 @@
...
+        elif kwargs and kwargs not in cache_store:
+            print("Calculating new result")
+            cache_store[kwargs] = func(*args)
+            return cache_store[kwargs]
+        elif args:
+            print("Getting from cache")
+            return cache_store[args]
+        elif kwargs:
+            print("Getting from cache")
+            return cache_store[kwargs]
```
ArsenPidhoretskyi: Why so many conditions???
so if i will call function with:
some_func(1, 2, 3)
some_func(a=1, b=2, c=3)
The result will be another?

ArsenPidhoretskyi (changes requested):

app/main.py
SpecializedBaby: yes I know but im crazy

app/main.py
```diff
@@ -1,6 +1,20 @@
...
 
 
 def cache(func: Callable) -> Callable:
-    # Write your code here
-    pass
+    cache_store = {}
+
+    @functools.wraps(func)
+    def inner(*args, **kwargs) -> Any:
+        key = args + tuple(kwargs.values())
```
vsmutok, line 10:
```suggestion
        key = args + tuple(kwargs.values()) + (func.__name__,)
```

app/main.py
```diff
@@ -1,6 +1,20 @@
...
+
+        if key not in cache_store:
+            print("Calculating new result")
+            cache_store[key] = func(*args, **kwargs)
+            return cache_store[key]
+        else:
+            print("Getting from cache")
+            return cache_store[key]
```
```suggestion
        if key not in cache_store:
            print("Calculating new result")
            cache_store[key] = func(*args, **kwargs)
        else:
            print("Getting from cache")
         return cache_store[key]
```

vsmutok (changes requested):

app/main.py
```diff
@@ -1,6 +1,20 @@
...
 
 
 def cache(func: Callable) -> Callable:
-    # Write your code here
-    pass
+    cache_store = {}
+
+    @functools.wraps(func)
+    def inner(*args, **kwargs) -> Any:
+        key = args + tuple(kwargs.values()) + (func.__name__,)
```
```suggestion
        key = args + tuple(kwargs.values())
```
For each function cache_store will be unique

ArsenPidhoretskyi (changes requested):

app/main.py
SpecializedBaby: Yes I think so. 
Approve please my new solut.
Thanks!

vasylhnatiuk (approved):

PR #1049, opened 2024-03-16
Solution#1

VitalyBashkiser:


app/main.py
```diff
@@ -1,6 +1,21 @@
...
+        serialized_args = str(args) + str(kwargs)
+
+        if serialized_args in cached_results:
+            print("Getting from cache")
+            return cached_results[serialized_args]
+        else:
+            print("Calculating new result")
+            result = func(*args, **kwargs)
+            cached_results[serialized_args] = result
+            return result
```
DmytroHlazyrin: it would look better if you changed the sequence of conditions and used a single return.

DmytroHlazyrin (changes requested):

app/main.py
```diff
@@ -1,6 +1,21 @@
...
+    cached_results = {}
+
+    @wraps(func)
+    def wrapper(*args, **kwargs) -> Any:
+        serialized_args = str(args) + str(kwargs)
+
+        if serialized_args not in cached_results:
+            print("Calculating new result")
+            result = func(*args, **kwargs)
+            cached_results[serialized_args] = result
```
LiudmylaKulinchenko: No need to create additional `result` variable

LiudmylaKulinchenko (changes requested):

YuriiKindrat (approved):
GJ!

PR #1050, opened 2024-03-16
Decorator is done

Nyxan:


SerhiiMar (approved):
Everything seems to be ok. Nice work!

app/main.py
```diff
@@ -2,5 +2,17 @@
...
+    def wrapper(*args, **kwargs) -> Callable:
+        key = (args, frozenset(kwargs.items()))
+        if key in cached_results:
+            print("Getting from cache")
+            return cached_results[key]
+        else:
+            print("Calculating new result")
+            result = func(*args, **kwargs)
+            cached_results[key] = result
+            return result
```
YuriiKindrat, line 16: could u use only one return here ?

YuriiKindrat (changes requested):

app/main.py
YuriiKindrat: could u use only one return here ?

YuriiKindrat (changes requested):

LiudmylaKulinchenko (approved):
Looks good! 🌼🌸

PR #1052, opened 2024-03-16
Implement cache function

bwaah1:


VitalyBashkiser (changes requested):
The sequence of actions looks good, but it would look better in one return

VitalyBashkiser (approved):
The code follows PEP 8 style standards and looks clean and tidy, you'll have to see what your mentor says.

app/main.py
```diff
@@ -2,5 +2,16 @@
 
 
 def cache(func: Callable) -> Callable:
-    # Write your code here
-    pass
+    closure_list = {}
```
YuriiKindrat: is it list ?)
it is better to name vars according to their types to avoid misunderstands

app/main.py
```diff
@@ -2,5 +2,16 @@
...
+
+    def inner(*args, **kwargs) -> Callable:
```
YuriiKindrat: does it returns callable ?

YuriiKindrat (changes requested):

YuriiKindrat (approved):
LGTM!

PR #1053, opened 2024-03-17
Solution

siaserb:


SerhiiMar (approved):
Congratulations. Good work!

app/main.py
```diff
@@ -2,5 +2,18 @@
 
 
 def cache(func: Callable) -> Callable:
-    # Write your code here
-    pass
+    cache_res = {}
```
LiudmylaKulinchenko: Please, use whole words as a variable names. It will make them more self-descriptive

app/main.py
```diff
@@ -2,5 +2,18 @@
...
+
+    def inner(*args, **kwargs) -> Callable:
+        key = (args, tuple(kwargs.items()))
```
LiudmylaKulinchenko: Consider using another variable name then `key` to make it more descriptive

LiudmylaKulinchenko (changes requested):

app/main.py
```diff
@@ -2,5 +2,18 @@
 
 
 def cache(func: Callable) -> Callable:
-    # Write your code here
-    pass
+    cache_result = {}
+
+    def inner(*args, **kwargs) -> Callable:
```
YuriiKindrat: does inner return callable ?

YuriiKindrat (changes requested):

app/main.py
```diff
@@ -2,5 +2,18 @@
 
 
 def cache(func: Callable) -> Callable:
-    # Write your code here
-    pass
+    cache_result = {}
+
+    def inner(*args, **kwargs) -> int:
```
YuriiKindrat: are u sure about int, what if function will return str ?
it seems to me, that result of function could be any)

YuriiKindrat (changes requested):

app/main.py
siaserb: image
тут ніби помилка в анотації типу, що повертається в файлі README.md, хіба не list мусить бути?

YuriiKindrat: yeah, u are right there is a typo in readme
but this is just an example of usage, this decorator could be used for any function, so inner func could return any type 
what if we will use this decoratot for function that will return str, dict or smth else, the type annotation will be wrong, for that cases it is better to use "Any" for type annotation

app/main.py
```diff
@@ -2,5 +2,18 @@
 
 
 def cache(func: Callable) -> Callable:
-    # Write your code here
-    pass
+    cache_result = {}
+
+    def inner(*args, **kwargs) -> list[int] | int:
```
YuriiKindrat: wrong type annotation here, read PR conversation

YuriiKindrat (changes requested):

LiudmylaKulinchenko (approved):
Good job 👍

PR #1056, opened 2024-03-18
solved task

RomanDanyl:


app/main.py
```diff
@@ -1,6 +1,20 @@
...
+import functools
 
 
 def cache(func: Callable) -> Callable:
-    # Write your code here
-    pass
+    cached_results = {}
+
+    @functools.wraps(func)
+    def wrapper(*args, **kwargs) -> dict:
```
pavlopro: Please change the return annotation from dict to Any, as we don't know the type of the return result

pavlopro (changes requested):

app/main.py
```diff
@@ -1,6 +1,20 @@
...
-    pass
+    cached_results = {}
+
+    @functools.wraps(func)
+    def wrapper(*args, **kwargs) -> any:
+        key = (args, tuple(sorted(kwargs.items())))
+        if key in cached_results:
+            print("Getting from cache")
+            return cached_results[key]
+        else:
```
Dimosphen1, line 14: There is no need for `else` since the code will not execute after the `return` statement anyway. Please, remove the redundant `else`

app/main.py
```diff
@@ -1,6 +1,20 @@
...
+import functools
 
 
 def cache(func: Callable) -> Callable:
-    # Write your code here
-    pass
+    cached_results = {}
+
+    @functools.wraps(func)
+    def wrapper(*args, **kwargs) -> any:
```
Dimosphen1, line 9: Since `Any` from `typing` is not deprecated, use it here instead of `any`

app/main.py
```diff
@@ -1,6 +1,20 @@
 from typing import Callable
+import functools
```
Dimosphen1, line 2: Order your imports in alphabetical order

Dimosphen1 (changes requested):

PR #1057, opened 2024-03-18
Solution

b4oody:


app/main.py
```diff
@@ -1,6 +1,18 @@
...
 from typing import Callable
 
 
 def cache(func: Callable) -> Callable:
-    # Write your code here
-    pass
+    cache_dict = {}
+
+    @wraps(func)
+    def inner(*args, **kwargs) -> None:
```
pavlopro: Please change the return annotation from None to Any, as we don't know the type of the return result

app/main.py
```diff
@@ -1,6 +1,18 @@
...
+        key = (args, tuple(kwargs.items()))
+        if key not in cache_dict:
+            print("Calculating new result")
+            result = func(*args, **kwargs)
+            cache_dict[key] = result
```
pavlopro: Please pass the result of the function execution directly to dict, instead of creating a separate `result ` variable.

pavlopro (changes requested):

Dimosphen1 (approved):
LGTM, well done!

PR #1068, opened 2024-03-27
implement py-cache-decorator

kola686:


app/main.py
```diff
@@ -1,6 +1,17 @@
...
-    pass
+    def inner(*args: Any) -> Any:
+        global results
+        if (func, args) in results:
+            print("Getting from cache")
+            return results[(func, args)]
+        else:
+            print("Calculating new result")
+            results[(func, args)] = (func(*args))
+            return results[(func, args)]
```
EdAlekseiev: Rewrite with using one return statement

EdAlekseiev (changes requested):

app/main.py
kola686: Done. But now i have two opened pullrequests — this and next in line

app/main.py
```diff
@@ -1,6 +1,17 @@
...
+from typing import Callable, Any
+
+
+results = {}
 
 
 def cache(func: Callable) -> Callable:
-    # Write your code here
-    pass
+    def inner(*args: Any) -> Any:
```
vsmutok: What if the function takes key arguments?

app/main.py
```diff
@@ -1,6 +1,17 @@
...
+        global results
+        if (func, args) in results:
```
```suggestion
        if (func.__name__, args) in results:
```
I think better to check func name

app/main.py
```suggestion
            print("Getting from cache")
        else:
            print("Calculating new result")
            results[(func, args)] = (func(*args))
        return results[(func, args)]
```

vsmutok (changes requested):

app/main.py
kola686: I dont know why but in such case even when i am using _functools_ I have one test failed.

Знімок екрана 2024-03-28 102827

app/main.py
```diff
@@ -1,6 +1,17 @@
...
 
 
 def cache(func: Callable) -> Callable:
-    # Write your code here
-    pass
+    results = {}
+
+    @wraps(func)
+    def inner(*args: Any, **kwargs: Any) -> Any:
+        key = (func.__name__, args, tuple(kwargs.items()))
```
EdAlekseiev: Why do you need name of function in your key? Each decorated function will have separate `results` dict

app/main.py
```diff
@@ -1,6 +1,17 @@
...
+from functools import wraps
 
 
 def cache(func: Callable) -> Callable:
-    # Write your code here
-    pass
+    results = {}
+
+    @wraps(func)
+    def inner(*args: Any, **kwargs: Any) -> Any:
```
EdAlekseiev: You shouldn't provide type hints for `*args` and `**kwargs`

EdAlekseiev (changes requested):

EdAlekseiev (approved):

app/main.py
kola686: Of course, thanks.

app/main.py
kola686: I thought that if different functions with the same arguments were called, I wouldn't get "Getting from cache". But I can see that the code works without it, but I don't understand how... How does the code understand that these are two different functions, having only the arguments as the key and no information about the function name?

PR #1070, opened 2024-03-28
Solution

danilsiv:


app/main.py
```diff
@@ -1,6 +1,22 @@
...
+            if key == kwargs:
+                print("Getting from cache")
+                return cached_results[key]
+        result = func(*args, **kwargs)
+        if len(args):
+            cached_results[args] = result
+        else:
+            cached_results[kwargs] = result
+        print("Calculating new result")
+        return result
```
ArsenPidhoretskyi: This code looks so hard... No need to iterate by all parameters
Also due to task it always accept only args:
So just check if args in cached_results

ArsenPidhoretskyi (changes requested):

app/main.py
```diff
@@ -2,5 +2,15 @@
...
 def cache(func: Callable) -> Callable:
-    # Write your code here
-    pass
+    cached_results = {}
+
+    def wrapper(*args) -> int:
+        for key in cached_results:
+            if key == args:
+                print("Getting from cache")
+                return cached_results[key]
```
ArsenPidhoretskyi: No need to use forloop here,
You can use 
`if args in cached_results:`
instead of forloop

ArsenPidhoretskyi (changes requested):

ArsenPidhoretskyi (approved):
Well done!

PR #1072, opened 2024-03-28
Solution

Chub-andrew:


kssshevchenko (approved):

app/main.py
```diff
@@ -1,6 +1,16 @@
 from typing import Callable
 
 
-def cache(func: Callable) -> Callable:
-    # Write your code here
-    pass
+def cache(func: Callable, cached_results: {} = None) -> Callable:
```
Oleksl888, line 4: Your decorator shouldn't be taking extra params other then a function. So you can simply declare this dict inside of decorator

app/main.py
```diff
@@ -1,6 +1,16 @@
...
+    if cached_results is None:
+        cached_results = {}
+
+    def wrapper(*args) -> None:
```
Oleksl888, line 8: What do we use in decorator to preserve decorate function's name?

Oleksl888, line 8: What would a better return type here? Decorated function will not necessarily return None

Oleksl888 (changes requested):

PR #1073, opened 2024-03-28
Solution

VasylRybalka:


app/main.py
```diff
@@ -1,6 +1,19 @@
...
+from functools import wraps
 
 
 def cache(func: Callable) -> Callable:
-    # Write your code here
-    pass
+    results_cache = {}
+
+    @wraps(func)
+    def wrapper(*args) -> Any:
```
EdAlekseiev: Decorated function can have keyword arguments

EdAlekseiev (changes requested):

app/main.py
VasylRybalka: changes have been made, the code is ready to be reviewed by a mentor

PR #1075, opened 2024-03-31
create cache_decorator

nrjss67:


app/main.py
```diff
@@ -2,5 +2,15 @@
 
 
 def cache(func: Callable) -> Callable:
-    # Write your code here
-    pass
+    cachet = {}
+
+    def wrapper(*args) -> int | str:
```
vsmutok: - What if the function takes key arguments?

- Also what if decorator is applied to two different functions that take the same arguments?

app/main.py
```diff
@@ -2,5 +2,15 @@
 
 
 def cache(func: Callable) -> Callable:
-    # Write your code here
-    pass
+    cachet = {}
```
```suggestion
    cache = {}
```

app/main.py

vsmutok (changes requested):

app/main.py
nrjss67: Thank you for your corrections.
Fixed.

vasylhnatiuk (approved):

PR #1079, opened 2024-04-04
implementation of cache decorator

RuslanLevchuk:


app/main.py
```diff
@@ -2,5 +2,18 @@
 
 
 def cache(func: Callable) -> Callable:
-    # Write your code here
-    pass
+    # cache stores in format [[data-1, result-1] ... [data-n, result-n]]
+    results = []
```
EdAlekseiev: You should use dict here

app/main.py
```diff
@@ -2,5 +2,18 @@
...
+
+    def wrapper(*args, **kwargs) -> int:
+        # check is the input data is present in cache
+        if args in [_[0] for _ in results]:
```
EdAlekseiev: _ is bad name for variable

EdAlekseiev (changes requested):

app/main.py
RuslanLevchuk: Yeah, I forgot that tuples from *args can be as keys in dict. Thanx!

app/main.py
RuslanLevchuk: But in video lessons speaker named underscores as garbage variables. And he told that it is good way to store temporary data

EdAlekseiev: We can use underscore nane for variables if we don't use them in code

app/main.py
```diff
@@ -2,5 +2,18 @@
...
+
+    def wrapper(*args, **kwargs) -> int:
+        # check is the input data is present in cache
+        if args in results.keys():
+            print("Getting from cache")
+            # return result from cache
+            return results[args]
+        # calculate data and adding it to cache if data is not present in cache
+        res = func(*args, **kwargs)
+        results[args] = res
```
EdAlekseiev: What about kwargs?

EdAlekseiev (changes requested):

app/main.py
RuslanLevchuk: Pytest didn't show any problems with named paremeters, so I decided that kwargs can be ignored...

EdAlekseiev (approved):

PR #1080, opened 2024-04-07
implement py-cache-decorator

olenazaritska:


app/main.py
```diff
@@ -1,6 +1,21 @@
-from typing import Callable
+from typing import Callable, Any
 
 
 def cache(func: Callable) -> Callable:
-    # Write your code here
-    pass
+    completed_runs = []
```
viktoria-rybenchuk: It is better to define the cache as a dictionary because dictionaries provide constant time complexity for key-based operations such as insertion, deletion, and retrieval

viktoria-rybenchuk (changes requested):

app/main.py
olenazaritska: thank you for the comment! I changed the code accordingly. I have initially used a list because creating a dict was not an option for me as I have been keeping both args and kwargs in a tuple. Since kwargs packs values as a dictionary, that tuple couldn't work as a dict key: TypeError: unhashable type: 'dict'. Now I removed the kwargs. Please, let me know how would you go about it if kwargs were required too. Thank you in advance :)

app/main.py
```diff
@@ -1,6 +1,15 @@
-from typing import Callable
+from typing import Callable, Any
 
 
 def cache(func: Callable) -> Callable:
-    # Write your code here
-    pass
+    completed_runs = {}
+
+    def wrapper(*args) -> Any:
```
vsmutok: What if functions takes key argument?

vsmutok (changes requested):

ArsenPidhoretskyi (approved):
Well done!

PR #1081, opened 2024-04-08
Create cache dicorator

AndreyGrPy:


app/main.py
```diff
@@ -2,5 +2,16 @@
 
 
 def cache(func: Callable) -> Callable:
-    # Write your code here
-    pass
+    cache_box = {}   # Create dict
+
+    def wrapper(*args, **kwargs) -> None:
+        if args not in cache_box:
```
vsmutok: Why you check only `args`? 
What about `kwargs`?

app/main.py
```diff
@@ -2,5 +2,16 @@
 
 
 def cache(func: Callable) -> Callable:
-    # Write your code here
-    pass
+    cache_box = {}   # Create dict
+
+    def wrapper(*args, **kwargs) -> None:
```

vsmutok (changes requested):

EdAlekseiev (approved):

PR #1088, opened 2024-04-15
Solution

kolya2731:


app/main.py
```diff
@@ -1,6 +1,15 @@
-from typing import Callable
+from typing import Callable, Any
 
 
 def cache(func: Callable) -> Callable:
-    # Write your code here
-    pass
+    cache_dict = {}
+
+    def wrapper(*args, **kwargs) -> Any:
```
Dimosphen1, line 9: You could use a decorator from `functools` to preserve the wrapped function name and docstring. Please, use it here

Dimosphen1 (changes requested):

app/main.py
```diff
@@ -1,6 +1,17 @@
...
 
 
 def cache(func: Callable) -> Callable:
-    # Write your code here
-    pass
+    cache_dict = {}
+
+    @wraps(func)
+    def wrapper(*args, **kwargs) -> Any:
+        if args not in cache_dict:
```
viktoria-rybenchuk: How do you check if the given argument is a keyword argument and whether it is contained within a cache?

viktoria-rybenchuk (changes requested):

pavlopro (approved):
Good Job!

PR #1090, opened 2024-04-16
Solution

AlexTarasov57:


app/main.py
```diff
@@ -2,5 +2,14 @@
 
 
 def cache(func: Callable) -> Callable:
-    # Write your code here
-    pass
+    dict_cache = {}
+
+    def inner(*args) -> tuple:
```
viktoria-rybenchuk: add kwargs as well

app/main.py
```diff
@@ -2,5 +2,14 @@
 
 
 def cache(func: Callable) -> Callable:
-    # Write your code here
-    pass
+    dict_cache = {}
+
```
viktoria-rybenchuk, line 7: add @wraps decorator to provide metadata for your decorator

viktoria-rybenchuk (changes requested):

app/main.py
AlexTarasov57: Ok!

app/main.py
```diff
@@ -1,6 +1,17 @@
...
+    @functools.wraps(func)
+    def wraps(*args, **kwargs) -> tuple:
+        inner_args = args
```
EdAlekseiev: What about kwargs? You should include kwargs to key too

EdAlekseiev (changes requested):

viktoria-rybenchuk (approved):
LGTM!

PR #1091, opened 2024-04-16
triukhan solution

triukhan:


app/main.py
```diff
@@ -1,6 +1,21 @@
...
 def cache(func: Callable) -> Callable:
-    # Write your code here
-    pass
+    storage = []
+
+    def inner(*args: Any) -> Any:
+        for item in storage:
+            if item["arguments"] == args:
+                print("Getting from cache")
+                return item["result"]
```
LiudmylaKulinchenko: Why don't you just change `storage` to a dictionary and store `arguments: result there`?

app/main.py
```diff
@@ -1,6 +1,21 @@
-from typing import Callable
+from typing import Callable, Any
 
 
 def cache(func: Callable) -> Callable:
-    # Write your code here
-    pass
+    storage = []
+
+    def inner(*args: Any) -> Any:
```
LiudmylaKulinchenko: You can skip providing annotations for args, kwargs

LiudmylaKulinchenko (changes requested):

app/main.py
```diff
@@ -1,6 +1,17 @@
...
+
+    def inner(*args) -> Any:
+        if args in storage:
+            print("Getting from cache")
+            return storage[args]
+
+        print("Calculating new result")
+        result = func(*args)
+        storage[args] = result
+        return result
```
LiudmylaKulinchenko: You can avoid creating new variable
Also option to add else block and then you could avoid return duplicating

LiudmylaKulinchenko (changes requested):

LiudmylaKulinchenko (approved):
Well done 🌸

PR #1092, opened 2024-04-17
Обновить test.yml

IrinaLavrik:


.github/workflows/test.yml
```diff
@@ -1,7 +1,7 @@
 name: Test
 
 on:
-  pull_request_target:
+  pull_request:
```
ArsenPidhoretskyi, line 4: Why do you change this?

app/main.py
```diff
@@ -2,5 +2,17 @@
...
+
+    def wrapper(*args, **kwargs):
+        key = (args, tuple(sorted(kwargs.items())))
+        if key in cached_results:
+            print("Getting from cache")
+            return cached_results[key]
+        else:
+            print("Calculating new result")
+            result = func(*args, **kwargs)
+            cached_results[key] = result
```

ArsenPidhoretskyi (changes requested):

YukihiroSM (approved):
Looks good to me!

PR #1096, opened 2024-04-18
Add cache decorator

MRozdoba:


app/main.py
```diff
@@ -1,6 +1,18 @@
...
 
 def cache(func: Callable) -> Callable:
-    # Write your code here
-    pass
+    results = {}
+
+    @functools.wraps(func)
+    def wrapper(*args) -> Callable:
+        nonlocal results
+        if func.__name__ + str(args) in results:
```
viktoria-rybenchuk: you can save to the results args and check if they in cache, no need to use `func.__name__ `

app/main.py
```diff
@@ -1,6 +1,18 @@
...
+            print("Getting from cache")
+            return results[func.__name__ + str(args)]
+        else:
```
viktoria-rybenchuk: else is redundant

app/main.py
viktoria-rybenchuk: fix in all places

app/main.py
```diff
@@ -1,6 +1,18 @@
...
+import functools
 
 
 def cache(func: Callable) -> Callable:
-    # Write your code here
-    pass
+    results = {}
+
+    @functools.wraps(func)
+    def wrapper(*args) -> Callable:
```
viktoria-rybenchuk: provide kwargs as well

viktoria-rybenchuk (changes requested):

Polyakiv-Andrey (approved):
Good job!

PR #1100, opened 2024-04-25
'CacheSolution'

Kagerou4649:


Kagerou4649:
Had to change variables on the line def long_time_func(first: int, second: int, third: int) -> int:
from "a, b, c" to "first, second, third" because flake8 was yelling that it's not okay to make single word variables

app/main.py
```diff
@@ -2,5 +2,30 @@
...
+
+
+@cache
+def long_time_func(first: int, second: int, third: int) -> int:
+    return (first ** second ** third) % (first * third)
+
+
+@cache
+def long_time_func_2(n_tuple: tuple, power: int) -> list:
+    return [number ** power for number in n_tuple]
```
YuriiSmolii: This code excessive, should be removed.

YuriiSmolii (changes requested):
Major change requested.

YuriiSmolii (approved):
LGTM

PR #1104, opened 2024-04-29
Task completed with no errors.

CycleBreaker:
:D

### DIFF
--- a/app/main.py
+++ b/app/main.py
@@ -4,14 +4,14 @@ from typing import Callable
 def cache(func: Callable) -> Callable:
     cache_storage: dict = {}
 
-    def internal(*args) -> Callable:
-        current_arguments: tuple = args
+    def internal(*args, **kwargs) -> Callable:
+        current_arguments: tuple = (args, tuple(kwargs.items()))
         for cached_arguments in cache_storage.keys():
             if current_arguments == cached_arguments:
                 print("Getting from cache")
                 return cache_storage[cached_arguments]
         print("Calculating new result")
-        result = func(*args)
+        result = func(*args, **kwargs)
         cache_storage[current_arguments] = result
         return result
     return internal
--- a/app/main.py
+++ b/app/main.py
@@ -2,5 +2,18 @@ from typing import Callable
 
 
 def cache(func: Callable) -> Callable:
-    # Write your code here
-    pass
+    cache_arguments: list = []
+    cache_results: list = []
+
+    def internal(*args, **kwargs) -> Callable:
+        argument_list: list = [args, kwargs]
+        for entry_index in range(len(cache_arguments)):
+            if argument_list == cache_arguments[entry_index]:
+                print("Getting from cache")
+                return cache_results[entry_index]
+        print("Calculating new result")
+        result = func(*args, **kwargs)
+        cache_arguments.append(argument_list)
+        cache_results.append(result)
+        return result
+    return internal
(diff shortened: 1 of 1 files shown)


app/main.py
```diff
@@ -2,5 +2,18 @@
 
 
 def cache(func: Callable) -> Callable:
-    # Write your code here
-    pass
+    cache_arguments: list = []
+    cache_results: list = []
```
vsmutok: use dict to store cache

vsmutok (changes requested):

app/main.py
CycleBreaker: Updated! I liked the solution with lists more, though, because that's going to be one ugly dictionary for human reading (in case we need such functionality one day).

app/main.py
```diff
@@ -2,5 +2,16 @@
...
+
+    def internal(*args, **kwargs) -> Callable:
+        argument_list: list = [args, kwargs]
+        for cached_result, cached_arguments in cache_storage.items():
+            if argument_list == cached_arguments:
+                print("Getting from cache")
+                return cached_result
+        print("Calculating new result")
+        result = func(*args, **kwargs)
+        cache_storage[result] = argument_list
```
EdAlekseiev: The keys in the dictionary should correspond to the function's arguments, and the values should represent the function's output. This structure allows for efficient retrieval and mapping of function results, making it easy to access the outcomes based on specific inputs.

EdAlekseiev (changes requested):

app/main.py
CycleBreaker: Okay, but that wasn't stated in the task. Updated.

app/main.py
```diff
@@ -2,5 +2,16 @@
 
 
 def cache(func: Callable) -> Callable:
-    # Write your code here
-    pass
+    cache_storage: dict = {}
+
+    def internal(*args) -> Callable:
+        current_arguments: tuple = args
+        for cached_arguments in cache_storage.keys():
```
vsmutok: don't use loop, use `in`

app/main.py
```diff
@@ -2,5 +2,16 @@
 
 
 def cache(func: Callable) -> Callable:
-    # Write your code here
-    pass
+    cache_storage: dict = {}
+
+    def internal(*args) -> Callable:
```
vsmutok: don't forgot about `functools.wraps`

vsmutok: What if function take key arguments?

vsmutok (changes requested):

app/main.py
CycleBreaker: Indeed. Fixed!

app/main.py
```diff
@@ -2,5 +2,16 @@
...
-    # Write your code here
-    pass
+    cache_storage: dict = {}
+
+    def internal(*args, **kwargs) -> Callable:
+        current_arguments: tuple = (args, tuple(kwargs.items()))
+        for cached_arguments in cache_storage.keys():
+            if current_arguments == cached_arguments:
+                print("Getting from cache")
+                return cache_storage[cached_arguments]
```
```suggestion
        if cached_arguments in cache_storage:
            print("Getting from cache")
            return cache_storage[cached_arguments]
```

EdAlekseiev (changes requested):

app/main.py
CycleBreaker: Yeah, that's pretty awkward. Fixed!

ArsenPidhoretskyi (approved):
Well done!

PR #1108, opened 2024-05-01
Solution

SashaKisliy:


app/main.py
```diff
@@ -2,5 +2,14 @@
 
 
 def cache(func: Callable) -> Callable:
-    # Write your code here
-    pass
+    def inner(*args, **kwargs) -> Callable:
+        if not hasattr(func, "storage"):
+            func.storage = {}
```
Serhii-Leonenko, line 7: closure storage in the outer func
`def cache(func: Callable) -> Callable:`
     `storage = {}`
     `def inner(*args, **kwargs):`
        ....
   `

Serhii-Leonenko (changes requested):

PR #1113, opened 2024-05-04
solution

necobra:


app/main.py
```diff
@@ -1,6 +1,25 @@
...
 
 
 def cache(func: Callable) -> Callable:
-    # Write your code here
-    pass
+    cached_data = {}
+
+    def inner(*args) -> Any:
+        is_already_saved = True
+        working_dict = cached_data
```
DevSpaciX: In inner function you have access for cached_data viariable , you do not need to create another one

app/main.py
```diff
@@ -1,6 +1,25 @@
...
+
+        for value in args:
```
DevSpaciX: 1. You do not need to create for loop here 
2. Just check if args not in your cached_data and that way you'll get rid of is_already_saved viariable also

app/main.py
```diff
@@ -1,6 +1,25 @@
...
+            if value not in working_dict:
+                is_already_saved = False
+                working_dict[value] = {}
+            working_dict = working_dict[value]
+
+        if not is_already_saved:
```
DevSpaciX: Like that: if args not in cache_backet -> add func(*args) as a value for cached_data , where args that will be your key

app/main.py
```diff
@@ -1,6 +1,25 @@
...
+            print("Calculating new result")
```
DevSpaciX: It is considered good practice to write prints at the end of the logical chain

app/main.py
```diff
@@ -1,6 +1,25 @@
-from typing import Callable
+from typing import Callable, Any
 
 
 def cache(func: Callable) -> Callable:
-    # Write your code here
-    pass
+    cached_data = {}
+
+    def inner(*args) -> Any:
```
DevSpaciX: And add pls also type hint for args :)

DevSpaciX (changes requested):

DevSpaciX (approved):
Good job!

PR #1114, opened 2024-05-06
'Solution'

Anon920:


app/main.py
```diff
@@ -1,6 +1,25 @@
...
+
+    def inner(*args) -> Any:
+        already_saved = True
+        working_dict = cache_data
+
+        for arg in args:
+            if arg not in working_dict:
+                already_saved = False
+                working_dict[arg] = {}
+            working_dict = working_dict[arg]
```
ArsenPidhoretskyi: Why do you need this loop you  can use:
 if args in cache_data

ArsenPidhoretskyi (changes requested):

ArsenPidhoretskyi (approved):
Well done!

PR #1116, opened 2024-05-09
Solution

taras-tkachuk:


app/main.py
```diff
@@ -1,6 +1,14 @@
-from typing import Callable
+from typing import Callable, Any
 
 
 def cache(func: Callable) -> Callable:
-    # Write your code here
-    pass
+    cache_data = {}
+
+    def inner(*args) -> Any:
```
also, what if function takes key arguments?

vsmutok (changes requested):

ArsenPidhoretskyi (approved):
Well done!

PR #1117, opened 2024-05-11
'Solution'

arsenmarkotskyi:


app/main.py
```diff
@@ -2,5 +2,26 @@
...
+    caches = {}
+
+    def inner(*args) -> None:
+        if args in caches:
+            print("Getting from cache")
+            return caches[args]
+        else:
+            print("Calculating new result")
+            res = func(*args)
+            caches[args] = res
```

ArsenPidhoretskyi (changes requested):

app/main.py
arsenmarkotskyi: do u mean like this?
        else:
            __print("Calculating new result")
            __caches[args] = res = func(*args)
        return res

app/main.py
```diff
@@ -2,5 +2,25 @@
...
+
+
+@cache
+def long_time_func(num_a: int, num_b: int, num_c: int) -> int:
+    return (num_a ** num_b ** num_c) % (num_a * num_c)
+
+
+@cache
+def long_time_func_2(n_tuple: tuple, power: int) -> list:
+    return [number ** power for number in n_tuple]
```
vsmutok: remove this code

app/main.py
```diff
@@ -2,5 +2,25 @@
 
 
 def cache(func: Callable) -> Callable:
-    # Write your code here
-    pass
+    caches = {}
+
+    def inner(*args) -> None:
+        if args in caches:
```
vsmutok: add `functools.wraps`

Also, what if functions take key arguments?

vsmutok (changes requested):

app/main.py
arsenmarkotskyi: @vsmutok what do u mean? i don’t get it

app/main.py
```diff
@@ -1,6 +1,18 @@
...
+import functools
 
 
 def cache(func: Callable) -> Callable:
-    # Write your code here
-    pass
+    caches = {}
+
+    @functools.wraps(func)
+    def inner(*args) -> None:
```
```suggestion
    def inner(*args, **kwargs) -> None:
```
You should consider that the function can accept keyword arguments as well.

app/main.py
```diff
@@ -1,6 +1,18 @@
...
+        if args in caches:
+            print("Getting from cache")
+            return caches[args]
+        else:
+            print("Calculating new result")
+            caches[args] = res = func(*args)
+        return res
```
```suggestion
        else:
            print("Calculating new result")
            caches[args] = func(*args)
        return caches[args]
```

vsmutok (changes requested):

vasylhnatiuk (approved):

PR #1119, opened 2024-05-11
Develop

BohdanKuzik:


app/main.py
```diff
@@ -2,5 +2,32 @@
...
+            result = func(*args)
+            arguments.update({args: result})
+            print("Calculating new result")
+            return result
+        print("Getting from cache")
+        return arguments[args]
+    return wrapper
+
+
+@cache
```
DevSpaciX: I see that you teseted your code , but please remove it before creating PR :) 
Push only solution for task
( Remove all code from 18 to 33 line)

app/main.py
```diff
@@ -2,5 +2,32 @@
 
 
 def cache(func: Callable) -> Callable:
-    # Write your code here
-    pass
+    arguments = {}
```
DevSpaciX: Maybe it’s worth naming the variable something else, considering that it will be our "cache"

app/main.py
```diff
@@ -2,5 +2,32 @@
...
+
+    def wrapper(*args) -> Callable:
+        if args not in arguments and isinstance(args, (int, float, tuple)):
```
DevSpaciX: Is it necessary for checking isisnstance(...) ? Try to remove that check

app/main.py
```diff
@@ -2,5 +2,32 @@
...
+            result = func(*args)
+            arguments.update({args: result})
```
DevSpaciX: update is not often used , for me it is better to use -> cache[args] = result

DevSpaciX (changes requested):

app/main.py
```diff
@@ -2,5 +2,32 @@
...
+            print("Calculating new result")
+            return result
```
DevSpaciX: try to rewrite your decorator not to have `return` duplicated )

DevSpaciX (changes requested):

DevSpaciX (approved):
Good job!

app/main.py
```diff
@@ -2,5 +2,13 @@
 
 
 def cache(func: Callable) -> Callable:
-    # Write your code here
-    pass
+    results_cache = {}
+
+    def wrapper(*args) -> Callable:
```
DevSpaciX, line 7: For future add type hint for *args parameter too )

PR #1120, opened 2024-05-11
'Solution'

RuslanSamsonovych:


app/main.py
```diff
@@ -2,5 +2,17 @@
 
 
 def cache(func: Callable) -> Callable:
-    # Write your code here
-    pass
+    cache_store = {}
+
+    def inner(*args) -> tuple:
```
```suggestion
    def inner(*args, **kwargs) -> tuple:
```
You should consider that the function can accept keyword arguments as well.

vsmutok: add `functools.wraps`

vsmutok (changes requested):

app/main.py
```diff
@@ -1,6 +1,20 @@
...
+    def inner(*args, **kwargs) -> tuple:
+        key = (args, tuple(kwargs.items()))
+        if key in cache_store:
+            print("Getting from cache")
+            return cache_store[key]
+        else:
+            print("Calculating new result")
+            result = func(*args, **kwargs)
+            cache_store[key] = result
+            return result
```
vasylhnatiuk: use only one return

vasylhnatiuk:
```suggestion
        return cache_store[key]
```

app/main.py
```diff
@@ -1,6 +1,20 @@
...
-    # Write your code here
-    pass
+    cache_store = {}
+
+    @wraps(func)
+    def inner(*args, **kwargs) -> tuple:
+        key = (args, tuple(kwargs.items()))
+        if key in cache_store:
+            print("Getting from cache")
+            return cache_store[key]
```
vasylhnatiuk: and delete this line

app/main.py
```diff
@@ -1,6 +1,20 @@
...
+        else:
+            print("Calculating new result")
+            result = func(*args, **kwargs)
+            cache_store[key] = result
```
```suggestion
            cache_store[key] = func(*args, **kwargs)
```

vasylhnatiuk: you can avoid redundant varible

vasylhnatiuk (changes requested):

vasylhnatiuk (approved):

PR #1122, opened 2024-05-12
Develop

AndrewwSSS:
Solution

### DIFF
--- a/app/main.py
+++ b/app/main.py
@@ -1,6 +1,15 @@
 from typing import Callable
-
+_cache = {}
 
 def cache(func: Callable) -> Callable:
-    # Write your code here
-    pass
+    def wrapper(*args):
+        if func not in _cache:
+            _cache[func] = {}
+        if args not in _cache[func]:
+            print("Calculating new result")
+            _cache[func][args] = func(*args)
+        else:
+            print("Getting from cache")
+
+        return _cache[func][args]
+    return wrapper
--- a/app/main.py
+++ b/app/main.py
@@ -11,6 +11,5 @@ def cache(func: Callable) -> Callable:
             print("Calculating new result")
         else:
             print("Getting from cache")
-    
         return _CACHE[func][args]
     return wrapper
--- a/app/main.py
+++ b/app/main.py
@@ -1,10 +1,10 @@
-from typing import Callable
+from typing import Callable, Any
 
 
 def cache(func: Callable) -> Callable:
     cache_storage = {}
 
-    def wrapper(*args: int) -> int:
+    def wrapper(*args: Any) -> Any:
         if args not in cache_storage:
             cache_storage[args] = func(*args)
             print("Calculating new result")
--- a/app/main.py
+++ b/app/main.py
@@ -3,7 +3,7 @@ _cache = {}
 
 
 def cache(func: Callable) -> Callable:
-    def wrapper(*args):
+    def wrapper(*args) -> int:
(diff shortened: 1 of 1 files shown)


app/main.py
```diff
@@ -1,6 +1,16 @@
 from typing import Callable
+_cache = {}
```
DevSpaciX: If you use global variables, you should call them using the capitalize mode . For example: CACHE_STORAGE = {}

app/main.py
```diff
@@ -1,6 +1,16 @@
...
 
 
 def cache(func: Callable) -> Callable:
-    # Write your code here
-    pass
+    def wrapper(*args) -> int:
+        if func not in _cache:
+            _cache[func] = {}
+        if args not in _cache[func]:
```
DevSpaciX: 1. Just check if args not in your cache bucket. 
2. If that args not in your cache storage , get the func(*args) as the result and then add as a key for your sorage this way : CACHE_STORAGE[args] = result

app/main.py
```diff
@@ -1,6 +1,16 @@
...
+            print("Calculating new result")
+            _cache[func][args] = func(*args)
```
DevSpaciX: It is considered good practice to write prints at the end of the logical chain

DevSpaciX (changes requested):

app/main.py
```diff
@@ -1,6 +1,16 @@
 from typing import Callable
+_cache = {}
 
 
 def cache(func: Callable) -> Callable:
-    # Write your code here
-    pass
+    def wrapper(*args) -> int:
```
DevSpaciX: Are you sure you are returning int here ? And add pls also type hint for args :)

app/main.py
AndrewwSSS: I don't understand how my decorator can work with different functions in this way. I have a global variable where i store common cache for some functions

app/main.py
AndrewwSSS: And yes, we have 2 functions to decorate, and both functions return int

app/main.py
```diff
@@ -1,6 +1,15 @@
 from typing import Callable
+_CACHE = {}
 
 
 def cache(func: Callable) -> Callable:
-    # Write your code here
-    pass
+    def wrapper(*args: int) -> int:
```
Oleg-s-s: _And yes, we have 2 functions to decorate, and both functions return int_
We are writing not a decorator for tests, when you write a function you should assume that you work with some sort of data, in our case what type of data can be cached?

app/main.py
```diff
@@ -1,6 +1,15 @@
...
+        if func not in _CACHE:
+            _CACHE[func] = {}
+        if args not in _CACHE[func]:
+            _CACHE[func][args] = func(*args)
+            print("Calculating new result")
+        else:
+            print("Getting from cache")
+        return _CACHE[func][args]
```
Oleg-s-s: There is no need to store info about function in cache, you are working harder, and we need to think. Why not storing cache inside outer function? Then each function that will be decorated will have its own cache, and you can only store args of that function to check if we already worked with that data.
Also, this approach will help you to reduce and simplify your code

Oleg-s-s (changes requested):

app/main.py
AndrewwSSS: I don't understand what you want to tell me

DevSpaciX (approved):
Good job!

PR #1123, opened 2024-05-12
Solution

MaksymProtsak:


app/main.py
```diff
@@ -1,6 +1,16 @@
...
 
 
 def cache(func: Callable) -> Callable:
-    # Write your code here
-    pass
+    results = {}
+
+    @functools.wraps(func)
+    def wrapper(*args, **kwargs) -> Callable:
+        if args not in results:
```
vsmutok: you should check not only `args` but also `kwargs` together

vsmutok (changes requested):

app/main.py
MaksymProtsak: if I do `if args not in results and kwargs not in result` I have an error.

MaksymProtsak: Done.

app/main.py
```diff
@@ -1,6 +1,19 @@
...
+            results[args] = func(*args, **kwargs)
+            print("Calculating new result")
+        elif kwargs and kwargs not in results:
+            results[kwargs] = func(*args, **kwargs)
+            print("Calculating new result")
+        else:
+            print("Getting from cache")
+        return results[args]
+    return wrapper
```
```suggestion
        key = args + ... kwargs 
        if key not in results:
            results[key] = func(*args, **kwargs)
            print("Calculating new result")
        else:
            print("Getting from cache")
        return results[key]
    return wrapper
```

vsmutok (changes requested):

app/main.py
MaksymProtsak: I can concatenate tuple and tuple, not tuple and dict.
Do not anderstand what is `...` in  `key = args + ... kwargs`.

vsmutok: @MaksymProtsak  you have to think about what to use instead of `...` 
to combine `args ` and `kwargs ` into one key

app/main.py
```diff
@@ -1,6 +1,17 @@
...
 
 
 def cache(func: Callable) -> Callable:
-    # Write your code here
-    pass
+    results = {}
+
+    @functools.wraps(func)
+    def wrapper(*args, **kwargs) -> Callable:
+        key = args
```
vsmutok: combine `args ` and `kwargs ` into one key

vsmutok (changes requested):

app/main.py
MaksymProtsak: key = args + tuple(kwargs.keys())

app/main.py
```diff
@@ -1,6 +1,17 @@
...
 
 
 def cache(func: Callable) -> Callable:
-    # Write your code here
-    pass
+    results = {}
+
+    @functools.wraps(func)
+    def wrapper(*args, **kwargs) -> Callable:
+        key = args + tuple(kwargs.keys())
```
```suggestion
        key = args + tuple(kwargs.items())
```

vsmutok (changes requested):

app/main.py
MaksymProtsak: Done.

vsmutok (approved):
Good Job 👍

PR #1125, opened 2024-05-14
implement py-cache-decorator

nikita-plotnytskyi:


app/main.py
```diff
@@ -1,6 +1,15 @@
-from typing import Callable
+from typing import Callable, Any
 
 
 def cache(func: Callable) -> Callable:
-    # Write your code here
-    pass
+    result_cache = {}
+
```

app/main.py
```diff
@@ -1,6 +1,15 @@
...
+    def inner(*args, **kwargs) -> Any:
+        if args in result_cache:
```
you must use `args` + `kwargs` as the keys of `result_cache `

vsmutok (changes requested):

ArsenPidhoretskyi (approved):
Well done!

PR #1128, opened 2024-05-16
Solution

SuskyiVolodymyr:


app/main.py
```diff
@@ -2,5 +2,15 @@
 
 
 def cache(func: Callable) -> Callable:
-    # Write your code here
-    pass
+    cache_results = dict()
```
DevSpaciX: Why not using just -> cache_results = {}. instead of calling dict() function ?

app/main.py
```diff
@@ -2,5 +2,15 @@
...
+
+    def inner(*args) -> int:
```
DevSpaciX: You can receive here any type of data ) Add pls type hint on 'args' , and change the 'int' type hint

DevSpaciX (changes requested):
Almost! More attention to details pls :)

app/main.py
SuskyiVolodymyr: i thought that set would be created insted of dict

app/main.py
```diff
@@ -1,6 +1,16 @@
...
 def cache(func: Callable) -> Callable:
-    # Write your code here
-    pass
+    cache_results = {}
+
+    def inner(*args: Any) -> Any:
+        if args not in cache_results:
+            print("Calculating new result")
+            result = func(*args)
+            cache_results[args] = result
```
DevSpaciX: It can be ->
`cache_results[args] = func(*args) `
is not it ? : )

app/main.py
```diff
@@ -1,6 +1,16 @@
...
+        else:
+            print("Getting from cache")
+            result = cache_results[args]
+        return result
```
DevSpaciX: You can return here just cache_results[args] , instead of writing this value to a variable 'result' Besides which you already have on 10 line , try always use different variables naming

DevSpaciX (changes requested):
Almost ! Try to use less various variables that can be avoided

DevSpaciX (approved):
Good job !

PR #1129, opened 2024-05-17
add solution

jpdreamthug:
image

### DIFF
--- a/app/main.py
+++ b/app/main.py
@@ -1,10 +1,10 @@
-from typing import Callable
+from typing import Callable, Any
 
 
 def cache(func: Callable) -> Callable:
     cache_dict = {}
 
-    def wrapper(*args, **kwargs) -> int:
+    def wrapper(*args: Any, **kwargs: Any) -> Any:
         key = (args, frozenset(kwargs.items()))
         if key not in cache_dict:
             print("Calculating new result")
--- a/app/main.py
+++ b/app/main.py
@@ -5,12 +5,12 @@ def cache(func: Callable) -> Callable:
     cache_dict = {}
 
     def wrapper(*args, **kwargs) -> int:
-        key = (func.__name__, args, frozenset(kwargs.items()))
-        if key in cache_dict:
+        key = (args, frozenset(kwargs.items()))
+        if key not in cache_dict:
+            print("Calculating new result")
+            cache_dict[key] = func(*args, **kwargs)
+        else:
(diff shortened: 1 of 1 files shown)


app/main.py
```diff
@@ -2,5 +2,15 @@
 
 
 def cache(func: Callable) -> Callable:
-    # Write your code here
-    pass
+    cache_dict = {}
+
+    def wrapper(*args, **kwargs) -> int:
+        key = (func.__name__, args, frozenset(kwargs.items()))
+        if key in cache_dict:
```
DevSpaciX: Try always to fetch negative case first ( `if smth not in cache_dict:` )

app/main.py
```diff
@@ -2,5 +2,15 @@
...
+            print("Getting from cache")
+            return cache_dict[key]
+        print("Calculating new result")
+        result = func(*args, **kwargs)
+        cache_dict[key] = result
```
DevSpaciX: If negative case will be first , try to call `return` only one time

app/main.py
```diff
@@ -2,5 +2,15 @@
 
 
 def cache(func: Callable) -> Callable:
-    # Write your code here
-    pass
+    cache_dict = {}
+
+    def wrapper(*args, **kwargs) -> int:
+        key = (func.__name__, args, frozenset(kwargs.items()))
```
DevSpaciX: In this case, it is not necessary to store the function name as a key; you can cache only the args. ( `cache_dict[args]` )

DevSpaciX (changes requested):
All good , a little optimisation required : )

app/main.py
jpdreamthug: I know I don't need a func __name__, but I wanted to improve it a little)

DevSpaciX: If you want ) Okay,  let it be here : )

app/main.py
```diff
@@ -2,5 +2,15 @@
 
 
 def cache(func: Callable) -> Callable:
-    # Write your code here
-    pass
+    cache_dict = {}
+
+    def wrapper(*args, **kwargs) -> int:
```
DevSpaciX: Also do not forget to add type hints on `*args and **kwargs` , also you do not every time returning int , it can be any type of data )
As a task says `You have a big array with data`

DevSpaciX (approved):
Well done !

PR #1131, opened 2024-05-17
Solution

lilarin:


app/main.py
```diff
@@ -2,5 +2,14 @@
 
 
 def cache(func: Callable) -> Callable:
-    # Write your code here
-    pass
+    results_cache = {}
+
+    def inner(*args) -> dict:
```
DevSpaciX: Add here type hint for args please :) And you not  returning dict , it can be any type ( int , float etc. )

DevSpaciX (changes requested):
Add / change type hints pls :)

app/main.py
lilarin: > Add here type hint for args please :) And you not returning dict , it can be any type ( int , float etc. )

lilarin: Yes, of course, but I was just testing under tests and wanted to get rid of "Any" to lighten the code while keeping functionality

DevSpaciX (approved):
Good job!

PR #1132, opened 2024-05-17
Solution

VShveda:


app/main.py
```diff
@@ -1,6 +1,17 @@
...
+from typing import Callable, Any
 
 
 def cache(func: Callable) -> Callable:
-    # Write your code here
-    pass
+    cache_ = {}
+
+    def inner(*args) -> Any:
+        key = args
```
DevSpaciX: Why is this variable needed? Just use `args` instead of creating varible with the ssame value

app/main.py
```diff
@@ -1,6 +1,17 @@
...
+        if key in cache_:
```
DevSpaciX: ( relates to previous comment )
You can do it this way -> `if args in cache_:`

app/main.py
```diff
@@ -1,6 +1,17 @@
...
+            print("Getting from cache")
+            return cache_[key]
+        else:
+            print("Calculating new result")
+            result = func(*args)
+            cache_[key] = result
+            return result
```
DevSpaciX: Think how you can call `return` only once . If `cache_[key] = result` you can `return cached[key]` and it is makes 2 identical returns in your func . Optimise it : )

app/main.py
```diff
@@ -1,6 +1,17 @@
-from typing import Callable
+from typing import Callable, Any
 
 
 def cache(func: Callable) -> Callable:
-    # Write your code here
-    pass
+    cache_ = {}
+
+    def inner(*args) -> Any:
```
DevSpaciX: Also please add type hint for `args` :)

DevSpaciX (changes requested):

wQuelS (changes requested):
I don't see additional commits that would fix the problems mentioned by a mentor. Please read all the comments that were left by @DevSpaciX and make improvements

app/main.py
```diff
@@ -4,14 +4,13 @@
 def cache(func: Callable) -> Callable:
     cache_ = {}
 
-    def inner(*args) -> Any:
-        key = args
-        if key in cache_:
+    def inner(*args: tuple) -> Any:
```
wQuelS: please Google how to annotate `*args` and why. Even `Any` would be better here

app/main.py
```diff
@@ -1,6 +1,16 @@
...
+
+    def inner(*args: tuple) -> Any:
+        if args in cache_:
+            print("Getting from cache")
+            result = cache_[args]
+        else:
+            print("Calculating new result")
+            result = func(*args)
+            cache_[args] = result
+        return result
```
wQuelS: this can be simplified by avoiding line 10 and making lines 13 and 14 in one action without a new variable. 
You will simply need to return `cache[key]` in your  return statement instead

This is not a critical mistake but a suggestion

wQuelS (changes requested):

Oleg-s-s (approved):
:+1:

PR #1133, opened 2024-05-18
18/05 cache-decorator

NikitaNikolaitsev:


app/main.py
```diff
@@ -1,6 +1,16 @@
...
+
+    def inner(*args: Any) -> Any:
+        if args in cache_dict:
+            print("Getting from cache")
+            return cache_dict[args]
+        else:
+            print("Calculating new result")
+            result = func(*args)
+            cache_dict[args] = result
+            return result
```
DevSpaciX: Think how you can call return only once . If `cache_dict[args] = result ` you can return  cache_dict[args] and it is makes 2 identical returns in your func . Optimise it : )

DevSpaciX (changes requested):

wQuelS (changes requested):
all tests should pass, including flake8 routine

wQuelS (changes requested):
reject, since not all GH actions are green

wQuelS (approved):
Looks good to me

PR #1134, opened 2024-05-18
Test solution

VasylRysan:


app/main.py
```diff
@@ -1,6 +1,15 @@
-from typing import Callable
+from typing import Callable, Any
 
 
 def cache(func: Callable) -> Callable:
-    # Write your code here
-    pass
+    storage = {}
+
```

app/main.py
```diff
@@ -1,6 +1,15 @@
...
+    def wrapper(*args, **kwargs) -> Any:
+        if args in storage:
+            print("Getting from cache")
+            return storage.get(args)
```
you must use `args` + `kwargs` as the keys of `storage `

vsmutok (changes requested):

vsmutok (changes requested):
.

vasylhnatiuk (approved):

PR #1135, opened 2024-05-18
solution

haldaniko:


app/main.py
```diff
@@ -1,6 +1,17 @@
...
 
 
 def cache(func: Callable) -> Callable:
-    # Write your code here
-    pass
+    func_cache: Dict[Tuple, any] = {}
+
+    def wrapper(*args) -> Callable:
+        if args in func_cache:
+            print("Getting from cache")
```
you must use `args` + `kwargs` as the keys of `func_cache`

app/main.py
```diff
@@ -1,6 +1,17 @@
-from typing import Callable
+from typing import Callable, Dict, Tuple
 
 
 def cache(func: Callable) -> Callable:
-    # Write your code here
-    pass
+    func_cache: Dict[Tuple, any] = {}
+
+    def wrapper(*args) -> Callable:
```

vsmutok (changes requested):

app/main.py
```diff
@@ -1,6 +1,20 @@
...
+    def wrapper(*args, **kwargs) -> Callable:
+        key = (args, frozenset(kwargs.items()))
+        if key in func_cache:
+            print("Getting from cache")
+            return func_cache[key]
+        else:
+            print("Calculating new result")
+            result = func(*args, **kwargs)
+            func_cache[key] = result
+            return result
```
```suggestion
        if key in func_cache:
            print("Getting from cache")
        else:
            print("Calculating new result")
            func_cache[key] = func(*args, **kwargs)
        return func_cache[key]
```
Your code can be simplified a bit))

vsmutok (changes requested):

vsmutok (approved):
Well Done 🚀

PR #1138, opened 2024-05-19
Solution_V1

BileichukIvan:


app/main.py
```diff
@@ -1,6 +1,18 @@
...
+from typing import Callable, Any
 
 
 def cache(func: Callable) -> Callable:
-    # Write your code here
-    pass
+    cache_storage = {}
+
+    def inner(*args, **kwargs) -> Any:
+        key = (args, tuple(sorted(kwargs.items())))
```
DevSpaciX: You just can check if args in cache_storage , you don`t need to create another sorted tuple

app/main.py
```diff
@@ -1,6 +1,18 @@
...
+
+        if key in cache_storage:
+            print("Getting from cache")
+            return cache_storage[key]
+        else:
+            print("Calculating new result")
+            result = func(*args, **kwargs)
+            cache_storage[key] = result
+            return result
```
DevSpaciX: Try to call `return` only once . If `result ` thats a key of your cache storage , you can simply return cache_storage[key] outside the `if ... else` block is not it ?

app/main.py
```diff
@@ -1,6 +1,18 @@
-from typing import Callable
+from typing import Callable, Any
 
 
 def cache(func: Callable) -> Callable:
-    # Write your code here
-    pass
+    cache_storage = {}
+
+    def inner(*args, **kwargs) -> Any:
```
DevSpaciX: Add pls type hint for `args` , and also you do not need here kwargs ( explained below )

DevSpaciX (changes requested):

app/main.py
```diff
@@ -1,6 +1,19 @@
...
+        key = args
+
+        if key in cache_storage:
+            print("Getting from cache")
+            return cache_storage[key]
+
+        print("Calculating new result")
+        result = func(*args)
+        cache_storage[key] = result
+        return result
```
wQuelS: still two `return` statements.

Try to write the logic for each condition first. For example: if there is no such key in your cache -> `print()` needed phrase and apply that `func` on `*args`, otherwise make a print of another phrase, according to the task description.
Since both conditions require only `print()` statement your `return` can be written outside `if .. esle` block.

app/main.py
```diff
@@ -1,6 +1,19 @@
...
+from typing import Callable, Any
 
 
 def cache(func: Callable) -> Callable:
-    # Write your code here
-    pass
+    cache_storage = {}
+
+    def inner(*args: Any) -> Any:
+        key = args
```
wQuelS: Do you really need this additional variable?

wQuelS (changes requested):

wQuelS (approved):
Nice

PR #1139, opened 2024-05-19
Solution

SrMusienko:


app/main.py
```diff
@@ -2,5 +2,17 @@
 
 
 def cache(func: Callable) -> Callable:
-    # Write your code here
-    pass
+    my_cache = {}
+
+    def wrapper(*args, **kwargs) -> Callable:
+        dict_key = (*args, tuple(kwargs.items()))
```
DevSpaciX: You just can check if args already in  `my_cache` , you dont need to create `dict_key` variable

app/main.py
```diff
@@ -2,5 +2,17 @@
 
 
 def cache(func: Callable) -> Callable:
-    # Write your code here
-    pass
+    my_cache = {}
+
+    def wrapper(*args, **kwargs) -> Callable:
```
DevSpaciX: Add pls type hint for `args` , and also you do not need here kwargs ( explained below ) . Also please think what type of data you returning , it is not `Callable`

app/main.py
```diff
@@ -2,5 +2,17 @@
...
+        dict_key = (*args, tuple(kwargs.items()))
+
+        if dict_key in my_cache:
+            print("Getting from cache")
+            return my_cache[dict_key]
```
DevSpaciX: You have 2 identical returns , try to call `return` only once outside the `if ... else block`

DevSpaciX (changes requested):
A little optimisation requested :) But overall good !

app/main.py
SrMusienko: I decided that this way the code would be more readable

SrMusienko: And the code will be even shorter, thanks to the fact that instead of a long operator we use a short variable several times

app/main.py
```diff
@@ -2,5 +2,16 @@
 
 
 def cache(func: Callable) -> Callable:
-    # Write your code here
-    pass
+    my_cache = {}
+
+    def wrapper(*args: list, **kwargs: dict) -> None:
```
wQuelS: are you sure about the `return` type annotation here? Take a look at line 15. What exactly this inner function can return?

wQuelS (changes requested):

app/main.py
SrMusienko: To make the code more readable, replace "*args, **kwargs" with "var: Any"

Oleg-s-s (approved):
:+1:

PR #1140, opened 2024-05-19
cache-decorator solution

DimaYankovskiy:


app/main.py
```diff
@@ -2,5 +2,15 @@
 
 
 def cache(func: Callable) -> Callable:
-    # Write your code here
-    pass
+    cached = {}
+
+    def inner(*args, **kwargs) -> Callable:
+        key = f"{func.__name__} {str(args)} {str(kwargs)}"
```
DevSpaciX: You just can check if args already in your cache storage , you dont need to create key variable

app/main.py
```diff
@@ -2,5 +2,15 @@
 
 
 def cache(func: Callable) -> Callable:
-    # Write your code here
-    pass
+    cached = {}
```
DevSpaciX: Try to use more descriptive naming . It can be cache_storage , cache , etc

app/main.py
```diff
@@ -2,5 +2,15 @@
...
+
+    def inner(*args, **kwargs) -> Callable:
+        key = f"{func.__name__} {str(args)} {str(kwargs)}"
+        if cached.get(key) is None:
```
DevSpaciX: You can check it more simply this way ->` if ... in your_cache:`

app/main.py
```diff
@@ -2,5 +2,15 @@
...
+            print("Calculating new result")
+            result = func(*args, **kwargs)
+            cached[key] = result
+            return result
```
DevSpaciX: Think how you can call `return` only once . If  cached[key] = `result` you can return `cached[key]` and it is makes 2 identical `returns` in your func . Optimise it : )

app/main.py
```diff
@@ -2,5 +2,15 @@
 
 
 def cache(func: Callable) -> Callable:
-    # Write your code here
-    pass
+    cached = {}
+
+    def inner(*args, **kwargs) -> Callable:
```
DevSpaciX: Add pls type hint for `args` , and also you do not need here `kwargs` ( explained below ) . Also please think what type of data you returning , it is not Callable

DevSpaciX (changes requested):

app/main.py
```diff
@@ -1,6 +1,15 @@
-from typing import Callable
+from typing import Callable, Any
 
 
 def cache(func: Callable) -> Callable:
-    # Write your code here
-    pass
+    cache_storage = {}
+
+    def inner(*args) -> Any:
```
wQuelS, line 7: still not fixed `*args`
please, re-read the first sentence in the below comment:
https://github.com/mate-academy/py-cache-decorator/pull/1140#discussion_r1606311641

app/main.py
```diff
@@ -1,6 +1,15 @@
...
+        if args in cache_storage:
+            print("Getting from cache")
+            return cache_storage[args]
+        result = func(*args)
+        cache_storage[args] = result
+        print("Calculating new result")
+        return result
```
wQuelS, line 14: And again: https://github.com/mate-academy/py-cache-decorator/pull/1140#discussion_r1606310835

Please, read all the comments carefully and add fixes where needed. You've ignored at least 2 comments already

wQuelS (changes requested):

PR #1141, opened 2024-05-19
Solution

Yuruch:


app/main.py
```diff
@@ -1,6 +1,15 @@
...
 
 def cache(func: Callable) -> Callable:
-    # Write your code here
-    pass
+    cache_dict = {}
+
+    def inner(*args: Any) -> Any:
+        if args in cache_dict:
+            print("Getting from cache")
+            return cache_dict[args]
```
DevSpaciX: You have 2 identical returns , try to use only 1 at the end

DevSpaciX (changes requested):
All good , a little optimisation requested

DevSpaciX (approved):
Good job!

PR #1142, opened 2024-05-20
py-cache-decorator v1

BornToLivee:


app/main.py
```diff
@@ -2,5 +2,15 @@
 
 
 def cache(func: Callable) -> Callable:
-    # Write your code here
-    pass
+    stored_dict = {}
+
+    def wrapper(*args, **kwargs) -> int | list:
```
wQuelS: add correct type annotations for your `args` and `kwargs`. Return type annotation can be also improved

app/main.py
```diff
@@ -2,5 +2,15 @@
...
+        if args in stored_dict:
+            print("Getting from cache")
+            return stored_dict[args]
+        else:
+            print("Calculating new result")
+            result = func(*args, **kwargs)
+            stored_dict[args] = result
+            return result
```
wQuelS: Take a look: you have two return statements. You can simplify it because you need only one.

You may perform `printing` logic depending on which condition is `True` and return as a last action outside your `if .. else` block

wQuelS (changes requested):

wQuelS (changes requested):
Please, fix issues stated in the comments above

app/main.py
```diff
@@ -2,5 +2,15 @@
 
 
 def cache(func: Callable) -> Callable:
-    # Write your code here
-    pass
+    stored_dict = {}
+
+    def wrapper(*args: int | tuple, **kwargs: int | tuple) -> int | list[int]:
```
wQuelS: image
According to tests `int` is not the only possible type that can be sent. You'd better add some type hint that can be suited for different `types`

app/main.py
```diff
@@ -2,5 +2,15 @@
...
+        if args in stored_dict:
+            print("Getting from cache")
+            result = stored_dict[args]
+        else:
+            print("Calculating new result")
+            result = func(*args, **kwargs)
+            stored_dict[args] = result
+        return result
```
wQuelS: that's okay, you have 1 return statement. But if you want you may also get rid of an additional variable. It is up to you

wQuelS (changes requested):

wQuelS (approved):
Good

PR #1143, opened 2024-05-20
Solution

AntonKorinchuk:


app/main.py
```diff
@@ -2,5 +2,13 @@
 
 
 def cache(func: Callable) -> Callable:
-    # Write your code here
-    pass
+    cached_dict = {}
+
+    def wrapper(*args, **kwargs) -> None:
```
wQuelS: Add type annotations for your `*args` and `**kwargs`.
Also, are you sure about current return type annotation?

wQuelS (changes requested):

wQuelS (approved):
LGTM

PR #1145, opened 2024-05-20
Solution

vladyslavchichkan:


app/main.py
```diff
@@ -2,5 +2,16 @@
...
+    def wrapper(*args) -> None:
+
+        if args in cache_dict:
+            print("Getting from cache")
+            return cache_dict[args]
+        else:
+            print("Calculating new result")
+            result = func(*args)
+            cache_dict[args] = result
+            return result
```
DevSpaciX: Think how you can call `return` only once . If cache_dict[args] = result you can return `cache_dict[args]` and it is makes 2 identical returns in your func . Optimise it : )

app/main.py
```diff
@@ -2,5 +2,16 @@
 
 
 def cache(func: Callable) -> Callable:
-    # Write your code here
-    pass
+    cache_dict = {}
+
+    def wrapper(*args) -> None:
```
DevSpaciX: Add type hint for args , and change the None type ( You do not returning None )

DevSpaciX (changes requested):

app/main.py
```diff
@@ -1,6 +1,18 @@
-from typing import Callable
+from typing import Callable, Any
 
 
 def cache(func: Callable) -> Callable:
-    # Write your code here
-    pass
+    cache_dict = {}
+
+    def wrapper(*args) -> Any:
```
wQuelS: `*args` still miss type hint

app/main.py
```diff
@@ -1,6 +1,18 @@
...
+
+        if args in cache_dict:
+            print("Getting from cache")
+            result = cache_dict[args]
+        else:
+            print("Calculating new result")
+            result = func(*args)
+            cache_dict[args] = result
+        return result
```
wQuelS: this also can be done without an additional `result` variable. This is not a crucial mistake and is totally **_up to you_**

wQuelS (changes requested):

app/main.py
```diff
@@ -2,5 +2,17 @@
 
 
 def cache(func: Callable) -> Callable:
-    # Write your code here
-    pass
+    cache_dict = {}
+
+    def wrapper(*args) -> Callable:
+
```
wQuelS, line 8: a `Callable` object is, for example, a function. And that's not what your `wrapper` returns.
The previous `return type annotation` was correct, you had only to add a type hint for `*args`

wQuelS (changes requested):

DevSpaciX (approved):
What okay ) But type hint for args is Any , and wrapper func returning Any type too : )

PR #1146, opened 2024-05-20
Solution

b3v3kt0r:


app/main.py
```diff
@@ -2,5 +2,13 @@
 
 
 def cache(func: Callable) -> Callable:
-    # Write your code here
-    pass
+    cash = {}
+
+    def inner(*args) -> Callable:
```
wQuelS: add type hint for `*args` .
also, are u sure about the return type annotation here?

app/main.py
```diff
@@ -2,5 +2,13 @@
 
 
 def cache(func: Callable) -> Callable:
-    # Write your code here
-    pass
+    cash = {}
```
wQuelS: "cash" is a pretty bad variable name here. Its meaning is completely different. Try to give good, meaningful names to make your code more readable, clean and easier to maintain

wQuelS (changes requested):

app/main.py
```diff
@@ -2,5 +2,13 @@
...
+
+    def inner(*args: Callable) -> Callable:
```
wQuelS: `Callable` object is, for example, a function. It is definitely not what is expected to "come" into `*args` attr or to be returned from a function.

You can use some specific type from the `typing` module to specify that any type is acceptable. As well as any type can be returned.

wQuelS (changes requested):
Please, read all the comments carefully and don't miss suggestions. For example, you still did not fix `cash` variable name which is not good.

app/main.py
```diff
@@ -1,14 +1,14 @@
-from typing import Callable
+from typing import Callable, Any
 
 
-def cache(func: Callable) -> Callable:
-    cash = {}
+def cache(func: Callable) -> Any:
```
wQuelS: why did you add `Any` here? 
What is expected to be returned by this "outer" function?

wQuelS (changes requested):

app/main.py
b3v3kt0r: sorry, had a little trouble with that stuff. 
for outer must stay Callable? because we expect to get a function from that

DevSpaciX (approved):
Great !

PR #1148, opened 2024-05-22
Solution

tsukimi1394:


app/main.py
```diff
@@ -2,5 +2,14 @@
 
 
 def cache(func: Callable) -> Callable:
-    # Write your code here
-    pass
+    saved_results = {}
+
+    def wrapper(*args, **kwargs) -> Callable:
+        if args in saved_results:
```
* what  if the function takes the same `args ` but different `kwargs` 
* you must use `args ` with `kwargs`  as the key of `saved_results`

vsmutok (changes requested):

app/main.py
tsukimi1394: 1. _use `functools.wraps`_ -> agree, makes sense
2. since kwargs are dict I firstly encountered `TypeError: unhashable type: 'dict'`, after some attempts and googling found `frozenset`

EdAlekseiev (approved):
Good job

PR #1152, opened 2024-05-24
Solution

voboian:


app/main.py
```diff
@@ -2,5 +2,17 @@
...
+        all_args = []
+        for arg in args:
+            all_args.append(arg)
+        if str(all_args) in all_cache:
+            print("Getting from cache")
+        else:
+            print("Calculating new result")
+            all_cache[str(all_args)] = func(*args)
+        return all_cache[str(all_args)]
+    return inner
```
vsmutok: no need to use loop here 

*  **Tuples as keys:**
   - Use tuples directly as dictionary keys instead of strings.

*  **Simplified caching logic:**
   - Check if arguments are in cache; if so, print "Getting from cache" and return cached result.
   - If not, compute result, store in cache, print "Calculating new result", and return result.

vsmutok (changes requested):

app/main.py
```diff
@@ -2,5 +2,14 @@
 
 
 def cache(func: Callable) -> Callable:
-    # Write your code here
-    pass
+
+    all_cache = {}
+
+    def inner(*args) -> int:
+        if args in all_cache:
```
* what if the function takes the same `args`  but different `kwargs`?
* you must use `args` with `kwargs` as the key of `all_cache`

vsmutok (changes requested):

app/main.py
```diff
@@ -2,5 +2,17 @@
 
 
 def cache(func: Callable) -> Callable:
-    # Write your code here
-    pass
+
+    all_cache = {}
+
+    def wrapper(*args, **kwargs) -> int:
+        if args in all_cache:
```
pavlopro, line 9: How about kwargs in cache?

app/main.py
```diff
@@ -2,5 +2,17 @@
...
+            print("Getting from cache")
+            return all_cache[args]
+        else:
```
pavlopro, line 12: `else` is redundant

pavlopro (changes requested):

PR #1153, opened 2024-05-24
Solution

odiakov:


app/main.py
```diff
@@ -2,5 +2,16 @@
 
 
 def cache(func: Callable) -> Callable:
-    # Write your code here
-    pass
+    func_cache = {}
+
+    def wrapper(*args) -> Callable:
```

app/main.py
```diff
@@ -2,5 +2,16 @@
...
+        if args in func_cache:
```
vsmutok, line 10: Why you check only `args`?
What if function takes `kwargs`?

vsmutok (changes requested):

app/main.py
odiakov: because function takes only **immutable** arguments 
 I think we don't need to check kwargs here

vasylhnatiuk (approved):

PR #1154, opened 2024-05-25
"Solution"

Fedot0v:


viktoria-rybenchuk (changes requested):
All tests should passed

app/main.py
```diff
@@ -1,6 +1,17 @@
...
 
 
 def cache(func: Callable) -> Callable:
-    # Write your code here
-    pass
+    result_dict = {}
+
+    def wrapper(*args, **kwargs) -> Any:
+        if args in result_dict:
+            print("Getting from cache")
```

Why you check only args?
What if function takes kwargs?
You must use args + kwargs as a key

vsmutok (changes requested):

app/main.py
```diff
@@ -1,6 +1,18 @@
 from typing import Callable
+from typing import Any
 
 
 def cache(func: Callable) -> Callable:
-    # Write your code here
-    pass
+    result_dict = {}
+
+    def wrapper(*args, **kwargs) -> Any:
```

app/main.py
```diff
@@ -1,6 +1,18 @@
...
+        key = args, frozenset(kwargs)
+        if key in result_dict:
+            print("Getting from cache")
+            return result_dict[key]
+        else:
+            print("Calculating new result")
+            result = func(*args)
+            result_dict[key] = result
+            return result
```
```suggestion
        if key in result_dict:
            print("Getting from cache")
        else:
            print("Calculating new result")
            result_dict[key] = func(*args)
        return result_dict[key]
```

vsmutok (changes requested):

app/main.py
```diff
@@ -1,6 +1,17 @@
...
+import functools
 
 
 def cache(func: Callable) -> Callable:
-    # Write your code here
-    pass
+    result_dict = {}
+
+    @functools.wraps(func)
+    def wrapper(*args, **kwargs) -> Callable:
```
vsmutok, line 9:
```suggestion
    def wrapper(*args, **kwargs) -> Any:
```

vsmutok (approved):
Great!

PR #1155, opened 2024-05-26
Solution

pikaOpika:


app/main.py
```diff
@@ -1,6 +1,16 @@
...
 
 def cache(func: Callable) -> Callable:
-    # Write your code here
-    pass
+    def inner(*args, **kwargs) -> int:
+        global cache_dict
+        if (func, args) in cache_dict:
+            print("Getting from cache")
+            return cache_dict[(func, args)]
+        else:
```

app/main.py
```diff
@@ -1,6 +1,16 @@
 from typing import Callable
 
+cache_dict = {}
+
 
 def cache(func: Callable) -> Callable:
-    # Write your code here
-    pass
+    def inner(*args, **kwargs) -> int:
```
Dimosphen1: Indicate that inner could return `Any` type in the return type annotation, not only `int`

Dimosphen1: You could use a decorator from `functools` to preserve the wrapped function name and docstring

app/main.py
```diff
@@ -1,6 +1,16 @@
 from typing import Callable
 
+cache_dict = {}
```
Dimosphen1: Create `cache_dict` as a class attribute and do not use `global` when calling it

Dimosphen1 (changes requested):

app/main.py
```diff
@@ -1,6 +1,16 @@
...
 
 
 def cache(func: Callable) -> Callable:
-    # Write your code here
-    pass
+    cache_dict = {}
+
+    @wraps(func)
+    def inner(*args, **kwargs) -> Any:
+        if (func, args) in cache_dict:
```
```suggestion
    def inner(*args, **kwargs) -> Any:
    key = func.__name__, args , tuple(kwargs.items())
        if key in cache_dict:
```

vsmutok (changes requested):

Dimosphen1 (approved):
LGTM, well done!

PR #1158, opened 2024-06-02
created cache decorator

Timur5050:


app/main.py
```diff
@@ -2,5 +2,16 @@
 
 
 def cache(func: Callable) -> Callable:
-    # Write your code here
-    pass
+    cached_data = {}
+
+    def wrapper(*args) -> any:
```

app/main.py
```diff
@@ -2,5 +2,16 @@
...
+        if args in cached_data:
```
vsmutok: Why you check only `args`?
What if function takes `kwargs`?

app/main.py
```diff
@@ -2,5 +2,16 @@
...
+            print("Getting from cache")
+            return cached_data[args]
+
+        res = func(*args)
```
vsmutok: `res` not clear name

vsmutok (changes requested):

app/main.py
Timur5050: but the dict is not hashed, and there will be a TypeError when I try to use the dict as a key, how can I save it?

app/main.py
```diff
@@ -1,6 +1,23 @@
...
 def cache(func: Callable) -> Callable:
-    # Write your code here
-    pass
+    cached_data = {}
+
+    @functools.wraps(func)
+    def wrapper(*args, **kwargs) -> any:
+        kwargs_value = tuple()
+        for key, value in kwargs.items():
+            kwargs_value += (key, value)
```
Dimosphen1, line 12: You could convert `kwargs` to string and store them in this way avoiding iteration

app/main.py
```diff
@@ -1,6 +1,23 @@
...
 from typing import Callable
 
 
 def cache(func: Callable) -> Callable:
-    # Write your code here
-    pass
+    cached_data = {}
+
+    @functools.wraps(func)
+    def wrapper(*args, **kwargs) -> any:
```
Dimosphen1, line 9: It is preferable to use `Any` from `typing` since it is not deprecated

Dimosphen1 (approved):
LGTM, several comments were left for consideration

PR #1160, opened 2024-06-05
Solution

Anatolik-ua:


app/main.py
```diff
@@ -2,5 +2,19 @@
...
+        if key in cache_store:
+            print("Getting from cache")
+            return cache_store[key]
+        else:
+            print("Calculating new result")
+            result = func(*args, **kwargs)
+            cache_store[key] = result
+            return result
+
+    wrapper.__name__ = func.__name__
```
vasylhnatiuk: use  @wraps(func)

app/main.py
```diff
@@ -2,5 +2,19 @@
...
+    def wrapper(*args, **kwargs) -> Callable:
+        key = args + tuple(sorted(kwargs.items()))
+        if key in cache_store:
+            print("Getting from cache")
+            return cache_store[key]
+        else:
+            print("Calculating new result")
+            result = func(*args, **kwargs)
+            cache_store[key] = result
+            return result
```
vasylhnatiuk: use on;y one return in wrapper()

vasylhnatiuk (changes requested):

app/main.py
Anatolik-ua: done

app/main.py
Anatolik-ua: with functools 
(if I understood correctly)

app/main.py
```diff
@@ -1,6 +1,20 @@
...
+    def wrapper(*args, **kwargs) -> Callable:
+        key = args + tuple(sorted(kwargs.items()))
+        if key in cache_store:
+            print("Getting from cache")
+            result = cache_store[key]
+        else:
+            print("Calculating new result")
+            result = func(*args, **kwargs)
+            cache_store[key] = result
+        return result
```
```suggestion
    @functools.wraps(func)
    def wrapper(*args, **kwargs) -> Callable:
        key = args + tuple(sorted(kwargs.items()))
        if key in cache_store:
            print("Getting from cache")
        else:
            print("Calculating new result")
            cache_store[key] = func(*args, **kwargs)
        return cache_store[key]
```

vsmutok (changes requested):

vasylhnatiuk (approved):

PR #1161, opened 2024-06-05
Solution

bknyrik:


app/main.py
```diff
@@ -1,6 +1,23 @@
-from typing import Callable
+from typing import Callable, Any
 
 
 def cache(func: Callable) -> Callable:
-    # Write your code here
```
pavlopro, line 5: Please simplify this method

app/main.py
```diff
@@ -1,6 +1,23 @@
...
-    pass
+    cache_of_results = []
+    cache_of_args = []
+
+    def wrapper(*args, **kwargs) -> Any:
+        nonlocal cache_of_results
+        nonlocal cache_of_args
+
+        total_args = args + tuple(kwargs.values())
```
pavlopro: include keys as well
`tuple(kwargs.items())`

app/main.py
```diff
@@ -1,6 +1,23 @@
-from typing import Callable
+from typing import Callable, Any
 
 
 def cache(func: Callable) -> Callable:
-    # Write your code here
-    pass
+    cache_of_results = []
```
pavlopro: Create only one variable `cache ` for exemple, which will be dict type

app/main.py
pavlopro: Name it properly
e.g  `key`

app/main.py
```diff
@@ -1,6 +1,23 @@
...
+    cache_of_args = []
+
+    def wrapper(*args, **kwargs) -> Any:
+        nonlocal cache_of_results
+        nonlocal cache_of_args
+
+        total_args = args + tuple(kwargs.values())
+        if total_args in cache_of_args:
+            print("Getting from cache")
+            return cache_of_results[cache_of_args.index(total_args)]
```
pavlopro: in case if `key` in `cache`, return value by the key

app/main.py
```diff
@@ -1,6 +1,23 @@
...
+        else:
+            print("Calculating new result")
+            cache_of_results.append(func(*args, **kwargs))
```
pavlopro: `cache['key'] = func(*args, **kwargs)`

pavlopro (changes requested):
Please simplify method
Make code refactroing

app/main.py
bknyrik: Why decorator need to record keys?

app/main.py
bknyrik: Is this a good practise to give the same name to variable as function?

app/main.py
bknyrik: I already understood this, don't answer on it

Polyakiv-Andrey (approved):
Good job!

PR #1162, opened 2024-06-06
solved

Krasul1a:


app/main.py
```diff
@@ -2,5 +2,16 @@
 
 
 def cache(func: Callable) -> Callable:
-    # Write your code here
-    pass
+    cah = {}
```
vsmutok: `cah` not clear variable name

app/main.py
```diff
@@ -2,5 +2,16 @@
...
+
+    def warp(*args) -> int:
+        if args in cah:
+            print("Getting from cache")
+            return cah[args]
```
* what  if the function takes the same `args ` but different `kwargs` 
* you must use `args ` with `kwargs`  as the key of `cah `

vsmutok (changes requested):

EdAlekseiev (approved):

PR #1169, opened 2024-06-11
Solution

Wallian169:


app/main.py
```diff
@@ -1,6 +1,18 @@
...
-    # Write your code here
-    pass
+    unique_values = {}
+
+    @wraps(func)
+    def inner(*args) -> Callable:
+        if args in unique_values:
+            print("Getting from cache")
+            return unique_values[args]
+        else:
```
viktoria-rybenchuk, line 13: use just one return

viktoria-rybenchuk (changes requested):

Wallian169:
Hello, what does it mean? Should I use only one return statement in inner
function or in "else" block?

вт, 11 черв. 2024 р. о 19:27 Viktoriia Rybenchuk ***@***.***>
пише:

app/main.py
```diff
@@ -1,6 +1,16 @@
...
+from functools import wraps
 
 
 def cache(func: Callable) -> Callable:
-    # Write your code here
-    pass
+    unique_values = {}
+
+    @wraps(func)
+    def inner(*args) -> Callable:
```
Dimosphen1: `inner` can return not only `Callable`, it can return `Any`. Indicate this in the return type annotation

Dimosphen1 (changes requested):

app/main.py
Wallian169: Ok, got it.

app/main.py
```diff
@@ -1,6 +1,16 @@
...
+from functools import wraps
 
 
 def cache(func: Callable) -> Callable:
-    # Write your code here
-    pass
+    unique_values = {}
+
+    @wraps(func)
+    def inner(*args) -> Any:
```
pavlopro: Please enhance this to make it work with **kwargs as well.

pavlopro (changes requested):

Wallian169:
I don't know how to do this. Moreover, mentor Andrey Polyakov gave me
advice to delete kwargs then I asked for help in py_chat

ср, 12 черв. 2024 р. о 14:18 Pavlo Prokopets ***@***.***>
пише:

app/main.py
Wallian169: Hello, I have added an opportunity to work with *kwargs. BUT for what if README file says that "Decorator cache creating for decorating only functions that take **immutable arguments.**"

Wallian169: Hello, I've added an opportunity to work with kwargs

pavlopro (approved):
Good Job!

Wallian169:
Thank you so much for the great news!

ср, 12 черв. 2024 р. о 15:49 Pavlo Prokopets ***@***.***>
пише:

PR #1170, opened 2024-06-11
'Decorator'

ViktoriaGera1:


app/main.py
```diff
@@ -2,5 +2,16 @@
...
+
+    def wrapper(*args, **kwargs) -> Callable:
+        key = (args, frozenset(kwargs.items()))
+        if key in res:
+            print("Getting from cache")
+            return res[key]
+        else:
+            print("Calculating new result")
+            res[key] = func(*args, **kwargs)
+            return res[key]
```
EdAlekseiev: You can rewrite this function with using one return. It will make your code more readable
```suggestion
        if key in res:
            print("Getting from cache")
        else:
            print("Calculating new result")
            res[key] = func(*args, **kwargs)
        return res[key]
```

EdAlekseiev (changes requested):

EdAlekseiev (approved):

PR #1174, opened 2024-06-18
Solution

vpurshega:


app/main.py
```diff
@@ -2,5 +2,16 @@
...
+
+    def wrapper(*args) -> any:
+        if args in cache_db:
+            print("Getting from cache")
+            return cache_db[args]
+        else:
+            print("Calculating new result")
+            result = func(*args)
+            cache_db[args] = result
+            return result
```
EdAlekseiev: - what if the function takes the same args  but different kwargs
- you must use args  with kwargs as the key of cache_db

EdAlekseiev (changes requested):

app/main.py
vpurshega: Do I need to convert a dictionary to a tuple with values?

PR #1178, opened 2024-06-22
Solution

efirshey:
solution added

### DIFF
--- a/app/main.py
+++ b/app/main.py
@@ -1,18 +1,21 @@
 from typing import Callable, Any
+from functools import wraps
 
 
 def cache(func: Callable) -> Callable:
-
     cache_storage = {}
 
-    def inner(*args) -> Any:
-        if args in cache_storage:
+    @wraps(func)
+    def inner(*args, **kwargs) -> Any:
+        key = (args, tuple(sorted(kwargs.items())))
+
+        if key in cache_storage:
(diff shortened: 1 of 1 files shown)


app/main.py
```diff
@@ -1,6 +1,18 @@
...
+from typing import Callable, Any
 
 
 def cache(func: Callable) -> Callable:
-    # Write your code here
-    pass
+
+    cache_storage = {}
+
+    def inner(*args) -> Any:
```
NazarSheremet: - Good to add an annotation from `functools` module

```suggestion
    @wraps(func)
    def inner(*args, , **kwargs) -> Any:
```

app/main.py
```diff
@@ -1,6 +1,18 @@
...
+        if args in cache_storage:
+            print("Getting from cache")
+            return cache_storage[args]
+        else:
+            print("Calculating new result")
+            result = func(*args)
+            cache_storage[args] = result
+            return result
```
NazarSheremet: This else look useless 
```suggestion
        print("Calculating new result")
        result = func(*args)
        cache_storage[args] = result
        return result
```

NazarSheremet (changes requested):

NazarSheremet (approved):
Good job!

PR #1182, opened 2024-06-26
Solution for cache decorator

nmakivchuk:
Solution for cache decorator

### DIFF
--- a/app/main.py
+++ b/app/main.py
@@ -1,6 +1,19 @@
 from typing import Callable
+import functools
 
 
 def cache(func: Callable) -> Callable:
-    # Write your code here
-    pass
+    cache_storage = {}
+
+    @functools.wraps(func)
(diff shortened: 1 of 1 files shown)


app/main.py
```diff
@@ -1,6 +1,19 @@
...
 def cache(func: Callable) -> Callable:
-    # Write your code here
-    pass
+    cache_storage = {}
+
+    @functools.wraps(func)
+    def wrapper(*args) -> Callable:
+        if args in cache_storage:
+            print("Getting from cache")
+            return cache_storage[args]
```
NazarSheremet: it should work with `**kwargs` too. Create a unique key using args and kwargs

NazarSheremet (changes requested):
Need to fix suggestions

NazarSheremet (approved):
Good job!

PR #1185, opened 2024-06-28
Solution

Ramzes525:


app/main.py
```diff
@@ -1,6 +1,18 @@
...
 
 def cache(func: Callable) -> Callable:
-    # Write your code here
-    pass
+    results_cache = {}
+
+    def inner(*args: tuple) -> Any:
+        if args in results_cache:
+            print("Getting from cache")
+            return results_cache[args]
```
NazarSheremet: - Good to add an annotation from functools module
- Also, it should work with `**kwargs`. Create a unique key using `args` and `kwargs`

NazarSheremet (changes requested):
Need to fix suggestions

YuriiSmolii (approved):
LGTM

PR #1186, opened 2024-06-28
Store Cache

bythewaters:


app/main.py
```diff
@@ -2,5 +2,15 @@
...
 
 def cache(func: Callable) -> Callable:
-    # Write your code here
-    pass
+    cache_store = {}
+
+    def inner(*args, **kwargs) -> Callable:
+        if args in cache_store:
+            print("Getting from cache")
+            return cache_store[args]
```
vasylhnatiuk: this return is redundant, try to use only one return in inner()

app/main.py
```diff
@@ -2,5 +2,15 @@
...
+        else:
+            result = func(*args, **kwargs)
+            cache_store[args] = result
```
```suggestion
            cache_store[args] = func(*args, **kwargs)
```

vasylhnatiuk (changes requested):

vasylhnatiuk (approved):

PR #1188, opened 2024-06-29
Solution

andrewsiemakin:


app/main.py
```diff
@@ -1,6 +1,21 @@
...
+        key = args + tuple(sorted(kwargs.items()))
+
+        if key in cache_store:
+            print("Getting from cache")
+            return cache_store[key]
+        else:
+            print("Calculating new result")
+            result = func(*args, **kwargs)
+            cache_store[key] = result
+            return result
```
```suggestion
        key = (args, tuple(sorted(kwargs.items()), func.__name__,)

        if key in cache_store:
            print("Getting from cache")
        else:
            print("Calculating new result")
            cache_store[key] = func(*args, **kwargs)
        return cache_store[key]
```
`func.__name__`- is needed for that, because 2 different functions can have the same arguments

vsmutok (changes requested):

app/main.py
```diff
@@ -1,6 +1,21 @@
...
+    cache_store = {}
+
+    @functools.wraps(func)
+    def wrapper(*args, **kwargs) -> Callable:
+        key = (args, tuple(sorted(kwargs.items())), func.__name__)
+
+        if key in cache_store:
+            print("Getting from cache")
+            return cache_store[key]
+        else:
```
viktoria-rybenchuk, line 14: else is redundant

viktoria-rybenchuk (changes requested):

viktoria-rybenchuk (approved):
LGTM!

PR #1189, opened 2024-07-02
solution

poznianski:


app/main.py
```diff
@@ -2,5 +2,18 @@
 
 
 def cache(func: Callable) -> Callable:
-    # Write your code here
-    pass
+    cache_dict = {}
+
+    def wrapper(*args) -> Callable:
+        key = (func.__name__,) + args
+
```
NazarSheremet: - Good to add an annotation from `functools` module

NazarSheremet (changes requested):
Need to fix suggestions

NazarSheremet (approved):
Good job!

PR #1191, opened 2024-07-03
solution

dvitalijb:


app/main.py
```diff
@@ -2,5 +2,15 @@
...
 
 def cache(func: Callable) -> Callable:
-    # Write your code here
-    pass
+    results = {}
+
+    def wrap(*args) -> int:
+        if args in results:
+            print("Getting from cache")
+            return results[args]
```
NazarSheremet: - Good to add an annotation from `functools `module

NazarSheremet (changes requested):
Need to fix suggestions

app/main.py
dvitalijb: I hope I understood comments
Added fixes

YuriiSmolii (approved):
LGTM

PR #1193, opened 2024-07-10
solution

aaanyyy:


app/main.py
```diff
@@ -1,6 +1,17 @@
 from typing import Callable
 
+saved_runs = {}
+
```
NazarSheremet: Need to move `saved_runs` inside `cache` function

app/main.py
```diff
@@ -1,6 +1,17 @@
...
 
 def cache(func: Callable) -> Callable:
-    # Write your code here
-    pass
+    def arg_saver(*argv, **argc) -> callable:
```
```suggestion
    def arg_saver(*args, **kwargs) -> callable:
```
Use generally accepted names for non-keyword and keyword arguments

NazarSheremet (changes requested):
Need to fix suggestions

aaanyyy:
The variable has been moved inside the function, names of the arguments were corrected

app/main.py
aaanyyy: done

app/main.py
aaanyyy: done

YuriiSmolii (approved):
LGTM

PR #1198, opened 2024-07-12
Solution

OleksiukStepan:
Implemented method **cache()**

### DIFF
--- a/app/main.py
+++ b/app/main.py
@@ -6,11 +6,11 @@ def cache(func: Callable) -> Callable:
 
     def wrapper(*args) -> int:
         key = (func.__name__, args)
-        if key not in results:
-            print("Calculating new result")
-            results[key] = func(*args)
-        else:
+        if key in results:
             print("Getting from cache")
+            return results.get(key)
+        print("Calculating new result")
+        results[key] = func(*args)
         return results.get(key)
 
     return wrapper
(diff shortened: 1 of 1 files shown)


app/main.py
```diff
@@ -2,5 +2,15 @@
...
+    results = {}
+
+    def wrapper(*args) -> int:
+        key = (func.__name__, args)
+        if key not in results:
+            print("Calculating new result")
+            results[key] = func(*args)
+        else:
+            print("Getting from cache")
+        return results.get(key)
```
MNDonut: In order to simplify code readability, you can swap conditions
```
if key in results:
    ....
    return

....
results[key] = ...
return
```

Note that you actually don't need else statement here because if key is in results, it'll return it and will no reach the code after if statement ;)
You could do the same if you added return statement in if statement, then you would be able to remove else statement

MNDonut (changes requested):

app/main.py
```diff
@@ -2,5 +2,15 @@
...
 def cache(func: Callable) -> Callable:
-    # Write your code here
-    pass
+    results = {}
+
+    def wrapper(*args) -> int:
+        key = (func.__name__, args)
+        if key in results:
+            print("Getting from cache")
+            return results.get(key)
```
MNDonut, line 11: Also it's a good practice to add a blank line after if statement

MNDonut (approved):

MNDonut (approved):

PR #1199, opened 2024-07-13
Solution

LevAndrii:


app/main.py
```diff
@@ -2,5 +2,16 @@
 
 
 def cache(func: Callable) -> Callable:
-    # Write your code here
-    pass
+    storage = {}
+
+    def wrapper(*args) -> any:
```
```suggestion 
    @wraps(func)
    def wrapper(*args, **kwargs) -> Any:
```
- Good to add an annotation from `functools` module

NazarSheremet (changes requested):
Need to fix sugestions

NazarSheremet (approved):
Good job!

PR #1200, opened 2024-07-15
Solution

DSahalatyi:


app/main.py
```diff
@@ -2,5 +2,14 @@
...
+    saved_runs = {}
+
+    def wrapper(*args) -> (int, float):
+        if args not in saved_runs.keys():
+            print("Calculating new result")
+            saved_runs[args] = func(*args)
+            return saved_runs[args]
+        else:
+            print("Getting from cache")
+            return saved_runs[args]
```
MNDonut: It's recommended to do a direct condition first(not always, but when you have if and else, then you can simplify it) and then else because it'll be easier to read like
```
# this is easier to read
if something:
    pass
# else means "not something"
else:
    pass

# than this one
if not something:
    pass
# else means "**not** **not** something"
else:
    pass
```

And in general you don't need else statement here because "if" statement returns value, thus it'll not reach the code after this ;)

MNDonut (changes requested):

app/main.py
```diff
@@ -2,5 +2,14 @@
 
 
 def cache(func: Callable) -> Callable:
-    # Write your code here
-    pass
+    saved_runs = {}
+
+    def wrapper(*args) -> (int, float):
+        if args in saved_runs.keys():
```
oliasymak24, line 8: The check `if args in saved_runs` is more efficient than `if args in saved_runs.keys(),` as checking for the presence of a key in a dictionary by default checks the keys.

oliasymak24 (approved):
You change everything correct, good job!
P.s. left small comment :)

PR #1201, opened 2024-07-17
Solution

Atikiho:


app/main.py
```diff
@@ -2,5 +2,14 @@
...
+    cache_values = {}
+
+    def inner(*args) -> callable:
+        if args in cache_values:
+            print("Getting from cache")
+            return cache_values[args]
+        else:
+            cache_values[args] = func(*args)
+            print("Calculating new result")
+            return cache_values[args]
```
AnyoneClown: Try to rewrite your decorator not to have return duplicated. You can put return after a conditional statement

AnyoneClown (changes requested):

app/main.py
```diff
@@ -2,5 +2,14 @@
...
-    # Write your code here
-    pass
+    cache_values = {}
+
+    def inner(*args) -> callable:
+        if args in cache_values:
+            print("Getting from cache")
+            return cache_values[args]
+        else:
+            cache_values[args] = func(*args)
```

Nattalli (approved):

PR #1202, opened 2024-07-17
Solution

razumovskayada:


app/main.py
```diff
@@ -1,6 +1,19 @@
...
 
 def cache(func: Callable) -> Callable:
-    # Write your code here
-    pass
+    former_functions = {}
+
+    def wrapper(*args, **kwargs) -> Any:
+        nonlocal former_functions
+        values = tuple([value for value in kwargs.values()])
+        arguments = (*args, *values)
```
```suggestion
        key = (func.__name__, args,  frozenset(kwargs.items(),)
```

app/main.py
```diff
@@ -1,6 +1,19 @@
...
+from typing import Callable, Any
 
 
 def cache(func: Callable) -> Callable:
-    # Write your code here
-    pass
+    former_functions = {}
+
+    def wrapper(*args, **kwargs) -> Any:
+        nonlocal former_functions
```
vsmutok: this line redundan here

app/main.py
```diff
@@ -1,6 +1,19 @@
-from typing import Callable
+from typing import Callable, Any
 
 
 def cache(func: Callable) -> Callable:
-    # Write your code here
-    pass
+    former_functions = {}
+
+    def wrapper(*args, **kwargs) -> Any:
```

vsmutok (changes requested):

app/main.py
razumovskayada: can you please explain why the nonlocal line is rebundant? i thought we need to use it in order to permanently change the "former_functions" variable in the cache function's scope

app/main.py
```diff
@@ -1,6 +1,19 @@
...
+    @functools.wraps(func)
+    def wrapper(*args, **kwargs) -> Any:
+        key = (func.__name__, args, frozenset(kwargs.items(),))
+        if key in former_functions:
+            print("Getting from cache")
+            return former_functions[key]
+        else:
+            print("Calculating new result")
+            former_functions[key] = func(*args, **kwargs)
+            return former_functions[key]
```
vasylhnatiuk: use only one return in wrapper()

vasylhnatiuk (changes requested):

vsmutok (approved):
Good Job!

PR #1203, opened 2024-07-19
pull with errors (probably)

BastovOleksandr:


BastovOleksandr:
quite a questionable decision, but I don't understand how to control cache clearing without the context of the tests themselves

app/main.py
```diff
@@ -1,6 +1,31 @@
...
+                return calc_data(*args)
+
+        cached_data[func.__name__] = {}
+        return calc_data(*args)
+
+    def calc_data(*args) -> Any:
+        print("Calculating new result")
+        new_data = func(*args)
+        cached_data[func.__name__][args] = new_data
+        return new_data
```
LiudmylaKulinchenko: There's no need to create this function. You can avoid creating new `new_data` variable and rewrite it in one line

app/main.py
```diff
@@ -1,6 +1,31 @@
...
+    def wrapper(*args) -> Callable:
+        if func.__name__ in cached_data:
+            if args in cached_data[func.__name__]:
+                print("Getting from cache")
+                return cached_data[func.__name__][args]
+            else:
+                return calc_data(*args)
+
+        cached_data[func.__name__] = {}
+        return calc_data(*args)
```
LiudmylaKulinchenko: try to avoid duplicating returns

app/main.py
```diff
@@ -1,6 +1,31 @@
...
+
+
+def clear_cache() -> None:
+    cached_data.clear()
 
 
 def cache(func: Callable) -> Callable:
-    # Write your code here
-    pass
+    @wraps(func)
```
LiudmylaKulinchenko: You should create cached_data dict in `cache` function. That's how you'll use closure in this case

tests/conftest.py
```diff
@@ -0,0 +1,9 @@
+from typing import Any
+
+import pytest
+
+
+@pytest.fixture(autouse=True)
+def clear_cache(request: Any) -> None:
+    from app.main import clear_cache
```
LiudmylaKulinchenko: All imports always should be at the top of a file. Also make sure you have a correct order:
1. Built-in libraries and modules
2. Third-party (installed)
3. Custom

app/main.py
```diff
@@ -1,6 +1,31 @@
-from typing import Callable
+from typing import Callable, Any
+from functools import wraps
+
+
+def clear_cache() -> None:
+    cached_data.clear()
```
LiudmylaKulinchenko: There won't be need in this function if you'll just create a dict in `cache` function. It will create a new object every time we use a decorator for a new function

LiudmylaKulinchenko (changes requested):

MNDonut (approved):

PR #1204, opened 2024-07-19
'Solution'

dartomOOv:


app/main.py
```diff
@@ -1,6 +1,24 @@
 from typing import Callable
+import functools
+cache_dict = {}
```
LiudmylaKulinchenko: What if you're gonna wrap several functions with the decorator, will it work? You will have same cache for all wrapped functions. You should create this variable inside of `cache` function

app/main.py
```diff
@@ -1,6 +1,24 @@
...
+    global cache_dict
+
+    @functools.wraps(func)
+    def wrapper(*args) -> int:
+        global cache_dict
+
+        if args not in cache_dict[func.__name__]:
+            cache_dict[func.__name__][args] = func(*args)
+            print("Calculating new result")
+            return cache_dict[func.__name__][args]
```
LiudmylaKulinchenko: You can avoid return duplicating

LiudmylaKulinchenko (changes requested):

app/main.py
dartomOOv: But why, if I create "cache_dict" inside of decorator cache isn't it gonna re-assign itself to empty list when decorator is initializing? I done that to create inside this "cache_dict" a key with name of a decorated function and a value as another dict with function own cache. @LiudmylaKulinchenko

MNDonut: @dartomOOv no, decorator is initialized only once(if you use syntax sugar), so the code out of wrapper is run only once. But wrapper is called each time you call the function that is wrapper by this decorator. You can set breakpoint on cache variable inside decorator, call **decorated** function and check how many breakpoints you'll get
Example:
```
from typing import Callable


    cache_dict = {}

    def wrapper(*args) -> int:
        nonlocal cache_dict

        if args not in cache_dict:
        else:
            print("Getting from cache")

        return cache_dict[args]

    return wrapper

@cache
def func(arg):
    return arg + 1

func(1)
func(2)
func(2)
# cache_dict = {} will be called only once
```
It's because you use syntax sugar and when python runs the code, it knows that **cache** is decorator.
However if you remove decorator from **func** and run it in another way(without syntax sugar) like this:
```
cache(func)(1)
cache(func)(2)
cache(func)(2)
```
then yes, cache_dict will be cleared on each call. This is the magic of syntax sugar :)

MNDonut (approved):

app/main.py
dartomOOv: Ok, I see now. Thx very much!

PR #1206, opened 2024-07-20
add cache function

vladislav-tsybuliak1:


app/main.py
```diff
@@ -1,6 +1,19 @@
...
+    @functools.wraps(func)
+    def wrapper(*args, **kwargs) -> Any:
+        key = (func.__name__, args, frozenset(kwargs.items()))
+        if key in stored_data:
+            print("Getting from cache")
+            return stored_data[key]
+
+        print("Calculating new result")
+        stored_data[key] = func(*args, **kwargs)
+        return stored_data[key]
```
LiudmylaKulinchenko: You can avoid duplicating returns here

app/main.py
```diff
@@ -1,6 +1,19 @@
...
 
 
 def cache(func: Callable) -> Callable:
-    # Write your code here
-    pass
+    stored_data = {}
+
+    @functools.wraps(func)
+    def wrapper(*args, **kwargs) -> Any:
+        key = (func.__name__, args, frozenset(kwargs.items()))
```
LiudmylaKulinchenko: Use more descriptive name here instead of `key`

LiudmylaKulinchenko (changes requested):
Small fixes, looks good overall!

MNDonut (approved):

PR #1207, opened 2024-07-20
Solved

Alexandr-Politov:
Sorry. Somehow test file have been added to my commit.

### DIFF
--- a/app/main.py
+++ b/app/main.py
@@ -2,5 +2,19 @@ from typing import Callable
 
 
 def cache(func: Callable) -> Callable:
-    # Write your code here
-    pass
+
+    results_storage = {}
+
+    def inner_checker(*args) -> str:
+
+        key = args
+        if key in results_storage:
+            print("Getting from cache")
+            return results_storage[key]
+        else:
+            print("Calculating new result")
+            value = func(*args)
+            results_storage[key] = value
+            return value
+
+    return inner_checker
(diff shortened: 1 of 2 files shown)


app/main.py
```diff
@@ -2,5 +2,19 @@
...
+
+        key = args
+        if key in results_storage:
+            print("Getting from cache")
+            return results_storage[key]
+        else:
+            print("Calculating new result")
+            value = func(*args)
+            results_storage[key] = value
+            return value
```
LiudmylaKulinchenko: No need to create additional `value` variable here. After this fix you could also move return outsite of if/else to avoid return duplicating

app/main.py
```diff
@@ -2,5 +2,19 @@
...
 
 def cache(func: Callable) -> Callable:
-    # Write your code here
-    pass
+
+    results_storage = {}
+
+    def inner_checker(*args) -> str:
+
+        key = args
```
LiudmylaKulinchenko: key is not a really descriptive name. You can leave `args` to show what exaclty you want a key in your dictionary to be

LiudmylaKulinchenko (changes requested):

app/main.py
```diff
@@ -2,5 +2,17 @@
...
-    # Write your code here
-    pass
+
+    results_storage = {}
+
+    def inner_checker(*args) -> str:
+
+        if args in results_storage:
+            print("Getting from cache")
+            return results_storage[args]
```
LiudmylaKulinchenko: Will it work the same if you remove this return statement?

LiudmylaKulinchenko (changes requested):

LiudmylaKulinchenko (approved):
Great job! 🌸

PR #1209, opened 2024-07-21
Solution

mihavryliuk:


app/main.py
```diff
@@ -1,6 +1,17 @@
...
+
+    def wrapper(*args: Any) -> Any:
+        if args in cache_store:
+            print("Getting from cache")
+            return cache_store[args]
+        else:
+            print("Calculating new result")
+            result = func(*args)
+            cache_store[args] = result
+            return result
```
MNDonut: you don't need else because if statement has return, it means it will not reach the code after it :)

MNDonut (changes requested):

MNDonut (approved):

PR #1210, opened 2024-07-21
Solution

dpiuro:


app/main.py
```diff
@@ -1,6 +1,27 @@
...
 def cache(func: Callable) -> Callable:
-    # Write your code here
-    pass
+    cache_storage: Dict[Tuple, Any] = {}
+
+    def wrapper(*args) -> Any:
+        if args in cache_storage:
+            print("Getting from cache")
+            return cache_storage[args]
+        else:
```
MNDonut: you don't need else statement here because if has return, thus it'll not reach the code after it

MNDonut (changes requested):

app/main.py
dpiuro: Thanks, else has been removed

app/main.py
```diff
@@ -1,6 +1,27 @@
...
+
+    def wrapper(*args) -> Any:
+        if args in cache_storage:
+            print("Getting from cache")
+            return cache_storage[args]
+
+        print("Calculating new result")
+        result = func(*args)
+        cache_storage[args] = result
+        return result
```
LiudmylaKulinchenko: No need to create result variable here
You'll be able to avoid duplicating return after making the change 🤠

LiudmylaKulinchenko (changes requested):

app/main.py
dpiuro: Thanks, it seems i fixed it

LiudmylaKulinchenko (approved):
Good job 🤠

PR #1213, opened 2024-07-21
Solution

vmykytenko:


app/main.py
```diff
@@ -2,5 +2,17 @@
 
 
 def cache(func: Callable) -> Callable:
-    # Write your code here
-    pass
+
+    cache_results = {}
+
+    def inner(*args) -> None:
```
vsmutok: what if function takes not only `args` but also `kwargs`?

vsmutok (changes requested):

EdAlekseiev (approved):

PR #1214, opened 2024-07-21
Cache decorator

yashkunn:


app/main.py
```diff
@@ -1,6 +1,27 @@
...
-    # Write your code here
-    pass
+    cache_storage = {}
+
+    def wrapper(*args: Callable) -> object:
+        if args in cache_storage:
+            print("Getting from cache")
+            return cache_storage[args]
+        else:
+            print("Calculating new result")
```
MNDonut: you don't need else because if has return, it means it'll not reach the code after it

MNDonut (changes requested):

app/main.py
```diff
@@ -1,6 +1,27 @@
...
+
+    def wrapper(*args: Callable) -> object:
+        if args in cache_storage:
+            print("Getting from cache")
+            return cache_storage[args]
+
+        print("Calculating new result")
+        result = func(*args)
+        cache_storage[args] = result
+        return result
```
LiudmylaKulinchenko: You can avoid creating `result` variable here

LiudmylaKulinchenko (changes requested):

MNDonut (approved):

PR #1215, opened 2024-07-21
Solution

Bogdan-Rylov:


app/main.py
```diff
@@ -1,6 +1,32 @@
...
+    def wrapper(*args, **kwargs) -> Any:
+        is_cachable = True
+        if kwargs:
+            is_cachable = False
+        else:
+            for arg in args:
+                if not isinstance(
+                        arg, (bool, int, float, str, tuple, frozenset, bytes)
+                ):
+                    is_cachable = False
```
MNDonut: In this task it isn't mentioned that you can get kwargs, but if you decided to handle this case, then one note that it would be better to use "if isinstance" and set check mutable types. Easier to read and you see what types are mutable

MNDonut (changes requested):

app/main.py
```diff
@@ -1,6 +1,27 @@
...
 
 
 def cache(func: Callable) -> Callable:
-    # Write your code here
-    pass
+    cached_results = {}
+
+    @wraps(func)
+    def wrapper(*args) -> Any:
+        is_cachable = True
```
Oleksl888: I see you went an extra mile and added this check. Good job on that!

But on the other hand it is not required in task definition and overcomplicates the code, this function can be written in 6 lines with only difference that it will raise an exception if you try to add unhashable type to the dict as a key. 
As well when you perform your check it may slow down your code even more in case if you have a long list of arguments.

app/main.py
```diff
@@ -1,6 +1,27 @@
...
+                break
+
+        if is_cachable:
+            for arguments, result in cached_results.items():
+                if arguments == args:
+                    print("Getting from cache")
+                    return result
+        print("Calculating new result")
+        calculated_result = func(*args)
+        if is_cachable:
```
Oleksl888: Can you avoid having two of the same if statements and returns?

Oleksl888 (changes requested):
I see what you did there and it's good forward thinking but can you please refactor your code so it is more optimal? No need to make this check about cachable/uncachable

app/main.py
```diff
@@ -1,6 +1,19 @@
...
 
 
 def cache(func: Callable) -> Callable:
-    # Write your code here
-    pass
+    cached_results = {}
+
+    @wraps(func)
+    def wrapper(*args) -> Any:
+        for arguments, result in cached_results.items():
```
Oleksl888: here you can just check if `key` in `cached_results`, so no need to use `items()` method

app/main.py
```diff
@@ -1,6 +1,19 @@
...
+            if arguments == args:
+                print("Getting from cache")
+                return result
```
Oleksl888: Try to rewrite with one return statement like so:
if data not in cache -> add data to cache
...
return data from cache

app/main.py
```diff
@@ -1,6 +1,19 @@
-from typing import Callable
+from typing import Callable, Any
+from functools import wraps
```
Oleksl888, line 1: check order of imports (it is a convention to sort imports alphabetically

Oleksl888 (changes requested):

MNDonut (approved):

PR #1218, opened 2024-07-23
Task 4

Poznanskyi:


app/main.py
```diff
@@ -1,6 +1,29 @@
...
+    @wraps(func)
+    def wrapper(*args: Tuple[Any, ...]) -> Any:
+        if args in cache_store:
+            print("Getting from cache")
+            return cache_store[args]
+        else:
+            print("Calculating new result")
+            result = func(*args)
+            cache_store[args] = result
+            return result
```
Oleksl888: Can you simplify this so you have only one return statement?

app/main.py
```diff
@@ -1,6 +1,29 @@
...
+
+    return wrapper
+
+
+@cache
```
Oleksl888: Examples shouldn't be included, please remove

Oleksl888 (changes requested):

MNDonut (approved):

PR #1224, opened 2024-07-31
implemennt cache decorator

Morphin20th:


app/main.py
```diff
@@ -2,5 +2,16 @@
 
 
 def cache(func: Callable) -> Callable:
-    # Write your code here
-    pass
+    cache_data = {}
+
+    def wrapper(*args) -> int | tuple:
```
pavlopro: Also include **kwargs and correct return annotation to any

app/main.py
```diff
@@ -2,5 +2,16 @@
...
+        if args not in cache_data:
+            print("Calculating new result")
+            result = func(*args)
+            cache_data[args] = result
+            return cache_data[args]
+        else:
```
pavlopro: else operator is redundant

pavlopro (changes requested):

pavlopro (approved):
Good Job!

PR #1225, opened 2024-07-31
Solution

alheimurin-n:


app/main.py
```diff
@@ -2,5 +2,15 @@
 
 
 def cache(func: Callable) -> Callable:
-    # Write your code here
-    pass
+    cache_dict = {}
+
+    def wrapper(*args) -> int:
```
NazarSheremet: - Good to add an annotation from the functools module
- Also, it should work with **kwargs. Create a unique key using args and kwargs
```suggestion
     @wraps(func)
    def wrapper(*args, **kwargs) -> int:
```

NazarSheremet (changes requested):
Need to fix sugestions

app/main.py
alheimurin-n: @NazarSheremet could you please explain why you've rejected my solution? In course there are no mentions about @wraps(func)

app/main.py
```diff
@@ -2,5 +2,15 @@
...
-    # Write your code here
-    pass
+    cache_dict = {}
+
+    def wrapper(*args, **kwargs) -> int:
+        if args in cache_dict:
+            print("Getting from cache")
+        else:
+            print("Calculating new result")
+            cache_dict[args] = func(*args)
```
NazarSheremet, line 12: The function receives both args and kwargs as input. And kwargs should also be stored in the cache.
```suggestion
            cache_dict[args] = func(*args, **kwargs)
```

NazarSheremet (changes requested):

YuriiSmolii (approved):
LGTM

PR #1227, opened 2024-07-31
Solution

oleksashcherbakov:


app/main.py
```diff
@@ -2,5 +2,15 @@
 
 
 def cache(func: Callable) -> Callable:
-    # Write your code here
-    pass
+    collected_data = {}
+
+    def inner(*args, **kwargs) -> int:
+        if args in collected_data:
```
pavlopro, line 8: how about kwargs?

app/main.py
```diff
@@ -2,5 +2,15 @@
...
+            print("Getting from cache")
+            return collected_data[args]
+        else:
```
pavlopro, line 11: `else` operator is redundant

app/main.py
```diff
@@ -2,5 +2,15 @@
 
 
 def cache(func: Callable) -> Callable:
-    # Write your code here
-    pass
+    collected_data = {}
+
+    def inner(*args, **kwargs) -> int:
```
pavlopro, line 7: correct return annotation to any

pavlopro (changes requested):

pavlopro (changes requested):
Comments weren't resolved

pavlopro (changes requested):
Comments weren't resolved

app/main.py
Dimosphen1: The comment is no resolved, indicate in the return type annotation that you return `Any`

app/main.py
Dimosphen1: The comment is not resolved

app/main.py
Dimosphen1: Please, add `kwargs` processing as well

Dimosphen1 (changes requested):

viktoria-rybenchuk (changes requested):
Resolve all comments or visit Q&A session if you need a help

PR #1229, opened 2024-08-01
Solution

CheshireKate:


app/main.py
```diff
@@ -1,6 +1,16 @@
-from typing import Callable
+from typing import Callable, Any
 
 
 def cache(func: Callable) -> Callable:
-    # Write your code here
-    pass
+    cache_store = {}
+
+    def inner(*args) -> Any:
```
vsmutok: What if the function accepts keyword arguments?
```suggestion
    def inner(*args, **kwargs) -> Any:
```

vsmutok (changes requested):

app/main.py
```diff
@@ -1,6 +1,16 @@
...
+        if args in cache_store:
```

vsmutok (changes requested):

app/main.py
```diff
@@ -1,6 +1,18 @@
-from typing import Callable
+from typing import Callable, Any
 
 
 def cache(func: Callable) -> Callable:
-    # Write your code here
-    pass
+    cache_store = {}
+
+    def inner(*args, **kwargs) -> Any:
```

app/main.py
```diff
@@ -1,6 +1,18 @@
...
+        kwargs_tuple = tuple(sorted(kwargs.items()))
+        key = (args, kwargs_tuple)
```
```suggestion
        key = (args, kwargs_tuple, func.__name__,)
```

app/main.py
```diff
@@ -1,6 +1,18 @@
...
+        if key in cache_store:
+            print("Getting from cache")
+            return cache_store[key]
+        else:
+            print("Calculating new result")
+            result = func(*args, **kwargs)
+            cache_store[key] = result
+            return result
```
```suggestion
        if key in cache_store:
            print("Getting from cache")
        else:
            print("Calculating new result")
            cache_store[key] = func(*args, **kwargs)
        return cache_store[key]
```

vsmutok (changes requested):

app/main.py
```diff
@@ -1,6 +1,18 @@
...
+    @wraps(func)
+    def wrapper(*args, **kwargs) -> Any:
+        kwargs_tuple = tuple(sorted(kwargs.items()))
+        key = (args, kwargs_tuple, func.__name__,)
+        if key in cache_store:
+            print("Getting from cache")
+        else:
+            print("Calculating new result")
+            cache_store[key] = func(*args, **kwargs)
+            return cache_store[key]
```
```suggestion
        else:
            print("Calculating new result")
            cache_store[key] = func(*args, **kwargs)
        return cache_store[key]
```

vsmutok (changes requested):

vsmutok (approved):
Well Done!

PR #1230, opened 2024-08-02
add solution for task

nikvitkaua:


app/main.py
```diff
@@ -2,5 +2,16 @@
 
 
 def cache(func: Callable) -> Callable:
-    # Write your code here
-    pass
+    result_cache = {}
+
+    def inner(*args) -> Callable:
```
```suggestion
    def inner(*args, **kwargs) -> Any:
```

- Use `functools.wraps`
- What if the function accepts keyword arguments? 
- Wrong type annotation

app/main.py
```diff
@@ -2,5 +2,16 @@
...
+        if args in result_cache:
```

vsmutok (changes requested):

app/main.py
```diff
@@ -1,6 +1,22 @@
...
+from functools import wraps
 
 
 def cache(func: Callable) -> Callable:
-    # Write your code here
-    pass
+    result_cache = {}
+
+    @wraps(func)
+    def inner(*args, **kwargs) -> Any:
```
vasylhnatiuk, line 9: use only one return in inner()

vasylhnatiuk (changes requested):

app/main.py
```diff
@@ -1,6 +1,21 @@
...
+        key = (func.__name__, args, frozenset(kwargs.items()))
+
+        if key in result_cache:
+            print("Getting from cache")
+            return result_cache.get(key)
+        else:
+            print("Calculating new result")
+            result = func(*args, **kwargs)
+            result_cache[key] = result
+            return result
```
```suggestion
        if key in result_cache:
            print("Getting from cache")
        else:
            print("Calculating new result")
            result_cache[key] =  func(*args, **kwargs)
        return result_cache.get(key)
```

vsmutok (changes requested):

vsmutok (approved):
Good Job!

PR #1232, opened 2024-08-03
Solution

Serhii-Chubur:


Serhii-Chubur:
Чому flake8 виводить 300+ ідентичних помилок в одному і тому ж місці?

app/main.py
```diff
@@ -2,5 +2,15 @@
...
+
+    def wrapper(*args, **kwargs) -> int:
+        if args in runs:
+            print("Getting from cache")
+            return runs[args]
+        else:
+            result = func(*args, **kwargs)
+            runs[args] = result
+            print("Calculating new result")
+            return result
```
vasylhnatiuk: use only one return in wrapper()

vasylhnatiuk (changes requested):

app/main.py
```diff
@@ -2,5 +2,15 @@
 
 
 def cache(func: Callable) -> Callable:
-    # Write your code here
-    pass
+    runs = {}
+
```
vsmutok, line 7: - Use `functools.wraps`

app/main.py
```diff
@@ -2,5 +2,15 @@
...
+    def wrapper(*args, **kwargs) -> int:
+        if args in runs:
```

vsmutok (changes requested):

app/main.py
```diff
@@ -1,6 +1,22 @@
...
+        nonlocal name
+        arguments = (args, frozenset(kwargs.items()))
+        if arguments in runs and name == func.__name__:
+            print("Getting from cache")
+        else:
+            name = func.__name__
+            result = func(*args, **kwargs)
+            runs[arguments] = result
+            print("Calculating new result")
+        return runs[arguments]
```
```suggestion
    runs = {}

    @wraps(func)
    def wrapper(*args, **kwargs) -> int:
        key = (args, frozenset(kwargs.items()), func.__name__,)
        if key in runs:
            print("Getting from cache")
        else:
            runs[key] = func(*args, **kwargs)
            print("Calculating new result")
        return runs[key]
```

vsmutok (changes requested):

vsmutok (approved):
Good Job!

PR #1235, opened 2024-08-08
Solution

SerhiyNecha:


app/main.py
```diff
@@ -2,5 +2,17 @@
 
 
 def cache(func: Callable) -> Callable:
-    # Write your code here
-    pass
+    stored_kwargs = {}
+
+    def stored(*args, **kwargs) -> Callable:
```
vasylhnatiuk: use only one return in stored()

vasylhnatiuk (changes requested):

app/main.py
```diff
@@ -2,5 +2,16 @@
...
```
LLkaia: Use `wraps` from functolls here

app/main.py
```diff
@@ -2,5 +2,16 @@
...
+        checked_kwargs = tuple(kwargs.items())
+        checked = (args, checked_kwargs)
```
LLkaia: It is a good approach to save function name here as well to handle case when functions have same args

LLkaia (changes requested):

app/main.py
```diff
@@ -2,5 +2,16 @@
 
 
 def cache(func: Callable) -> Callable:
-    # Write your code here
-    pass
+    stored_kwargs = {}
+
+    def wrapper(*args, **kwargs) -> Callable:
```
LLkaia, line 9:
```suggestion
    @wraps(func)
    def wrapper(*args, **kwargs) -> Callable:
```

app/main.py
```diff
@@ -2,5 +2,16 @@
```
LLkaia, line 3:
```suggestion
from functools import wraps
```

app/main.py
```diff
@@ -2,5 +2,16 @@
 
 
 def cache(func: Callable) -> Callable:
-    # Write your code here
-    pass
+    stored_kwargs = {}
+
+    def wrapper(*args, **kwargs) -> Callable:
+        checked_kwargs = tuple(kwargs.items())
+        checked = (args, checked_kwargs)
```
```suggestion
        checked = (args, checked_kwargs, func.__name__)
```

LLkaia (changes requested):

vasylhnatiuk (approved):

PR #1236, opened 2024-08-09
Solution

ip-ilya:


app/main.py
```diff
@@ -1,6 +1,18 @@
-from typing import Callable
+from typing import Callable, Any
 
 
 def cache(func: Callable) -> Callable:
-    # Write your code here
-    pass
+    cache_dict = {}
+
+    def wrapper(*args) -> Any:
```
vsmutok: - Use `functools.wraps`

app/main.py
```diff
@@ -1,6 +1,18 @@
...
+        if args in cache_dict:
```

vsmutok (changes requested):

app/main.py
```diff
@@ -1,6 +1,22 @@
...
 def cache(func: Callable) -> Callable:
-    # Write your code here
-    pass
+    cache_dict = {}
+
+    @functools.wraps(func)
+    def wrapper(*args, **kwargs) -> Any:
+        func_details = (args, tuple(kwargs.items()), func.__name__)
+
+        if func_details in cache_dict.keys():
```
pavlopro: Simplify to
`if func_details in cache_dict:`

app/main.py
```diff
@@ -1,6 +1,22 @@
...
+            print("Getting from cache")
+            return cache_dict.get(func_details)
```
pavlopro: `cache_dict[func_details]`

app/main.py
```diff
@@ -1,6 +1,22 @@
-from typing import Callable
+from typing import Callable, Any
+import functools
```
pavlopro: Import only those what you need
`from functools import wraps`

pavlopro (changes requested):

LLkaia (approved):
Looks Perfect!

PR #1237, opened 2024-08-09
'Solution'

Bogdanneww:


app/main.py
```diff
@@ -1,6 +1,17 @@
-from typing import Callable
+from typing import Callable, Any
 
 
 def cache(func: Callable) -> Callable:
-    # Write your code here
-    pass
+    cache_storage = {}
+
+    def wrapper(*args) -> Any:
```
vasylhnatiuk: use only one return in wrapper(()

vasylhnatiuk (changes requested):

app/main.py
```diff
@@ -1,6 +1,17 @@
...
+        if args in cache_storage:
+            print("Getting from cache")
```

vsmutok (changes requested):

app/main.py
```diff
@@ -1,6 +1,20 @@
...
+
+        if cache_key in cache_storage:
+            print("Getting from cache")
+            result = cache_storage[cache_key]
+        else:
+            print("Calculating new result")
+            result = func(*args, **kwargs)
+            cache_storage[cache_key] = result
+
+        return result
```
```suggestion
        if cache_key in cache_storage:
            print("Getting from cache")
        else:
            print("Calculating new result")
            cache_storage[cache_key] = func(*args, **kwargs)

        return cache_storage[cache_key]
```

vsmutok (changes requested):

vasylhnatiuk (approved):

PR #1238, opened 2024-08-09
Solution

Deadmarvald:


NazarSheremet (changes requested):
Solution should also work with `**kwargs`. Create a unique key using `args`and `kwargs`

YuriiSmolii (approved):
LGTM

PR #1239, opened 2024-08-09
Solution

mykytafabrykator:


app/main.py
```diff
@@ -1,6 +1,19 @@
...
+from functools import wraps
 
 
 def cache(func: Callable) -> Callable:
-    # Write your code here
-    pass
+    cache_dict = {}
+
+    @wraps(func)
+    def inner(*args: tuple) -> Any:
```
vasylhnatiuk: use only one return in inner()

vasylhnatiuk (changes requested):

app/main.py
```diff
@@ -1,6 +1,20 @@
...
+        nonlocal cache_dict
+
+        if args in cache_dict.keys():
```

vsmutok (changes requested):

app/main.py
```diff
@@ -1,6 +1,20 @@
...
+
+    @wraps(func)
+    def inner(*args: tuple, **kwargs) -> Any:
+        key = (func.__name__, *args, tuple(kwargs))
+
+        if key in cache_dict.keys():
+            print("Getting from cache")
+        else:
+            print("Calculating new result")
+            cache_dict[key] = func(*args)
```
```suggestion
            cache_dict[key] = func(*args, **kwargs)
```

vsmutok (changes requested):

app/main.py
```diff
@@ -1,6 +1,20 @@
...
+
+    @wraps(func)
+    def inner(*args: tuple, **kwargs) -> Any:
+        key = (func.__name__, *args, tuple(kwargs))
+
+        if key in cache_dict.keys():
+            print("Getting from cache")
+        else:
+            print("Calculating new result")
+            cache_dict[key] = func(*args, **kwargs)
```
```suggestion
        if key in cache_dict:
            print("Getting from cache")
        else:
            print("Calculating new result")
            cache_dict[key] = func(*args, **kwargs)
```
`keys()` used by default

vsmutok (changes requested):

vsmutok (approved):
Great!

PR #1240, opened 2024-08-09
solution

nrjss67:


app/main.py
```diff
@@ -1,6 +1,17 @@
-from typing import Callable
+from typing import Callable, Any
 
 
 def cache(func: Callable) -> Callable:
-    # Write your code here
-    pass
+    storage = {}
+
+    def inner(*args) -> Any:
```
vsmutok: use `functools.wraps` here

app/main.py
```diff
@@ -1,6 +1,17 @@
...
+        if args not in storage:
```

vsmutok (changes requested):

app/main.py
```diff
@@ -1,6 +1,20 @@
...
+    def inner(*args, **kwargs) -> Any:
+        arguments = args or kwargs
+        if arguments not in storage:
+            print("Calculating new result")
+            func_result = func(*arguments)
+            storage[arguments] = func_result
+            return func_result
+        else:
+            print("Getting from cache")
+            return storage[arguments]
```
```suggestion
        key = (func.__name__, *args, tuple(kwargs))

        if key in storage:
            print("Getting from cache")
        else:
            print("Calculating new result")
            storage[key] = func(*args, **kwargs)

        return storage[key]
```

vsmutok (changes requested):

vsmutok (approved):
Well Done!

PR #1241, opened 2024-08-09
create cache decorator

Anzda:


app/main.py
```diff
@@ -2,5 +2,17 @@
 
 
 def cache(func: Callable) -> Callable:
-    # Write your code here
-    pass
+
+    allredy = {}
+
+    def wraper(*args, **kvargs) -> int:
```
vsmutok, line 8: * use `functools.wraps`
* `nonlocal` redundant in this case 
* You should check both `args` and `kwargs` because the function might accept the same `args` but different `kwargs`. Additionally, I recommend checking the function name since there could be two different functions that accept the same arguments.

vsmutok (changes requested):

PR #1245, opened 2024-08-12
Made one return in inner()

arsen-arutiunov:


app/main.py
```diff
@@ -2,5 +2,15 @@
...
+
+    def inner(*args, **kwargs) -> (int, list):
+        if str(args) in cache_value:
+            print("Getting from cache")
+            result = cache_value[str(args)]
+        else:
+            print("Calculating new result")
+            result = func(*args, **kwargs)
+            cache_value[str(args)] = result
+        return result
```
vasylhnatiuk, line 15:
```suggestion
        return cache_value[str(args)]
```

app/main.py
```diff
@@ -2,5 +2,15 @@
...
-    pass
+    cache_value = {}
+
+    def inner(*args, **kwargs) -> (int, list):
+        if str(args) in cache_value:
+            print("Getting from cache")
+            result = cache_value[str(args)]
+        else:
+            print("Calculating new result")
+            result = func(*args, **kwargs)
```
vasylhnatiuk, line 13: useless line

vasylhnatiuk (changes requested):
you use str(args) 3 times - try to avoid it, write it in varible

vasylhnatiuk (changes requested):
fix

PR #1247, opened 2024-08-12
First commit

Kulibchenko:


app/main.py
```diff
@@ -2,5 +2,14 @@
 
 
 def cache(func: Callable) -> Callable:
-    # Write your code here
-    pass
+    result = dict()
+
+    def inner(*args) -> int:
```
YuriiSmolii: You should also use *kwargs

YuriiSmolii (changes requested):
Refactoring required.

NazarSheremet (approved):
Good job!

PR #1248, opened 2024-08-12
Solution

PonomaryovNikita:


app/main.py
```diff
@@ -2,5 +2,16 @@
 
 
 def cache(func: Callable) -> Callable:
-    # Write your code here
-    pass
+    results = {}
+
+    def wrapper(*args, **kwargs) -> int:
```
viktoria-rybenchuk: use @wraps from functools to provide meta data for decorator

viktoria-rybenchuk (changes requested):

app/main.py
```diff
@@ -1,6 +1,19 @@
...
+from functools import wraps
 
 
 def cache(func: Callable) -> Callable:
-    # Write your code here
-    pass
+    results = {}
+
+    @wraps(func)
+    def wrapper(*args, **kwargs) -> int:
```
Dimosphen1: Please, indicate that `wrapper` could return `Any` instead of `int` in the return type annotation

app/main.py
```diff
@@ -1,6 +1,19 @@
...
+        key = args + tuple(kwargs.items())
+        if key in results:
+            print("Getting from cache")
+            return results[key]
+        else:
```

Dimosphen1 (changes requested):

pavlopro (approved):
Good Job!

PR #1254, opened 2024-08-14
Solution

Hennadii29:


app/main.py
```diff
@@ -1,6 +1,16 @@
-from typing import Callable
+from typing import Callable, Any
 
 
 def cache(func: Callable) -> Callable:
-    # Write your code here
-    pass
+    cache_store = {}
+
+    def wrapper(*args: tuple) -> Any:
```
vsmutok: use `@wraps` from functools to provide meta data for decorator

app/main.py
```diff
@@ -1,6 +1,16 @@
...
+        if args in cache_store:
```

vsmutok (changes requested):

LLkaia (changes requested):
It's a bad practice to have couple `return` statements in function when it is possible to have only one.
So, do only one `return` in `wrapper` function

vasylhnatiuk (approved):

PR #1256, opened 2024-08-16
Solution

asynyshyn:


app/main.py
```diff
@@ -1,6 +1,15 @@
-from typing import Callable
+from typing import Callable, Any
 
 
 def cache(func: Callable) -> Callable:
-    # Write your code here
-    pass
+    stored_results = {}
+
+    def wrapper(*args: Any) -> Any:
```

vsmutok: use `@wraps` from functools to provide meta data for decorator

vsmutok (changes requested):

app/main.py
```diff
@@ -1,6 +1,21 @@
...
+        key = (func.__name__, args, tuple(sorted(kwargs.items())))
+
+        if key in stored_results:
+            print("Getting from cache")
+            return stored_results[key]
+
+        print("Calculating new result")
+        result = func(*args, **kwargs)
+        stored_results[key] = result
+        return result
```
```suggestion
        if key in stored_results:
            print("Getting from cache")
        else:
            print("Calculating new result")
            stored_results[key] = func(*args, **kwargs)
        return stored_results[key]

```

vsmutok (changes requested):

app/main.py
asynyshyn: thanks, updated

vsmutok (approved):
Well Done!

PR #1264, opened 2024-08-19
'Solution'

KostiantinPastukh:


app/main.py
```diff
@@ -1,6 +1,15 @@
...
+from typing import Callable, Any
 
 
 def cache(func: Callable) -> Callable:
-    # Write your code here
-    pass
+    cache_list = {}
+
+    def inner(*args, **kwargs) -> Any:
+        if args not in cache_list:
```
pavlopro: Include **kwargs as well

app/main.py
```diff
@@ -1,6 +1,15 @@
...
+            cache_list[args] = func(*args, **kwargs)
+            print("Calculating new result")
+            return cache_list[args]
+        else:
```
pavlopro, line 14: else operator is redundant

pavlopro (changes requested):

app/main.py
```diff
@@ -1,6 +1,14 @@
...
 
 
 def cache(func: Callable) -> Callable:
-    # Write your code here
-    pass
+    cache_list = {}
+
+    def inner(*args, **kwargs) -> Any:
+        if args not in cache_list:
+            cache_list[args or kwargs] = func(*args, **kwargs)
```
pavlopro: This is not correct. You should combine combine args and kwargs. And then check if both they are in cache_list

`cache_list` is bad variable name because it is dictionary type

pavlopro (changes requested):

app/main.py
KostiantinPastukh: @pavlopro Do I need kwargs here at all?

LLkaia: ofc, because function can take both positional and key parameters, suggest you read difference between them

app/main.py
```diff
@@ -1,6 +1,14 @@
...
 
 
 def cache(func: Callable) -> Callable:
-    # Write your code here
-    pass
+    cache_dict = {}
+
+    def inner(*args) -> Any:
+        if args not in cache_dict:
+            cache_dict[args] = func(*args)
```
LLkaia: You need to take into account both args and kwargs here. Also different functions can take same args, so let's create complex key with args, kwargs and func name
`(args, tuple(kwargs.items()), func.__name__)`

app/main.py
```diff
@@ -1,6 +1,14 @@
...
+            print("Calculating new result")
+            return cache_dict[args]
+        print("Getting from cache")
+        return cache_dict[args]
```
LLkaia, line 16: Please, achieve one return statement in inner function since it best practice

app/main.py
```diff
@@ -1,6 +1,14 @@
-from typing import Callable
+from typing import Callable, Any
 
 
 def cache(func: Callable) -> Callable:
-    # Write your code here
-    pass
+    cache_dict = {}
+
```
LLkaia, line 7: use `@wraps` decorator imported from functools. Read what the purpose of this decorator in inner function

LLkaia (changes requested):

app/main.py
KostiantinPastukh: @LLkaia What if I do it this way?
args += tuple(kwargs.values())

app/main.py
KostiantinPastukh: переклади на англійську "Як це зробити в данному випадку?"

KostiantinPastukh: @LLkaia How can I do this in this case? I don`t understand.

KostiantinPastukh:
Check last solution pls

PR #1266, opened 2024-08-20
Cache decorator

vladyslavgusakovskiy:


NazarSheremet (changes requested):
A unique key needs to be created that will consist of `*args` and `**kwargs` because the function might return incorrect values depending on the parameters passed as `**kwargs`

YuriiSmolii (approved):
Gj

PR #1267, opened 2024-08-21
implemented cache Decorator

Yurii-Durden:


app/main.py
```diff
@@ -1,6 +1,18 @@
...
 
 
 def cache(func: Callable) -> Callable:
-    # Write your code here
-    pass
+    saved_data = {}
+
+    @wraps(func)
+    def inner(*args) -> Any:
+        if args not in saved_data:
```

vsmutok (changes requested):

app/main.py
Yurii-Durden: Thanks fixed it

EdAlekseiev (approved):

PR #1270, opened 2024-08-23
py-cache-decorator

Mxbely:
Created decorate function "cache"
Inside this function was implemented "inner" function with save data from outside function into dictionary

### DIFF
--- a/app/main.py
+++ b/app/main.py
@@ -5,11 +5,11 @@ def cache(func: Callable) -> Callable:
     uniques_names_functions = {}
 
     def inner(*args) -> Any:
-        if args not in uniques_names_functions:
-            result_function = func(*args)
-            uniques_names_functions[args] = result_function
-            print("Calculating new result")
-        else:
+        if args in uniques_names_functions.keys():
             print("Getting from cache")
-        return uniques_names_functions[args]
+            return uniques_names_functions[*args]
+        else:
+            uniques_names_functions[*args] = func(*args)
+            print("Calculating new result")
+            return uniques_names_functions[*args]
     return inner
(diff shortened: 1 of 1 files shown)


app/main.py
```diff
@@ -1,6 +1,15 @@
...
 
 
 def cache(func: Callable) -> Callable:
-    # Write your code here
-    pass
+    uniques_names_functions = {}
+
+    def inner(*args) -> Any:
+        if args not in uniques_names_functions:
+            result_function = func(*args)
```
AnyoneClown: Your approach is incorrect: you should store arguments and the result of the function with these arguments(As a key of dict).

AnyoneClown (changes requested):

app/main.py
```diff
@@ -1,6 +1,15 @@
-from typing import Callable
+from typing import Callable, Any
 
 
 def cache(func: Callable) -> Callable:
-    # Write your code here
-    pass
+    uniques_names_functions = {}
```
Nattalli: you c an simplify, just `storage`

app/main.py
```diff
@@ -1,6 +1,15 @@
...
+
+    def inner(*args) -> Any:
+        if args in uniques_names_functions.keys():
+            print("Getting from cache")
+            return uniques_names_functions[args]
+        else:
+            uniques_names_functions[args] = func(*args)
+            print("Calculating new result")
+            return uniques_names_functions[args]
```
Nattalli: can you use one `return` statement for both cases?

Nattalli (changes requested):

AnyoneClown (approved):
LGTM!

PR #1271, opened 2024-08-23
'Solution'

NazarSlavych:


app/main.py
```diff
@@ -1,6 +1,20 @@
...
 
 
 def cache(func: Callable) -> Callable:
-    # Write your code here
-    pass
+    cache_store = {}
+
+    @wraps(func)
+    def wrapper(*args) -> None:
+        key = args
```

vsmutok (changes requested):

app/main.py
```diff
@@ -1,6 +1,20 @@
...
+    def wrapper(*args, **kwargs) -> None:
+        key = (func.__name__, args, tuple(sorted(kwargs.items())))
+        if key in cache_store:
+            print("Getting from cache")
+            return cache_store[key]
+        else:
+            print("Calculating new result")
+            result = func(*args, **kwargs)
+            cache_store[key] = result
+            return result
```
```suggestion
        if key in cache_store:
            print("Getting from cache")
        else:
            print("Calculating new result")
            cache_store[key] = func(*args, **kwargs)
        return cache_store[key]
```
This will make it look cleaner

vsmutok (changes requested):

vsmutok (approved):
Great!

PR #1272, opened 2024-08-26
Solution

frontbastard:


app/main.py
```diff
@@ -1,6 +1,22 @@
...
+T = TypeVar("T")
 
 
 def cache(func: Callable) -> Callable:
-    # Write your code here
-    pass
+    cache_dict = {}
+
+    @wraps(func)
+    def wrapper(*args) -> T:
```
vasylhnatiuk: use only one return in wrapper, and what is T?

vasylhnatiuk (changes requested):

app/main.py
```diff
@@ -1,6 +1,17 @@
...
+from typing import Callable, Any
 
 
 def cache(func: Callable) -> Callable:
-    # Write your code here
-    pass
+    cache_dict = {}
+
+    @wraps(func)
+    def wrapper(*args) -> Any:
```
LLkaia: function can contain kwargs as well

app/main.py
```diff
@@ -1,6 +1,17 @@
...
+        if args not in cache_dict:
```
LLkaia: use (args, tuple(kwargs.items()), func.__name__) as key in cache_dict because args, kwargs and function name are all identifiers of result

LLkaia (changes requested):

vsmutok (approved):
Looks Good!

PR #1275, opened 2024-08-29
Solution

strauuch:


strauuch:
I can`t use flake8 to check like in Guideline, can u help with this?

app/main.py
```diff
@@ -1,6 +1,18 @@
...
 
 
 def cache(func: Callable) -> Callable:
-    # Write your code here
-    pass
+    cache_storage = {}
+
+    @functools.wraps(func)
+    def wrapper(*args) -> int:
+        if not (args in cache_storage):
```
```suggestion
        if args not in cache_storage:
```

LLkaia (changes requested):
Looks good, but function can have kwargs as well. Also, same arguments can be assigned to different function. 
so use args, kwargs and function name as key in `cache_storage`

strauuch:
I can`t use flake8 to check like in Guideline, can u help with this?

vasylhnatiuk (approved):

PR #1276, opened 2024-08-30
'Solution'

mshmygel:


app/main.py
```diff
@@ -1,6 +1,19 @@
...
+    @functools.wraps(func)
+    def wrapper(*args) -> int:
+        if args in decorator_cache:
+            print("Getting from cache")
+            return decorator_cache[args]
+        else:
+            print("Calculating new result")
+            result = func(*args)
+            decorator_cache[args] = result
+            return result
```
AnyoneClown: Try to rewrite your decorator not to have return duplicated. You can put return after a conditional statement

AnyoneClown (changes requested):

app/main.py
```diff
@@ -1,6 +1,19 @@
...
+
+    @functools.wraps(func)
+    def wrapper(*args) -> int:
+        if args in decorator_cache:
+            print("Getting from cache")
+            return decorator_cache[args]
+        else:
+            print("Calculating new result")
+            result = func(*args)
+            decorator_cache[args] = result
```

app/main.py
```diff
@@ -1,6 +1,18 @@
-from typing import Callable
+from typing import Callable, Any
+import functools
 
 
 def cache(func: Callable) -> Callable:
-    # Write your code here
-    pass
+
+    decorator_cache = {}
```
Nattalli: bad naming, maybe `storage` would be better

Nattalli (approved):
👍🏻

PR #1278, opened 2024-09-01
Solution

olgashydlovska:


app/main.py
```diff
@@ -1,6 +1,41 @@
...
 def cache(func: Callable) -> Callable:
-    # Write your code here
-    pass
+    # Store cache per function
+    func_cache: dict[tuple, any] = {}
+
+    @wraps(func)
+    def wrapper(*args) -> any:
+        # Use the arguments as the cache key
+        cache_key = args
```

app/main.py
```diff
@@ -1,6 +1,41 @@
...
+    return [number ** power for number in n_tuple]
+
+
+# Testing the functions
+long_time_func(1, 2, 3)  # Should calculate new result
+long_time_func(2, 2, 3)  # Should calculate new result
+long_time_func_2((5, 6, 7), 5)  # Should calculate new result
+long_time_func(1, 2, 3)  # Should get from cache
+long_time_func_2((5, 6, 7), 10)  # Should calculate new result
+long_time_func_2((5, 6, 7), 10)  # Should get from cache
```
vsmutok: please remove you tests

vsmutok (changes requested):

app/main.py
```diff
@@ -1,6 +1,32 @@
...
+
+
+@cache
+def long_time_func(number_1: int, number_2: int, number_3: int) -> int:
+    return (number_1 ** number_2 ** number_3) % (number_1 * number_3)
+
+
+@cache
+def long_time_func_2(n_tuple: tuple, power: int) -> list[any]:
+    return [number ** power for number in n_tuple]
```
EdAlekseiev: Delete your test code, please

app/main.py
```diff
@@ -1,6 +1,32 @@
...
+        # Use the arguments as the cache key
+        cache_key = (func.__name__, args, tuple(sorted(kwargs.items())))
+        if cache_key in func_cache:
+            print("Getting from cache")
+            return func_cache[cache_key]
+        else:
+            print("Calculating new result")
+            result = func(*args, **kwargs)
+            func_cache[cache_key] = result
+            return result
```
EdAlekseiev: Rewrite `wrapper` with using one `return`

EdAlekseiev (changes requested):

vasylhnatiuk (approved):

PR #1283, opened 2024-09-05
solution-1

ritoneet:
Everything works according to the pytest

### DIFF
--- a/app/main.py
+++ b/app/main.py
@@ -8,10 +8,9 @@ def cache(func: Callable) -> Callable:
         key = (func, args)
         if key in cache_store:
             print("Getting from cache")
-            return cache_store[key]
         else:
             print("Calculating new result")
             result = func(*args)
             cache_store[key] = result
-            return result
+        return cache_store[key] if key in cache_store else key in cache_store
     return wrapper
--- a/app/main.py
+++ b/app/main.py
@@ -1,16 +1,17 @@
-from typing import Tuple, Any, Callable
+from typing import Tuple, Any, Callable, Dict
+from functools import wraps
 
 
 def cache(func: Callable) -> Callable:
-    cache_store = {}
+    cache_store: Dict[Tuple, Any] = {}
 
-    def wrapper(*args: Tuple) -> Any:
-        key = (func, args)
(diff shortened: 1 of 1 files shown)


vasylhnatiuk (commented):
use only ine return in wrapper

vasylhnatiuk (changes requested):
use only ine return in wrapper

ritoneet:
I didn't notice such requirement in the assignment but now everything should be fine I think

app/main.py
```diff
@@ -1,6 +1,16 @@
...
 
 
 def cache(func: Callable) -> Callable:
-    # Write your code here
-    pass
+    cache_store = {}
+
+    def wrapper(*args: Tuple) -> Any:
+        key = (func, args)
+        if key in cache_store:
```

app/main.py
```diff
@@ -1,6 +1,16 @@
-from typing import Callable
+from typing import Tuple, Any, Callable
 
 
 def cache(func: Callable) -> Callable:
-    # Write your code here
-    pass
+    cache_store = {}
+
+    def wrapper(*args: Tuple) -> Any:
```
vsmutok: use `@wraps`

vsmutok (changes requested):

app/main.py
ritoneet: In general I get it, but what do kwargs take? I have only seen tuple in tests and I don't quite understand why kwargs are needed. Sorry if my questions seem stupid a little but please, can you explain in more details. I can speak Russian and Ukrainian if you are more comfortable with those languages. I don't really know all the rules of the academy so I started asking in English.

vsmutok (approved):
Great!

app/main.py
vsmutok: @ritoneet The issue here isn't what the tests are checking, but that it's necessary to write a universal decorator that will work for different functions that take both args and kwargs.

ritoneet: So it's not critical, just more correct? Anyway, thank you very much!

ritoneet:
Sorry, I just accidentally closed my pull request. There's nothing to check.

PR #1284, opened 2024-09-05
Solution

vladyslav-tmf:


app/main.py
```diff
@@ -2,5 +2,17 @@
...
+    def wrapper(*args) -> Callable:
+
+        if args in cache_storage:
+            print("Getting from cache")
+            return cache_storage[args]
+        else:
+            print("Calculating new result")
+            result = func(*args)
+            cache_storage[args] = result
+            return result
```
AnyoneClown: Use only one return in wrapper

AnyoneClown (changes requested):

app/main.py
```diff
@@ -2,5 +2,17 @@
...
+
+    def wrapper(*args) -> Callable:
+
+        if args in cache_storage:
+            print("Getting from cache")
+            return cache_storage[args]
+        else:
+            print("Calculating new result")
+            result = func(*args)
+            cache_storage[args] = result
```

AnyoneClown (changes requested):
You didn't fix last comment. Key of your cache dictionary must contain 2 values: `arguments`, `result` of the function

app/main.py
```diff
@@ -2,5 +2,18 @@
...
+
+    def wrapper(*args) -> Callable:
+
+        if args in cache_storage:
+            print("Getting from cache")
+            result = cache_storage[args]["result"]
+        else:
+            print("Calculating new result")
+            result = func(*args)
+            cache_storage[args] = {"args": args, "result": result}
```
Nattalli: you shouldn't create a `result` key.
Also try to use a DRY principle: https://github.com/mate-academy/py-cache-decorator/blob/master/checklist.md#dont-repeat-yourself

Nattalli (changes requested):
the main logic of the program is incorrect

AnyoneClown (approved):
Good Job!

PR #1287, opened 2024-09-07
creating cashe

dmytrik:


app/main.py
```diff
@@ -2,5 +2,14 @@
...
+
+    cashe = {}
+
+    def wrapper(*args) -> Callable:
+        if args in cashe:
+            print("Getting from cache")
+            return cashe[args]
+        print("Calculating new result")
+        cashe[args] = func(*args)
+        return cashe[args]
```
AnyoneClown: Use only one return in wrapper

AnyoneClown (changes requested):

app/main.py
```diff
@@ -2,5 +2,14 @@
 
 
 def cache(func: Callable) -> Callable:
-    # Write your code here
-    pass
+
+    cashe = {}
```
AnyoneClown: Use a better name for the variable

app/main.py
```diff
@@ -2,5 +2,14 @@
...
+
+    def wrapper(*args) -> Callable:
+        if args in cashe:
+            print("Getting from cache")
+            return cashe[args]
+        print("Calculating new result")
+        cashe[args] = func(*args)
```

app/main.py
```diff
@@ -2,5 +2,17 @@
 
 
 def cache(func: Callable) -> Callable:
-    # Write your code here
-    pass
+
+    result_storage = {}
```
Nattalli, line 6: btw, simple `storage` variable name would be better :)

Nattalli (approved):

PR #1295, opened 2024-09-17
Solution

honelyq:


viktoria-rybenchuk (changes requested):
All tests should passed

app/main.py
```diff
@@ -1,6 +1,16 @@
-from typing import Callable
+from typing import Callable, Any
 
 
 def cache(func: Callable) -> Callable:
-    # Write your code here
-    pass
+    result_list = {}
```
LLkaia: It is not a list, but dict

app/main.py
```diff
@@ -1,6 +1,16 @@
...
+
+    def wrapper(*args) -> Any:
+        result = args
```
LLkaia: use name of function, args and kwargs cause this what identifies function call

app/main.py
```diff
@@ -1,6 +1,16 @@
-from typing import Callable
+from typing import Callable, Any
 
 
 def cache(func: Callable) -> Callable:
-    # Write your code here
-    pass
+    result_list = {}
+
+    def wrapper(*args) -> Any:
```
LLkaia: Add usage of wraps from functools

app/main.py
LLkaia: Also it is not a result, it is key of cache in a dict

LLkaia (changes requested):
Please, name variables correctly

LLkaia (changes requested):
use name of function, args and kwargs as key in result dict

Dimosphen1 (approved):
LGTM, well done!

PR #1296, opened 2024-09-18
Solution

axe-kkk:


app/main.py
```diff
@@ -1,6 +1,19 @@
...
+    @wraps(func)
+    def wrapper(*args) -> Any:
+        if args in cache_data:
+            print("Getting from cache")
+            return cache_data[args]
+        else:
+            print("Calculating new result")
+            result = func(*args)
+            cache_data[args] = result
+            return result
```
EdAlekseiev: Rewrite this function with using one `return`

EdAlekseiev (changes requested):

axe-kkk:
I solved it

app/main.py
```diff
@@ -1,6 +1,19 @@
...
 
 
 def cache(func: Callable) -> Callable:
-    # Write your code here
-    pass
+    cache_data = {}
+
+    @wraps(func)
+    def wrapper(*args) -> Any:
+        if args in cache_data:
```

vsmutok (changes requested):

app/main.py
```diff
@@ -1,6 +1,21 @@
...
+        cache_key = (func.__name__, args, frozenset(kwargs.items()))
+
+        if cache_key in cache_data:
+            print("Getting from cache")
+            result = cache_data[cache_key]
+        else:
+            print("Calculating new result")
+            result = func(*args, **kwargs)
+            cache_data[cache_key] = result
+        return result
```
```suggestion
        if cache_key in cache_data:
            print("Getting from cache")
        else:
            print("Calculating new result")
            cache_data[cache_key] = func(*args, **kwargs)
        cache_data[cache_key]
```

vsmutok (changes requested):

vasylhnatiuk (approved):

PR #1299, opened 2024-09-20
Cache decorator solution

amylnikova:


NazarSheremet (changes requested):
This solution should work with `**kwargs` too. Create a unique key using args and kwargs

YuriiSmolii (approved):
Well done

PR #1301, opened 2024-09-22
Solution

sashasyrota:


app/main.py
```diff
@@ -1,6 +1,21 @@
-from typing import Callable
+from typing import Any, Callable
+
+data = []
```
Dimosphen1: Please, initialize this variable in the body of the `cache` function before an `inner` function

app/main.py
```diff
@@ -1,6 +1,21 @@
...
 
 
 def cache(func: Callable) -> Callable:
-    # Write your code here
-    pass
+    def inner(*args, **kwargs) -> Any:
+        aray = []
```
Dimosphen1: Use data type that is more efficient than array for storage

app/main.py
```diff
@@ -1,6 +1,21 @@
...
+        aray.append(func)
+        aray.append(args)
```
Dimosphen1: Perform adding data in one operation in case it is absent along with adding a corresponding value to return

app/main.py
```diff
@@ -1,6 +1,21 @@
-from typing import Callable
+from typing import Any, Callable
+
+data = []
 
 
 def cache(func: Callable) -> Callable:
-    # Write your code here
-    pass
+    def inner(*args, **kwargs) -> Any:
```
Dimosphen1, line 11: Use a decorator from `functools` to preserve the wrapped function name and documentation

Dimosphen1 (changes requested):

app/main.py
```diff
@@ -1,6 +1,22 @@
-from typing import Callable
+from typing import Any, Callable
+from functools import wraps
 
 
 def cache(func: Callable) -> Callable:
-    # Write your code here
-    pass
+
+    data = []
```
pavlopro: Rework this to use dictionary instead of list

app/main.py
```diff
@@ -1,6 +1,22 @@
...
+
+    @wraps(func)
+    def inner(*args) -> Any:
```
pavlopro: Please also handle **kwargs

pavlopro (changes requested):

app/main.py
```diff
@@ -1,6 +1,28 @@
...
+        data_type = [args]
+        if func in data and args in data[func]:
+            print("Getting from cache")
+            return data[func][data[func].index(args) + 1]
+
+        elif func not in data:
+            data.update({func: data_type})
+
+        elif args not in data[func]:
+            data[func].append(args)
```

vsmutok (changes requested):

app/main.py
```diff
@@ -1,6 +1,28 @@
-from typing import Callable
+from typing import Any, Callable
+
```
Dimosphen1, line 2: There is no need to place blank line between imports from the same group

app/main.py
```diff
@@ -1,6 +1,28 @@
...
 
 def cache(func: Callable) -> Callable:
-    # Write your code here
-    pass
+
+    data = {}
+
+    @wraps(func)
+    def inner(*args, **kwargs) -> Any:
+        data_type = [args, kwargs]
```
Dimosphen1, line 12: You could use tuple for storing `args` and `kwargs` instead of list

Dimosphen1 (approved):
LGTM, several comments were left for consideration

PR #1303, opened 2024-09-22
Solution

VladyslavBon:


app/main.py
```diff
@@ -2,5 +2,17 @@
...
+    def wrapper(*args) -> Callable:
+        if not all(isinstance(arg, (list, set, dict)) for arg in args):
+            if (func, *args) not in storage:
+                answer = func(*args)
+                storage[(func, *args)] = answer
+                print("Calculating new result")
+                return answer
+            else:
+                print("Getting from cache")
+                return storage[(func, *args)]
```
AnyoneClown: try to rewrite your decorator not to have `return` duplicated

AnyoneClown (changes requested):

app/main.py
```diff
@@ -2,5 +2,17 @@
 
 
 def cache(func: Callable) -> Callable:
-    # Write your code here
-    pass
+    storage = {}
+
+    def wrapper(*args) -> Callable:
+        if not all(isinstance(arg, (list, set, dict)) for arg in args):
```
AnyoneClown: This check is not mandatory, so it can be avoided

AnyoneClown (approved):
Good Job!

PR #1304, opened 2024-09-22
Solution

AchillesPython:


app/main.py
```diff
@@ -1,6 +1,18 @@
...
 
 def cache(func: Callable) -> Callable:
-    # Write your code here
-    pass
+
+    cache_store = {}
+
+    def inner(*args: Any) -> Any:
+        if args in cache_store:
+            print("Getting from cache")
```

app/main.py
```diff
@@ -1,6 +1,18 @@
...
+from typing import Callable, Any
 
 
 def cache(func: Callable) -> Callable:
-    # Write your code here
-    pass
+
+    cache_store = {}
+
+    def inner(*args: Any) -> Any:
```

vsmutok (changes requested):

vsmutok (approved):
Great!

PR #1305, opened 2024-09-23
Solution

omerlenko:


vasylhnatiuk (changes requested):
use only one return in inner

app/main.py
```diff
@@ -1,6 +1,15 @@
-from typing import Callable
+from typing import Callable, Any
 
 
 def cache(func: Callable) -> Callable:
-    # Write your code here
-    pass
+    cache_dict = {}
+
+    def inner(*args) -> Any:
```
LLkaia: Please, use decorator `wraps` imported from functools for `inner` function. Read the purpose of this decorator.

app/main.py
```diff
@@ -1,6 +1,15 @@
...
+        if args in cache_dict:
+            print("Getting from cache")
```
LLkaia: as key for cache_dict usetuple with *args, **kwargs and name of function since all of this will define unique call of function.

LLkaia (changes requested):

app/main.py
```diff
@@ -1,6 +1,18 @@
...
 
 
 def cache(func: Callable) -> Callable:
-    # Write your code here
-    pass
+    cache_dict = {}
+
+    @wraps(func)
+    def inner(*args, **kwargs) -> Any:
+        key = (args, tuple(kwargs), func.__name__)
```
```suggestion
        key = (args, tuple(kwargs.items()), func.__name__)
```

vsmutok (changes requested):

EdAlekseiev (approved):

PR #1306, opened 2024-09-23
Solution

Fantom245:


vasylhnatiuk (changes requested):
use only one return in `inner` func

app/main.py
```diff
@@ -2,5 +2,15 @@
 
 
 def cache(func: Callable) -> Callable:
-    # Write your code here
-    pass
+    cashed_dict = {}
+
+    def inner(*args, **kwargs) -> dict:
```

app/main.py
```diff
@@ -2,5 +2,15 @@
...
+        key = (args, tuple(kwargs.items()))
+        if key in cashed_dict:
+            print("Getting from cache")
+        else:
+            print("Calculating new result")
+            result = func(*args, **kwargs)
+            cashed_dict[key] = result
+        return cashed_dict[key]
```
```suggestion
            print("Calculating new result")
            cashed_dict[key] = func(*args, **kwargs)
        return cashed_dict[key]
```

vsmutok (changes requested):

vasylhnatiuk (approved):

PR #1307, opened 2024-09-23
Solution

nomercy144:


app/main.py
```diff
@@ -1,6 +1,18 @@
...
+from functools import wraps
 
 
 def cache(func: Callable) -> Callable:
-    # Write your code here
-    pass
+    cache = {}
+
+    @wraps(func)
+    def wrapper(*args, **kwargs) -> Callable:
```
LLkaia: It will not return `Callable`

app/main.py
```diff
@@ -1,6 +1,18 @@
...
+        key = (args, frozenset(kwargs.items()))
```
LLkaia: Use name of function as part of `key` tuple since we can face with case when different functions have same arguments

app/main.py
```diff
@@ -1,6 +1,18 @@
...
+        if key in cache:
+            print("Getting from cache")
+            return cache[key]
+        print("Calculating new result")
+        result = func(*args, **kwargs)
+        cache[key] = result
+        return result
```
LLkaia: It is possible to do only one return here

LLkaia (changes requested):

app/main.py
```diff
@@ -1,6 +1,19 @@
...
+    def wrapper(*args, **kwargs) -> Any:
+        key = (func.__name__ , args, frozenset(kwargs.items()))
+        if key in cache_dict:
+            print("Getting from cache")
+            result = cache_dict[key]
+        else:
+            print("Calculating new result")
+            result = func(*args, **kwargs)
+            cache_dict[key] = result
+        return result
```
```suggestion
        if key in cache_dict:
            print("Getting from cache")
        else:
            print("Calculating new result")
            cache_dict[key] = func(*args, **kwargs)
        return cache_dict[key]
```

LLkaia (changes requested):

LLkaia (approved):
Great!

PR #1309, opened 2024-09-24
Develop

YuriiZosia:


app/main.py
```diff
@@ -1,6 +1,22 @@
-from typing import Callable
+from typing import Callable, Any
 
 
 def cache(func: Callable) -> Callable:
-    # Write your code here
-    pass
+    # Словник для зберігання результатів, - наш Кеш,
```
vasylhnatiuk: delete useless comments, use docstring instead of it

vasylhnatiuk (changes requested):

app/main.py
```diff
@@ -1,6 +1,18 @@
...
 
 def cache(func: Callable) -> Callable:
-    # Write your code here
-    pass
+    cache_dict = {}
+
+    def wrapper(*args: Any) -> Any:
+        if args in cache_dict:
+            print("Getting from cache")
+            result = cache_dict[args]
```
vsmutok: 1. use `functools.wraps`
2. you should check both `args` and `kwargs` because the function might accept the same `args` but different `kwargs`.

vsmutok (changes requested):

vasylhnatiuk (approved):

PR #1310, opened 2024-09-25
A solved task has been added

ghost:


app/main.py
```diff
@@ -1,6 +1,37 @@
...
+def long_time_func_2(n_tuple: tuple, power: int) -> list:
+    return [number ** power for number in n_tuple]
+
+
+print(long_time_func(1, 2, 3))
+print(long_time_func(2, 2, 3))
+print(long_time_func_2((5, 6, 7), 5))
+print(long_time_func(1, 2, 3))
+print(long_time_func_2((5, 6, 7), 10))
+print(long_time_func_2((5, 6, 7), 10))
```
viktoria-rybenchuk: This code is redundant. Make sure to remove all code for debugging purposes before pushing to GitHub

app/main.py
```diff
@@ -1,6 +1,37 @@
...
+from functools import wraps
 
 
 def cache(func: Callable) -> Callable:
-    # Write your code here
-    pass
+    cache_data = {}
+
+    @wraps(func)
+    def wrapper(*args) -> None:
```
viktoria-rybenchuk: use kwargs as well

app/main.py
```diff
@@ -1,6 +1,37 @@
...
+        if args in cache_data:
+            print("Getting from cache")
+            return cache_data[args]
+        else:
```
viktoria-rybenchuk: else is redundant

viktoria-rybenchuk (changes requested):

app/main.py
blackbsdl: If I remove the else, then I can't figure out how to ensure that the conditions are met correctly?

pavlopro (approved):
GJ!

PR #1311, opened 2024-09-25
implement cache-decorator

94nj111:


Mxbely (approved):
I think, it is good

app/main.py
```diff
@@ -2,5 +2,14 @@
 
 
 def cache(func: Callable) -> Callable:
-    # Write your code here
-    pass
+    cache_memory = {}
+
+    def inner(*args) -> int:
+        key = tuple([*args])
```
AnyoneClown: Tuple can already take iterable arguments, so we don't need to use [] first for that. Use only tuple

app/main.py
```diff
@@ -2,5 +2,14 @@
...
+        if key in cache_memory:
+            print("Getting from cache")
+            return cache_memory[key]
+        cache_memory[key] = func(*args)
+        print("Calculating new result")
+        return cache_memory[key]
```
AnyoneClown: Refactor using only 1 `return`

AnyoneClown (changes requested):

AnyoneClown (approved):
Good job!

PR #1316, opened 2024-09-27
Decorator solution

molodsh1y1:


app/main.py
```diff
@@ -1,6 +1,18 @@
...
+    def wrapper(*args, **kwargs) -> Any:
+        if args not in cache_storage:
+            print("Calculating new result")
+            function_result = func(*args, **kwargs)
+            cache_storage[args] = function_result
+            return function_result
+        else:
+            print("Getting from cache")
+            return cache_storage.get(args)
+
```
94nj111: I think one return is enough here

app/main.py
```diff
@@ -1,6 +1,18 @@
...
+    wrapper.cache_storage = cache_storage
```
94nj111: looks unnecessary

94nj111 (changes requested):

94nj111 (approved):
lgtm!

AnyoneClown (approved):
Good job!

PR #1318, opened 2024-09-27
Cache-Decorator

Bondzik-S:


YaYaYaroslav (approved):
Everything is correct, but maybe you should add the @wraps decorator?

app/main.py
```diff
@@ -1,6 +1,19 @@
...
+        if args in cache_dict:
+            print("Getting from cache")
+            return cache_dict[args]
+
+        print("Calculating new result")
+        result = func(*args)
+
+        cache_dict[args] = result
+
+        return result
```
AnyoneClown: Refactor it using only 1 `return` inside of your `wrapper`

AnyoneClown (changes requested):

Y-Havryliv (approved):

PR #1321, opened 2024-09-28
solution

YaYaYaroslav:


app/main.py
```diff
@@ -1,6 +1,20 @@
 from typing import Callable
+from functools import wraps
+from typing import Any, Tuple
```
Den-k0: I think you can combine imports from typing in one line, like a "from typing import Callable, Any, Tuple" to make code more shorter

Den-k0 (approved):

app/main.py
```diff
@@ -1,6 +1,20 @@
...
+    @wraps(func)
+    def wrapper(*args: Tuple[Any, ...]) -> Any:
+        if args in func_cache:
+            print("Getting from cache")
+            return func_cache[args]
+        else:
+            print("Calculating new result")
+            result = func(*args)
+            func_cache[args] = result
+            return result
```
AnyoneClown: Refactor your `wrapper` using only 1 `return`

app/main.py
AnyoneClown: Good point, fix it

AnyoneClown (changes requested):

AnyoneClown (changes requested):
You forgot to refactor `wraper`. Check comment above

Y-Havryliv (approved):

PR #1322, opened 2024-09-28
'Solution'

YaroslavBordovoy:


FursykIvan:
In README it says: "Decorator cache creating for decorating only functions that take immutable arguments."
**kwargs are mutable arguments because they are represented as a dict.
And there are no tests that require **kwargs.
And I understood that you are not creating a key for caching named arguments **kwargs.

FursykIvan (changes requested):
In README it says: "Decorator cache creating for decorating only functions that take immutable arguments."
**kwargs are mutable arguments because they are represented as a dict.
And there are no tests that require **kwargs.
And I understood that you are not creating a key for caching named arguments **kwargs.

YaroslavBordovoy:
Removed **kwargs, initially deciding to process everything just in case.

app/main.py
```diff
@@ -2,5 +2,17 @@
...
+    def wrapper(*args) -> int:
+
+        if args not in collected_data:
+            function_result = func(*args)
+            collected_data[args] = function_result
+            print("Calculating new result")
+            return function_result
+
+        print("Getting from cache")
+        return collected_data[args]
```
AnyoneClown: Refactor using only 1 `return` in your `wrapper`

app/main.py
```diff
@@ -2,5 +2,17 @@
 
 
 def cache(func: Callable) -> Callable:
-    # Write your code here
-    pass
+    collected_data = {}
+
+    def wrapper(*args) -> int:
```
AnyoneClown: We should provide type annotation `Any` in this case. Your decorator might return different types(int, float and etc.)

AnyoneClown (changes requested):

Nattalli (approved):
👍🏻

PR #1326, opened 2024-10-01
Add cache decorator solution

charmheroku:


app/main.py
```diff
@@ -1,6 +1,21 @@
...
+from typing import Callable, Any
 
 
 def cache(func: Callable) -> Callable:
-    # Write your code here
-    pass
+    cache_storage = {}
+
+    @wraps(func)
+    def wrapper(*args) -> Any:
```
LLkaia: function can take `**kwars` as well

app/main.py
```diff
@@ -1,6 +1,21 @@
...
+        key = args
```
LLkaia: Use tuple with function name, args and kwargs as key for cache, please

app/main.py
```diff
@@ -1,6 +1,21 @@
...
+
+        if args in cache_storage:
+            print("Getting from cache")
+            return cache_storage[key]
+        else:
+            print("Calculating new result")
+            result = func(*args)
```
LLkaia: no need to create result variable, pass it to the cache dict directly

app/main.py
```diff
@@ -1,6 +1,21 @@
...
+            cache_storage[key] = result
+            return result
```
LLkaia: return `cache_storage[key]` once at the end of the function to avoid calling return twice

LLkaia (changes requested):

pavlopro (approved):
GJ!

PR #1327, opened 2024-10-01
Solution

VolodymyrPovroznyk:


app/main.py
```diff
@@ -1,6 +1,19 @@
...
 
 
 def cache(func: Callable) -> Callable:
-    # Write your code here
-    pass
+    cache_dict = {}
+
+    @functools.wraps(func)
+    def wrapper(*args, **kwargs) -> Any:
+        if args in cache_dict:
```
vsmutok: as key for cache_dict usetuple with *args, **kwargs and name of function since all of this will define unique call of function.

vsmutok (changes requested):

app/main.py
VolodymyrPovroznyk: Done

app/main.py
```diff
@@ -1,6 +1,21 @@
...
+        key = (args, tuple(sorted(kwargs.items())), func.__name__)
+
+        if key in cache_dict:
+            print("Getting from cache")
+            return cache_dict[key]
+        else:
+            print("Calculating new result")
+            result = func(*args, **kwargs)
+            cache_dict[key] = result
+            return result
```
```suggestion
        if key in cache_dict:
            print("Getting from cache")
        else:
            print("Calculating new result")
            cache_dict[key] = func(*args, **kwargs)
        return cache_dict[key]
```

vsmutok (changes requested):

app/main.py
VolodymyrPovroznyk: Understood, thank you. Fixed it.

vasylhnatiuk (approved):

PR #1331, opened 2024-10-03
Cache Decorator Solution

OleksandrSasha:


app/main.py
```diff
@@ -1,6 +1,42 @@
...
+long_time_func(1, 2, 3)
+long_time_func_2((5, 6, 7), 10)
+long_time_func_2((5, 6, 7), 10)
+
+# Calculating new result
+# Calculating new result
+# Calculating new result
+# Getting from cache
+# Calculating new result
+# Getting from cache
```
vsmutok: please remove your custom tests

app/main.py
```diff
@@ -1,6 +1,42 @@
-from typing import Callable
+from typing import Callable, Any
 
 
 def cache(func: Callable) -> Callable:
-    # Write your code here
-    pass
+    res_dict = {}
+
+    def wrapper(*args, **kwargs) -> Any:
```

app/main.py
```diff
@@ -1,6 +1,42 @@
...
+        if args in res_dict:
```
vsmutok: as key for cache_dict usetuple with *args, **kwargs and name of function since all of this will define unique call of function.

vsmutok (changes requested):

app/main.py
```diff
@@ -1,6 +1,20 @@
...
+    def wrapper(*args, **kwargs) -> Any:
+        key = (args, tuple(kwargs.items()), func.__name__,)
+        if key in res_dict:
+            print("Getting from cache")
+            return res_dict[key]
+        else:
+            print("Calculating new result")
+            result = func(*args, **kwargs)
+            res_dict[key] = result
+            return result
```
```suggestion
        if key in res_dict:
            print("Getting from cache")
        else:
            print("Calculating new result")
            res_dict[key] = func(*args, **kwargs)
        return res_dict[key]
```

vsmutok (changes requested):

vsmutok (approved):
Great!

PR #1332, opened 2024-10-04
Create cache decorator

4zik88:


app/main.py
```diff
@@ -1,6 +1,17 @@
-from typing import Callable
+from typing import Any, Callable, Dict, Tuple
 
 
 def cache(func: Callable) -> Callable:
-    # Write your code here
-    pass
+    func_cache: Dict[Tuple, Any] = {}
+
+    def wrapper(*args) -> Any:
```
vsmutok: use functools.wraps

app/main.py
```diff
@@ -1,6 +1,17 @@
...
+        if args in func_cache:
```

vsmutok (changes requested):

app/main.py
```diff
@@ -1,6 +1,20 @@
...
+    def wrapper(*args, **kwargs) -> Any:
+        key = (func.__name__, args, frozenset(kwargs.items()))
+        if key in func_cache:
+            print("Getting from cache")
+            return func_cache[key]
+        else:
+            print("Calculating new result")
+            result = func(*args, **kwargs)
+            func_cache[key] = result
+            return result
```
```suggestion
        key = (func.__name__, args, frozenset(kwargs.items()))
        if key in func_cache:
            print("Getting from cache")
        else:
            print("Calculating new result")
            func_cache[key] = func(*args, **kwargs)
        return func_cache[key]
```

vsmutok (changes requested):

vsmutok (approved):
Good Job!

PR #1333, opened 2024-10-05
'Solution'

AndriiDementiev:


app/main.py
```diff
@@ -1,6 +1,20 @@
...
+    def wrapper(*args, **kwargs) -> Callable:
+        key = (args, tuple(kwargs.items()))
+
+        if key in cache_dict:
+            print("Getting from cache")
+            return cache_dict[key]
+        else:
+            print("Calculating new result")
+            cache_dict[key] = func(*args, **kwargs)
+            return cache_dict[key]
```
YaYaYaroslav: it is better not to use the “else” method here

YaYaYaroslav (approved):

app/main.py
Nattalli: agree, return cache_dict[key] once at the end of the function to avoid calling return twice

Nattalli (changes requested):

app/main.py
```diff
@@ -1,6 +1,19 @@
...
+    def wrapper(*args, **kwargs) -> Callable:
+        key = (args, tuple(kwargs.items()))
+
+        if key in cache_dict:
+            print("Getting from cache")
+            return cache_dict[key]
+
+        print("Calculating new result")
+        cache_dict[key] = func(*args, **kwargs)
+        return cache_dict[key]
```
Nattalli: you still have 1 return statements with the same returned value. so you can add `if else` statements but use onle one `return` statement

```suggestion

        if key in cache_dict:
            print("Getting from cache")
        else:
            print("Calculating new result")
            cache_dict[key] = func(*args, **kwargs)
        return cache_dict[key]
```

Nattalli (changes requested):

Nattalli (approved):

PR #1335, opened 2024-10-06
Solution

sumseer:


Rabb1t-Artem (approved):
GJ:)

app/main.py
```diff
@@ -2,5 +2,17 @@
...
+
+    def wrapper(*args, **kwargs) -> None:
+        cache_key = (args, tuple(sorted(kwargs.items())))
+
+        if cache_key in cache_results:
+            print("Getting from cache")
+            return cache_results.get(cache_key)
+        print("Calculating new result")
+        result = func(*args, **kwargs)
+        cache_results[cache_key] = result
```
Nattalli: don't create `result` variable, it's unnecessary

app/main.py
```diff
@@ -2,5 +2,17 @@
...
+        return result
```
Nattalli: you can call `return` statement at once

Nattalli (changes requested):

Nattalli (approved):

PR #1338, opened 2024-10-09
Solution

KhrypunovMaxim:


app/main.py
```diff
@@ -2,5 +2,16 @@
 
 
 def cache(func: Callable) -> Callable:
-    # Write your code here
-    pass
+
+    store = {}
+
+    def wrapped(*args) -> Callable:
```
viktoria-rybenchuk: you should add kwargs as well

app/main.py
```diff
@@ -2,5 +2,16 @@
...
+        if args in store:
+            print("Getting from cache")
+            return store[args]
+        else:
```
viktoria-rybenchuk, line 15: else is redundant

viktoria-rybenchuk (changes requested):

app/main.py
```diff
@@ -1,6 +1,22 @@
...
 from typing import Callable
 
 
 def cache(func: Callable) -> Callable:
-    # Write your code here
-    pass
+    store = {}
+
+    @wraps(func)
+    def wrapped(*args, **kwargs) -> dict:
```
pavlopro: Correct return annotation. It should be `Any`

app/main.py
```diff
@@ -1,6 +1,22 @@
...
+
+        key = (args, frozenset(kwargs.items()))
```
pavlopro: include function name as well

pavlopro (changes requested):

app/main.py
KhrypunovMaxim: key = (func.__name__, args, frozenset(kwargs.items())) like this?

app/main.py
```diff
@@ -1,6 +1,22 @@
...
+    def wrapped(*args, **kwargs) -> Any:
+
+        key = (func.__name__, args, frozenset(kwargs.items()))
+
+        if key in store:
+            print("Getting from cache")
+            return store[key]
+
+        print("Calculating new result")
+        result = func(*args, **kwargs)
```
LLkaia: no reason to create `result`, assign it to `store[key]` directly

app/main.py
```diff
@@ -1,6 +1,22 @@
...
+        store[key] = result
+        return result
```
LLkaia: And return `store[key]`

LLkaia: And then because you will have two returns of `store[key]`, change logic to return `store[key]` only once

LLkaia (changes requested):

LLkaia (approved):
AWESOME!

PR #1339, opened 2024-10-10
Solution

serg121192:


app/main.py
```diff
@@ -2,5 +2,19 @@
 
 
 def cache(func: Callable) -> Callable:
-    # Write your code here
-    pass
+    cached_results = {}
+
+    def inner(*args) -> str:
```
Dimosphen1: Please, indicate that return type could be `Any`, not only string

Dimosphen1: Use a decorator from `functools` to preserve the wrapped function name and docstring

Dimosphen1: Please, process `kwargs` as well, not only `args`

app/main.py
```diff
@@ -2,5 +2,19 @@
...
+        if func.__name__ not in cached_results:
+            print("Calculating new result")
+            cached_results[func.__name__] = {args: func(*args)}
+        else:
+            if args in cached_results[func.__name__]:
+                print("Getting from cache")
+            else:
+                print("Calculating new result")
+                cached_results[func.__name__][args] = func(*args)
```
Dimosphen1: Try to avoid using nested if conditions, one-level if else block should be enough

Dimosphen1 (changes requested):

app/main.py
```diff
@@ -1,6 +1,20 @@
...
 
 
 def cache(func: Callable) -> Callable:
-    # Write your code here
-    pass
+    cached_results = {}
+
+    @functools.wraps(func)
+    def inner(*args, **kwargs) -> Any:
+        if func.__name__ not in cached_results:
```
pavlopro: Please combine func.__name__, args and kwargs into one key

pavlopro (changes requested):

pavlopro (approved):
GJ!

PR #1349, opened 2024-10-17
Solution

maximprysyazhnikov:


Alkaponees (changes requested):
Looks like code write by using ChatGPT. Recommend to use AI tools for checking example code not as solution

maximprysyazhnikov:
> Looks like code write by using ChatGPT. Recommend to use AI tools for checking example code not as solution

yes, I used it to check and refactor the code to make it more readable, did I complete the task? if something doesn't work in relation to the terms of reference or I did something wrong, please show me what exactly I should redo

Alkaponees (approved):
LGMT!

PR #1357, opened 2024-10-21
Completing the task

Greena1c:


NazarSheremet (changes requested):
This solution should work with `**kwargs` too. Create a unique key which will contains `args `and `kwargs`

Greena1c:
You can check one more time

Alkaponees (approved):
LGMT!

PR #1358, opened 2024-10-22
Solution

dottfmar:


app/main.py
```diff
@@ -2,5 +2,13 @@
 
 
 def cache(func: Callable) -> Callable:
-    # Write your code here
-    pass
+    cache_saved_results = {}
+
+    def wrapper(*args) -> int:
```
petrykivd: Inner can return not only integer, it's better to use Any

app/main.py
```diff
@@ -2,5 +2,13 @@
...
+        if args in cache_saved_results:
+            print("Getting from cache")
+            return cache_saved_results[args]
+        print("Calculating new result")
+        cache_saved_results[args] = func(*args)
+        return cache_saved_results[args]
```
petrykivd: Avoid using additional variable, it is not needed here

petrykivd (changes requested):

app/main.py
```diff
@@ -2,5 +2,12 @@
...
-    # Write your code here
-    pass
+    cache_saved_results = {}
+
+    def wrapper(*args) -> any:
+        if args in cache_saved_results:
+            print("Getting from cache")
+            return cache_saved_results[args]
+        print("Calculating new result")
+        return cache_saved_results.setdefault(args, func(*args))
```
L1nk3rrr: try to rewrite your decorator not to have `return` duplicated

L1nk3rrr (changes requested):

L1nk3rrr (approved):
Well done!

PR #1381, opened 2024-11-01
cache function implemented

Beliar24:


app/main.py
```diff
@@ -1,6 +1,18 @@
...
 from typing import Callable
 
 
 def cache(func: Callable) -> Callable:
-    # Write your code here
-    pass
+    cache_dict = {}
+
+    @wraps(func)
+    def wrapper(*args, **kwargs) -> Callable:
```
Alkaponees: The return type annotation for the `wrapper` function is incorrect. It should match the return type of the `func` being decorated, not `Callable`. Consider changing it to `Any` or removing the return type annotation.

Alkaponees (changes requested):
Please check my review

Alkaponees (changes requested):
Check my review

Alkaponees (approved):
Awesome job!

PR #1387, opened 2024-11-05
'solution'

julia4406:


app/main.py
```diff
@@ -2,5 +2,16 @@
...
-    # Write your code here
-    pass
+    cached_data = {}
+
+    def wrapper(*args, **kwargs) -> Callable:
+        parameters = tuple(args)
+        if parameters not in cached_data.keys():
+            result = func(*args, **kwargs)
+            cached_data[parameters] = result
+            print("Calculating new result")
```
petrykivd: Avoid using additional variable, it is not needed here

app/main.py
```diff
@@ -2,5 +2,16 @@
 
 
 def cache(func: Callable) -> Callable:
-    # Write your code here
-    pass
+    cached_data = {}
+
+    def wrapper(*args, **kwargs) -> Callable:
+        parameters = tuple(args)
```
petrykivd: Are you sure you can't do without tuple?

petrykivd (changes requested):

app/main.py
```diff
@@ -1,6 +1,14 @@
...
-    pass
+    storage = {}
+
+    def wrapper(*args) -> Callable:
+        if args not in storage:
+            print("Calculating new result")
+            return storage.setdefault(args, func(*args))
+        else:
+            print("Getting from cache")
+            return storage.get(args)
```
L1nk3rrr: Try to rewrite wrapper not using a duplicated `return`, without duplicating it

app/main.py
```diff
@@ -1,6 +1,14 @@
...
+    return wrapper
```
L1nk3rrr, line 15: Always add a blank line at the end of the file.

app/main.py
```diff
@@ -1,6 +1,14 @@
...
+from collections.abc import Callable
 
 
 def cache(func: Callable) -> Callable:
-    # Write your code here
-    pass
+    storage = {}
+
+    def wrapper(*args) -> Callable:
+        if args not in storage:
```
L1nk3rrr: In many cases , it is better not to use `not`, because we always thinking if something `is in` but not `is not in`. But a it's a minor thing, that can save you a 3 symbols xD
By the way, it's opinion from my side for this task, sometimes we really need to check `not in` because it looks better with variable names and task requirements.
At that task, if you swap you `if` code and `else` to avoid using not, nothing will change logically!

L1nk3rrr (changes requested):

app/main.py
julia4406: Yeh, sure, but it is there (flake8 always push me to do this)

app/main.py
```diff
@@ -1,6 +1,15 @@
...
-    # Write your code here
-    pass
+    storage = {}
+
+    def wrapper(*args) -> Callable:
+        if args in storage:
+            print("Getting from cache")
+        else:
+            print("Calculating new result")
+            storage.setdefault(args, func(*args))
```
L1nk3rrr, line 12: There is no more sense use `setdefault` method, because you are getting value at the end, we can simply set value by a key. Now it's doing one extra move - return you a value into nothing, so we don't need it.

L1nk3rrr (commented):
That was your last try, so review my comment and accept it. Anyway you are doing great, that was nice to see that you used a `setdefault` method (for previous solution it makes sense!!), remember how it works and use it next time if it's only needed.
Well done!

L1nk3rrr (approved):

PR #1393, opened 2024-11-10
Solution

RuslanKravchuk2907:


app/main.py
```diff
@@ -1,6 +1,38 @@
...
+    # Simulating a different long-running computation
+    return single_number * 2
+
+
+# Testing
+print(long_running_function(1, 2))
+print(long_running_function(1, 2))
+print(long_running_function(2, 3))
+print(another_function(3))
+print(another_function(3))
```
L1nk3rrr, line 38: Dont push such changes

app/main.py
```diff
@@ -1,6 +1,38 @@
...
+
+    def wrapper(*args) -> any:
+        if args in cache_storage:
+            print("Getting from cache")
+            return cache_storage[args]
+        else:
+            print("Calculating new result")
+            result = func(*args)
+            cache_storage[args] = result
+            return result
```
L1nk3rrr, line 15: Try to use only 1 return

app/main.py
```diff
@@ -1,6 +1,38 @@
-from typing import Callable
+from typing import Callable, Dict, Tuple
 
 
 def cache(func: Callable) -> Callable:
-    # Write your code here
-    pass
+    cache_storage: Dict[Tuple, any] = {}
```
L1nk3rrr, line 5: Any must be imported as type, `any` is a bult-in construction, not a type

app/main.py
```diff
@@ -1,6 +1,38 @@
...
+
+    def wrapper(*args) -> any:
```
L1nk3rrr, line 7: same about any

app/main.py
```diff
@@ -1,6 +1,38 @@
-from typing import Callable
+from typing import Callable, Dict, Tuple
```
L1nk3rrr, line 1: You already have dict and tuple in python and can use it for annotation

L1nk3rrr (changes requested):

PR #1396, opened 2024-11-12
solution

VovaGld:


julia4406 (commented):
Everything looks great. In my opinion your solution is simple and correct )

julia4406 (approved):

PR #1398, opened 2024-11-13
Solution to py-cache-decorator task

Trynadis:
Written decorator cache that stores results of completed runs with different arguments, number of arguments can be also different. If decorated function runs with repeating arguments it should return stored result instead of calling function again. Decorator cache creating for decorating only functions that take immutable arguments.

### DIFF
--- a/app/main.py
+++ b/app/main.py
@@ -1,6 +1,22 @@
 from typing import Callable
+import functools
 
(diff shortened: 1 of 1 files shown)


Alkaponees (approved):
LGMT!

PR #1399, opened 2024-11-13
'Solution'

Dizatvarska:


app/main.py
```diff
@@ -1,6 +1,16 @@
...
+    def wrapper(*args, **kwargs) -> Any:
+        def inner(func: Callable) -> Any:
+            results = []
+            for _ in range(numero):
+                results.append(func(*args, **kwargs))
+                for result in results:
+                    if results.count(result) > 1:
+                        print(f"Getting from cache {result}")
+                    else:
+                        print(f"Calculating new result {result}")
```
vsmutok: try to solve this task without using loops

vsmutok (changes requested):
Tests Failed!

vsmutok (changes requested):
Tests Failed! 
Before submitting the task for review, ensure that your code passes all tests.

app/main.py
```diff
@@ -1,6 +1,16 @@
...
+from typing import Callable, Any
 
 
-def cache(func: Callable) -> Callable:
-    # Write your code here
-    pass
+def cache(func: Callable) -> Any:
+    results = {}
+
+    def wrapper(*args) -> Any:
```
vsmutok: what if the function will also accept key arguments?
you should also consider this point

vsmutok (changes requested):

app/main.py
Dizatvarska: Per task conditions, the functions take only immutable arguments. Am I correct, or is the point of the question in anything else?

app/main.py
```diff
@@ -1,6 +1,16 @@
...
 
 
-def cache(func: Callable) -> Callable:
-    # Write your code here
-    pass
+def cache(func: Callable) -> Any:
+    results = {}
+
+    def wrapper(*args, **kwargs) -> Any:
+        if args in results:
```
vsmutok: You should create a key based on both keyword and positional arguments, rather than using only `args` as the key. This is because the function might accept the same `args` but different `kwargs`.

vsmutok (changes requested):

app/main.py
```diff
@@ -1,6 +1,17 @@
...
-    pass
+def cache(func: Callable) -> Any:
+    results = {}
+
+    def wrapper(*args, **kwargs) -> Any:
+        key = tuple(args, *kwargs.items())
+        if key in results:
+            print("Getting from cache")
+            return results[key]
+        else:
```
pavlopro: `else` operator is redundant, you can remove it

pavlopro (changes requested):

app/main.py
```diff
@@ -1,6 +1,17 @@
...
-def cache(func: Callable) -> Callable:
-    # Write your code here
-    pass
+def cache(func: Callable) -> Any:
+    results = {}
+
+    def wrapper(*args, **kwargs) -> Any:
+        key = tuple(args)
+        key += tuple(*kwargs.items())
+        if key in results:
```
vsmutok: try to write it in one line

app/main.py
```diff
@@ -1,6 +1,17 @@
...
+from typing import Callable, Any
 
 
-def cache(func: Callable) -> Callable:
-    # Write your code here
-    pass
+def cache(func: Callable) -> Any:
+    results = {}
+
+    def wrapper(*args, **kwargs) -> Any:
```
vsmutok: use `wraps`

vsmutok (changes requested):

app/main.py
```diff
@@ -1,6 +1,18 @@
...
 
 
-def cache(func: Callable) -> Callable:
-    # Write your code here
-    pass
+def cache(func: Callable) -> Any:
+    results = {}
+    wraps(func)
+
+    def wrapper(*args, **kwargs) -> Any:
```
pavlopro, line 9:
```
@wraps(func)
def wrapper(....
```

pavlopro (changes requested):

pavlopro (changes requested):
Tests failed

vsmutok (approved):
GJ!

PR #1408, opened 2024-11-18
solution to py-cache-decorator

Annolel:
Checks by pytest and flake8 were passed successfully.

### DIFF
--- a/app/main.py
+++ b/app/main.py
@@ -2,5 +2,16 @@ from typing import Callable
 
 
 def cache(func: Callable) -> Callable:
-    # Write your code here
-    pass
+
+    tmp_cache = {}
+
+    def inner(*args) -> None:
+        if args in tmp_cache:
+            print("Getting from cache")
+            return tmp_cache[args]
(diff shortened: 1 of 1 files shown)


app/main.py
```diff
@@ -2,5 +2,16 @@
 
 
 def cache(func: Callable) -> Callable:
-    # Write your code here
-    pass
+
+    tmp_cache = {}
+
+    def inner(*args) -> None:
```
Alkaponees: Recommend importing from typing Any also. Use it for this function.

Alkaponees (changes requested):
Check my review

app/main.py
Annolel: Why should I even import this "any" thing? I can just make it like:
def inner(*args): 
Isn't it the same?

Annolel:
"Any" type was imported and added to the decorator function to describe it's output type correctly.

Alkaponees (approved):
LGMT!

PR #1415, opened 2024-11-22
first solution

Mykola-Osolinskyi:


app/main.py
```diff
@@ -2,5 +2,17 @@
 
 
 def cache(func: Callable) -> Callable:
-    # Write your code here
-    pass
+    history = {}
+
+    def wrapper(*args) -> Callable:
+        if isinstance(args, (int, float, str, tuple)):
```
Nattalli: not neccessary

app/main.py
```diff
@@ -2,5 +2,17 @@
...
+            if args in history:
+                print("Getting from cache")
+                return history[args]
+            else:
+                result = func(*args)
+                print("Calculating new result")
+                history[args] = result
+                return result
```
Nattalli: you can use only one return after `if else` statement

Nattalli (changes requested):

app/main.py
Mykola-Osolinskyi: don't get this point
could you please clarify it?

Nattalli (approved):
looks good to me!

PR #1427, opened 2024-11-28
Create cache decorator

ihorhalyskiy:


4ikkamony (changes requested):
Nice job!
But a few things to note: two return statements are unnecessary—you can go with just one. Like this:

```python
if cache_data_check in cached_data:
    print("Getting from cache")
else:
    result = func(*args)
    cached_data.update({cache_data_check: result})
return cached_data[cache_data_check]
```
This way, when you reach the return statement, it's guaranteed that the key you want to store is there, because it was either already there, or you just added it if it wasn't.

It's a minor detail, but good to know)

Also, you have a few typos:

Not "cash" but "cache".
Not "cashed" but "cached." Cash is money)
Oh, and your wrapper doesn't actually return None, as you specified in the type hint. It returns whatever the decorated function returns. This is why you should use typing.Any as the return type. Like this:
```python
...
def wrapper(*args) -> Any:
    ...
```
Otherwise, as I said, great work! Fix this minor stuff, and it'll be outright perfect)

4ikkamony (approved):
Nice

4ikkamony (approved):
Even nicer

app/main.py
```diff
@@ -2,5 +2,17 @@
 
 
 def cache(func: Callable) -> Callable:
-    # Write your code here
-    pass
+    cached_data = {}
+
+    def wrapper(*args) -> None:
+
+        cache_data_check = args
```
L1nk3rrr: You don't need a variable here

app/main.py
```diff
@@ -2,5 +2,17 @@
 
 
 def cache(func: Callable) -> Callable:
-    # Write your code here
-    pass
+    cached_data = {}
+
+    def wrapper(*args) -> None:
```
L1nk3rrr: It looks that you can return not only None

app/main.py
```diff
@@ -2,5 +2,17 @@
...
+
+        cache_data_check = args
+
+        if cache_data_check in cached_data:
+            print("Getting from cache")
+        else:
+            result = func(*args)
+            cached_data.update({cache_data_check: result})
```
L1nk3rrr: Try to write it into one line, and think about do you need an update method here, or it can be simplified

L1nk3rrr (changes requested):
Check my comments to improve your code!

app/main.py
```diff
@@ -2,5 +2,16 @@
 
 
 def cache(func: Callable) -> Callable:
-    # Write your code here
-    pass
+    cached_data = {}
+
+    def wrapper(*args) -> Callable:
```
L1nk3rrr: Wrapper return value from function that was calculated, looks like Any type, not Callable here

app/main.py
```diff
@@ -2,5 +2,16 @@
...
+
+        if args in cached_data:
+            print("Getting from cache")
+            return cached_data[args]
+        else:
+            print("Calculating new result")
+            cached_data[args] = func(*args)
+            return cached_data[args]
```
L1nk3rrr: double return looks bad, maybe we try can to use only return one here?

L1nk3rrr (changes requested):

petrykivd (approved):
Good job! 😎

PR #1432, opened 2024-11-29
Added decorator cache

DanilTrukhanov:
Added decorator cache that stores repeated arguments and a result of a function it wraps to prevent function calling with the same arguments.


Dogober (approved):
Grate job!

PR #1433, opened 2024-11-29
'Solution'

Dogober:


app/main.py
```diff
@@ -2,5 +2,15 @@
 
 
 def cache(func: Callable) -> Callable:
-    # Write your code here
-    pass
+    cache_data = {}
+
+    def wrapper(*args) -> Callable:
+        if args in cache_data.keys():
```
L1nk3rrr: Looks like calling `keys()` not needed, as we already checking at dict keys

app/main.py
```diff
@@ -2,5 +2,15 @@
 
 
 def cache(func: Callable) -> Callable:
-    # Write your code here
-    pass
+    cache_data = {}
+
+    def wrapper(*args) -> Callable:
```
L1nk3rrr: You can return not only callable, because your functions can return any type

app/main.py
```diff
@@ -2,5 +2,15 @@
...
+        if args in cache_data.keys():
+            print("Getting from cache")
+            return cache_data[args]
+
+        cache_data[args] = func(*args)
+        print("Calculating new result")
+        return cache_data[args]
```
L1nk3rrr: You can try to simplify your solution to use only one `return`

L1nk3rrr (changes requested):

app/main.py
```diff
@@ -1,6 +1,15 @@
-from typing import Callable
+from typing import Any
 
 
-def cache(func: Callable) -> Callable:
-    # Write your code here
-    pass
+def cache(func: Any) -> Any:
```
L1nk3rrr: `cache` accept and return only callable, no need to change this

L1nk3rrr (changes requested):
Well done! Minor point about what decorator accept

app/main.py
Dogober: Many thanks for your correction. I got it. I understood why exactly so and for what

petrykivd (approved):
Nice! 😎

PR #1442, opened 2024-12-02
solution

10kkyvl:


app/main.py
```diff
@@ -1,6 +1,21 @@
...
+    def wrapper(*args, **kwargs) -> Callable:
+        key = (args, tuple(kwargs.items()))
+        if key in cached:
+            print("Getting from cache")
+            return cached[key]
+
+        print("Calculating new result")
+        result = func(*args, **kwargs)
+        cached[key] = result
+        return result
```
L1nk3rrr: try to rewrite your decorator not to have `return` duplicated

L1nk3rrr (changes requested):

app/main.py
10kkyvl: got it, commited fix

L1nk3rrr (approved):
Well done!

PR #1457, opened 2024-12-17
Solution

bogdAAAn1:


app/main.py
```diff
@@ -2,5 +2,14 @@
 
 
 def cache(func: Callable) -> Callable:
-    # Write your code here
-    pass
+    cached_data = {}
+
+    def wrapper(*args) -> None:
```
petrykivd: wrapper return None  in all cases?

petrykivd (changes requested):

app/main.py
```diff
@@ -2,5 +2,14 @@
 
 
 def cache(func: Callable) -> Callable:
-    # Write your code here
-    pass
+    cached_data = {}
+
+    def wrapper(*args) -> Callable:
```
L1nk3rrr: Wrapper can return Callable but not always. Think about what is can be stored at your dict, you can use this decorator with any function (that functions can return int, float, list, dict, tuple, set, etc.), and you return result of that functions

L1nk3rrr (changes requested):

L1nk3rrr (approved):
Well done!

PR #1466, opened 2024-12-21
Solution

kuinorymes:


app/main.py
```diff
@@ -2,5 +2,16 @@
...
+    def inner(*args, **kwargs) -> func:
+        key = (args, tuple(sorted(kwargs.items())))
+        if key not in results:
+            print("Calculating new result")
+            result = func(*args, **kwargs)
+            results[key] = result
+            return result
+        else:
+            print("Getting from cache")
+            return results[key]
```
L1nk3rrr: I think you can use here one return, or use 2 returns but without else

app/main.py
```diff
@@ -2,5 +2,16 @@
 
 
 def cache(func: Callable) -> Callable:
-    # Write your code here
-    pass
+    results = {}
+
+    def inner(*args, **kwargs) -> func:
```
L1nk3rrr: it's not returning func, it's return any result of decorated function

L1nk3rrr: also inner must be decorated by @wraps from functools

L1nk3rrr (changes requested):
Nice work, left some points for improvement!

L1nk3rrr (approved):
Well done!

PR #1471, opened 2024-12-26
Solution

vasyl-main-dev:


app/main.py
```diff
@@ -1,6 +1,14 @@
-from typing import Callable
+from typing import Callable, Any
 
 
 def cache(func: Callable) -> Callable:
-    # Write your code here
-    pass
+    results = {}
+
+    def wrapper(*args) -> Any:
```
vsmutok, line 7: please use `@functools.wraps`

app/main.py
```diff
@@ -1,6 +1,14 @@
...
+        if args in results:
+            print("Getting from cache")
+            return results[args]
```
vsmutok, line 10: You should check both `args` and `kwargs` because the function might accept the same `args` but different `kwargs`. Additionally, I recommend checking the function name since there could be two different functions that accept the same arguments.

vsmutok (changes requested):

PR #1476, opened 2024-12-28
Solution

Khomychh:


app/main.py
```diff
@@ -1,6 +1,19 @@
...
+from typing import Callable, Any
 
 
 def cache(func: Callable) -> Callable:
-    # Write your code here
-    pass
+    cache_history = {}
+
+    def wrapper(*args: Any) -> Callable:
+        nonlocal cache_history
```
vsmutok: `nonlocal cache_history` redundant here

app/main.py
```diff
@@ -1,6 +1,19 @@
-from typing import Callable
+from typing import Callable, Any
 
 
 def cache(func: Callable) -> Callable:
-    # Write your code here
-    pass
+    cache_history = {}
+
+    def wrapper(*args: Any) -> Callable:
```
vsmutok: please use `@functools.wraps`

app/main.py
```diff
@@ -1,6 +1,19 @@
...
+
+        if cache_key in cache_history:
+            print("Getting from cache")
+            return cache_history[cache_key]
+        else:
+            print("Calculating new result")
+            result = func(*args)
+            cache_history[cache_key] = result
+            return result
+    return wrapper
```

vsmutok (changes requested):

app/main.py
Khomychh: I fixed the errors you pointed out

app/main.py
```diff
@@ -1,6 +1,20 @@
...
 
 
 def cache(func: Callable) -> Callable:
-    # Write your code here
-    pass
+    cache_history = {}
+
+    @functools.wraps(func)
+    def wrapper(*args, **kwargs) -> Callable:
+        cache_key = args
```
pavlopro: include kwargs into cache_key as well

app/main.py
```diff
@@ -1,6 +1,20 @@
...
+
+        if cache_key in cache_history:
+            print("Getting from cache")
+            return cache_history[cache_key]
+        else:
```
pavlopro: `else` operator is redundant

pavlopro (changes requested):

pavlopro (approved):
GJ

PR #1478, opened 2025-01-01
'Solution'

Nikita-pyth:


app/main.py
```diff
@@ -2,5 +2,16 @@
...
+
+    def wrapper(*args) -> int:
+        if args in cached_values.keys():
+            print("Getting from cache")
+            return cached_values[args]
+        else:
+            print("Calculating new result")
+            result = func(*args)
+            cached_values[args] = result
+            return result
```

app/main.py
```diff
@@ -2,5 +2,16 @@
...
+
+    return wrapper
```
oliasymak24, line 16: Always add a blank line at the end of the file.

app/main.py
```diff
@@ -2,5 +2,16 @@
 
 
 def cache(func: Callable) -> Callable:
-    # Write your code here
-    pass
+    cached_values = {}
+
+    def wrapper(*args) -> int:
+        if args in cached_values.keys():
```
oliasymak24: no need to use keys() method here, it will search through keys by default

oliasymak24 (changes requested):
Pay attention to some comments)

app/main.py
Nikita-pyth: hi, i'm adding a blank line to the end of the file, however on github it isn't displayed. what should i do?

oliasymak24 (approved):

PR #1479, opened 2025-01-03
feat(cache): implement caching decorator with test support

kram3ko:


app/main.py
```diff
@@ -1,6 +1,20 @@
...
+    def wrapper(*args, **kwargs) -> Callable:
+        key = (args, tuple(kwargs.items()))
+        if key in cash_store:
+            print("Getting from cache")
+            return cash_store[key]
+        else:
+            result = func(*args, **kwargs)
+            cash_store[key] = result
+            print("Calculating new result")
+            return result
```
oliasymak24: You can use a single return statement out of the if/else statement instead of two inside it

app/main.py
```diff
@@ -1,6 +1,20 @@
...
+
+    return wrapper
```
oliasymak24, line 20: Always add a blank line at the end of the file.

oliasymak24 (changes requested):

app/main.py
kram3ko: added

app/main.py
kram3ko: > You can use a single return statement out of the if/else statement instead of two inside it

Done it ;)

kram3ko (commented):
refactored code

oliasymak24 (approved):

PR #1482, opened 2025-01-05
Solution

Zikkolion:


app/main.py
```diff
@@ -1,6 +1,19 @@
...
+from typing import Callable, Any
 
 
 def cache(func: Callable) -> Callable:
-    # Write your code here
-    pass
+    cache_dict = {}
+
+    def inner(*args, **kwargs) -> Any:
+        func_name = f"{func.__name__}"
```
vsmutok: Using an f-string in this case, doesn't make sense because `func.__name__` is already a string. 
```suggestion
        func_name = func.__name__
```

app/main.py
```diff
@@ -1,6 +1,19 @@
-from typing import Callable
+from typing import Callable, Any
 
 
 def cache(func: Callable) -> Callable:
-    # Write your code here
-    pass
+    cache_dict = {}
+
+    def inner(*args, **kwargs) -> Any:
```
vsmutok, line 9: Don't forget to use `functools.wraps`.

app/main.py
```diff
@@ -1,6 +1,19 @@
...
+        func_name = f"{func.__name__}"
+
+        if func_name not in cache_dict:
+            cache_dict[func_name] = {}
+        if args not in cache_dict[func_name]:
+            cache_dict[func_name][args] = func(*args, **kwargs)
```
vsmutok: Here, you should use not only `args` as the key. The function might have the same `args` but different `kwargs`. In such cases, these should be treated as two separate values.

vsmutok (changes requested):

PR #1517, opened 2025-02-02
Solution

Baranotik15:


app/main.py
```diff
@@ -2,5 +2,19 @@
...
+        key = (args, frozenset(kwargs.items())) if kwargs else args
+
+        if key in storage:
+            print("Getting from cache")
+            return storage[key]
+        else:
+            print("Calculating new result")
+            result = func(*args, **kwargs)
+            storage[key] = result
+            return result
```

app/main.py
```diff
@@ -2,5 +2,19 @@
 
 
 def cache(func: Callable) -> Callable:
-    # Write your code here
-    pass
+    storage = {}
+
+    def wrapper(*args, **kwargs) -> int:
```
oliasymak24: Add an annotation

oliasymak24 (changes requested):

app/main.py
```diff
@@ -1,6 +1,19 @@
...
+
+        key = (args, frozenset(kwargs.items())) if kwargs else args
+
+        if key in storage:
+            print("Getting from cache")
+            return storage[key]
+        else:
+            print("Calculating new result")
+            storage[key] = func(*args, **kwargs)
+            return storage[key]
```
oliasymak24: You still didn't change the code by me comment

> You can use a single return statement out of the if/else statement instead of two inside it.

oliasymak24 (changes requested):

app/main.py
TopalRuslan: I understand, but for me in such long one-line solutions the readability is worse, is it common practice to write them in 1 line?

app/main.py
```diff
@@ -1,6 +1,16 @@
...
+    storage = {}
+
+    def wrapper(*args: Any, **kwargs: Any) -> int:
+
+        key = (args, frozenset(kwargs.items())) if kwargs else args
+
+        print("Getting from cache" if key in storage
+              else "Calculating new result")
+        return storage[key] if key in storage \
+            else storage.setdefault(key, func(*args, **kwargs))
```
oliasymak24: You missunderstood my comment
What I meant 
`if key in storage:`
    ` print("Getting from cache")`
`else:`
   ` print("Calculating new result")`
   ` storage[key] = func(*args, **kwargs)`

`return storage[key]`

I didn't mentioned to write it in one line

oliasymak24 (changes requested):

Y-Havryliv (approved):
gj!

PR #1543, opened 2025-02-14
Solution

iSevenpwnz:


app/main.py
```diff
@@ -1,6 +1,18 @@
...
-    # Write your code here
-    pass
+    cache_data = {}
+
+    def wrapper(*args, **kwargs) -> Any:
+        key = (args, tuple(sorted(kwargs.items())))
+        if key in cache_data:
+            print("Getting from cache")
+            return cache_data[key]
+        else:
```
Yurkiv313, line 12: This is just an alternative way of writing, not a mistake. You can avoid using (else) because if the code doesn’t enter the (if), it will continue to the next lines anyway.

Yurkiv313 (approved):
This is just an alternative way of writing, not a mistake. You can avoid using (else) because if the code doesn’t enter the (if), it will continue to the next lines anyway.

app/main.py
iSevenpwnz: Thanks for the advice

PR #1546, opened 2025-02-16
Solution 'Cache_decorator'

Kart0l:


PrimeGlorious (changes requested):
The code works correctly, but these functions `long_time_func` and `long_time_func_2` are not needed and should be removed from the code, as they are causing warnings in pytest.

Kart0l:
Thank you for your comment.
Functions for which the remark was deleted

PR #1547, opened 2025-02-16
Implement caching decorator

PrimeGlorious:


ntkachenko21 (approved):

app/main.py
```diff
@@ -1,6 +1,18 @@
...
+
+    @wraps(func)
+    def wrapper(*args) -> Any:
+        if args in cache_dict:
+            print("Getting from cache")
+            return cache_dict[args]
+        else:
+            print("Calculating new result")
+            cache_dict[args] = func(*args)
+            return cache_dict[args]
```
oliasymak24: try to rewrite your decorator not to have `return` duplicated

app/main.py
```diff
@@ -1,6 +1,18 @@
...
+from functools import wraps
 
 
 def cache(func: Callable) -> Callable:
-    # Write your code here
-    pass
+    cache_dict = {}
+
+    @wraps(func)
+    def wrapper(*args) -> Any:
```
oliasymak24: Fix your annotation

app/main.py
```diff
@@ -1,6 +1,18 @@
...
+        if args in cache_dict:
+            print("Getting from cache")
+            return cache_dict[args]
```
oliasymak24: Pay attention that in this case args is tuple, so it can be hashable, but if it is list you will have a Type Error unhashable type: 'list'

oliasymak24 (changes requested):

app/main.py
```diff
@@ -1,6 +1,21 @@
...
+from functools import wraps
 
 
 def cache(func: Callable) -> Callable:
-    # Write your code here
-    pass
+    cache_dict = {}
+
+    @wraps(func)
+    def wrapper(*args, **kwargs) -> Any:
```
oliasymak24, line 9: *args: Any

oliasymak24 (approved):

PR #1562, opened 2025-02-17
task solution

yarms1:


app/main.py
```diff
@@ -1,6 +1,18 @@
...
+    @wraps(func)
+    def wrapper(*args: Any, **kwargs: Any) -> Any:
+        key = (args, tuple(sorted(kwargs.items())))
+        if key in cache_store:
+            print("Getting from cache")
+            return cache_store[key]
+        print("Calculating new result")
+        result = func(*args, **kwargs)
+        cache_store[key] = result
+        return result
```
PrimeGlorious: Try to rewrite using a single return statement

PrimeGlorious (changes requested):

PR #1563, opened 2025-02-17
'Solution'

RuslanSamsonovych:


app/main.py
```diff
@@ -1,6 +1,19 @@
...
 
 
 def cache(func: Callable) -> Callable:
-    # Write your code here
-    pass
+    storage = {}
+
+    @wraps(func)
+    def inner(*args) -> Any:
+        temp = tuple(args)
```

vsmutok (changes requested):

vasylhnatiuk (approved):

PR #1564, opened 2025-02-17
implementing cash decorator

Maaax256:


app/main.py
```diff
@@ -1,6 +1,15 @@
-from typing import Callable
-
+from typing import Callable, Any
 
+cash_dict = {}
 def cache(func: Callable) -> Callable:
```
Anollinna, line 6: Add 2 blank lines before the function))

Anollinna (changes requested):
Add 2 blank lines before the function))

app/main.py
Maaax256: Done

PR #1570, opened 2025-02-20
implementing cash decorator (corrected)

Maaax256:


app/main.py
```diff
@@ -1,6 +1,19 @@
...
+    def inner(*args) -> Any:
+        if func not in cash_dict:
+            cash_dict[func] = {}
+        if args not in cash_dict[func]:
+            cash_dict[func][args] = func(*args)
+            print("Calculating new result")
+            return cash_dict[func][args]
+        else:
+            print("Getting from cache")
+            return cash_dict[func][args]
```

app/main.py
```diff
@@ -1,6 +1,19 @@
...
+import functools
+
+cash_dict = {}
 
 
 def cache(func: Callable) -> Callable:
-    # Write your code here
-    pass
+    @functools.wraps(func)
+    def inner(*args) -> Any:
```
oliasymak24: Add an annotations for `*args`

oliasymak24 (changes requested):

app/main.py
```diff
@@ -1,6 +1,18 @@
...
+import functools
+
+cash_dict = {}
 
 
 def cache(func: Callable) -> Callable:
-    # Write your code here
-    pass
+    @functools.wraps(func)
+    def inner(*args: Tuple[Union[int, str, Tuple[int]]]) -> Any:
```
oliasymak24, line 9: Just Any will be enough))

oliasymak24 (approved):

PR #1592, opened 2025-03-06
'Solution'

AndyStarGitHub:


app/main.py
```diff
@@ -1,6 +1,39 @@
...
+    return result
+
+
 def cache(func: Callable) -> Callable:
-    # Write your code here
-    pass
+    func_name = func.__name__
+    global_result = read_cache()
+
+    def inner(*args) -> list:
```
Boyok1337: i d like to recomend use `@wraps(func)` to preserve the function metadata

app/main.py
```diff
@@ -1,6 +1,39 @@
...
 
 
+def read_cache() -> dict:
+    result = dict()
+    if os.path.isfile("cache.txt"):
+        with open("cache.txt", "r") as cache_file:
+            for line in cache_file.readlines():
+                parts = line.split(":")
+                if not result.get(parts[0]):
+                    result[parts[0]] = parts[1][:-1]
```
Boyok1337: Maybe here good to use `.strip()`?

app/main.py
```diff
@@ -1,6 +1,39 @@
...
+        if result_inner is None:
+            print("Calculating new result")
+            result_inner = func(*args)
+            global_result[one_key] = result_inner
+        else:
+            print("Getting from cache")
+        save_cache(global_result)
+        return result_inner
+
+    return inner
```
Boyok1337: Think about how this can be implemented through the dictionary

Boyok1337 (changes requested):

app/main.py
AndyStarGitHub: Fixed, will be committed later with other issues.

app/main.py
AndyStarGitHub: Fixed, will be committed later with other issues.

app/main.py
AndyStarGitHub: I must say, that I don't understand what you mean. Could you please give me a hint?

AndyStarGitHub: Refactored with dictionary.

app/main.py
```diff
@@ -1,6 +1,36 @@
...
+
+
 def cache(func: Callable) -> Callable:
-    # Write your code here
-    pass
+    func_name = func.__name__
+
+    global cache_dict
+    cache_dict = dict()
+    global_result = read_cache_dict()
```
Boyok1337: Global variables for global scopes, so define it at the beginning of the file. Then update `read_cache_dict` func to return this values and reuse it in `cache` func

Boyok1337 (changes requested):

app/main.py
AndyStarGitHub: cache_dict and global_result were just a duplication. I eliminated the global_result variable.

oliasymak24 (approved):

PR #1618, opened 2025-03-21
Add cache decorator to optimize function calls.

farzamgt:


app/main.py
```diff
@@ -1,6 +1,18 @@
-from typing import Callable
+from typing import Callable, Dict, Tuple, Any
 
 
 def cache(func: Callable) -> Callable:
-    # Write your code here
-    pass
+    cache_storage: Dict[Tuple, Any] = {}
```
MNDonut: you can use built-in classes like `dict`, `tuple`

app/main.py
```diff
@@ -1,6 +1,18 @@
...
+
+    def wrapper(*args, **kwargs) -> Any:
+        key = (args, frozenset(kwargs.items()))
+        if key in cache_storage:
+            print("Getting from cache")
+            return cache_storage[key]
+        else:
```
MNDonut: no need to use `else` statement because `if` statement has `return` keyword, thus if this statement returns `cache_storage[key]`, then everything else will be unreachable

app/main.py
```diff
@@ -1,6 +1,18 @@
...
+from typing import Callable, Dict, Tuple, Any
 
 
 def cache(func: Callable) -> Callable:
-    # Write your code here
-    pass
+    cache_storage: Dict[Tuple, Any] = {}
+
+    def wrapper(*args, **kwargs) -> Any:
+        key = (args, frozenset(kwargs.items()))
```
MNDonut, line 8: actually in this case you can make `args` as a key because parameters are not passed into `kwargs`

MNDonut (changes requested):

app/main.py
farzamgt: I agree but I kept key = (args, frozenset(kwargs.items())) for simplicity and consistency. It works reliably with or without kwargs and avoids unnecessary conditionals.

farzamgt:
@MNDonut 
Hi! Could you take another look at my work?

AnyoneClown (approved):
LGTM

PR #1631, opened 2025-03-26
Solution

willian001jp:


willian001jp (commented):
Correction function def wrapper for -> any

PR #1637, opened 2025-04-05
result

Salmon-eat:


app/main.py
```diff
@@ -2,5 +2,16 @@
...
-    # Write your code here
-    pass
+    cache_dict = {}
+
+    def wrapper(*args, **kwargs) -> Callable:
+        key = args
+        if key in cache_dict:
+            print("Getting from cache")
+            return cache_dict[key]
+        else:
```
MNDonut, line 12: no need to use `else` statement because `if` statement has `return` keyword, thus if this statement returns `cache_dict[key]`, then everything else will be unreachable

MNDonut (changes requested):

app/main.py
MNDonut: And one more note, create commits via `git`, not `github`

PR #1655, opened 2025-04-18
Solution

Nib20:


app/main.py
```diff
@@ -1,6 +1,18 @@
...
+    @wraps(func)
+    def wrapper(*args) -> Any:
+        if args in cache_memory:
+            print("Getting from cache")
+            return cache_memory[args]
+        else:
+            print("Calculating new result")
+            result = func(*args)
+            cache_memory[args] = result
+            return result
```
MNDonut: No need to use `else` because `if` statement has `return`, thus `if args is in cache memory`, it'll not reach the other part of the code

MNDonut (changes requested):

app/main.py
Nib20: fixed

MNDonut (approved):

PR #1665, opened 2025-04-28
writed decorator Cache

vasyl-main-dev:


app/main.py
```diff
@@ -1,6 +1,19 @@
...
+    def wrapper(*args, **kwargs) -> str:
+        keys = (args, tuple(sorted(kwargs.items())))
+        if keys in cache_dict:
+            print("Getting from cache")
+            return cache_dict[keys]
+        else:
+            print("Calculating new result")
+            res = func(*args, **kwargs)
+            cache_dict[keys] = res
+            return res
```
```suggestion
        else:
            print("Calculating new result")
            cache_dict[keys] =  func(*args, **kwargs)
        return cache_dict[keys]
```

This will simplify your code a bit.

vsmutok (changes requested):

app/main.py
vasyl-main-dev: Good advice, thanks. Just did it

app/main.py
```diff
@@ -1,6 +1,18 @@
...
+    @wraps(func)
+    def wrapper(*args, **kwargs) -> str:
+        keys = (args, tuple(sorted(kwargs.items())))
+        if keys in cache_dict:
+            print("Getting from cache")
+            return cache_dict[keys]
+        else:
+            print("Calculating new result")
+            cache_dict[keys] = func(*args, **kwargs)
+            return cache_dict[keys]
```
```suggestion
        keys = (args, tuple(sorted(kwargs.items())))
        if keys in cache_dict:
            print("Getting from cache")
        else:
            print("Calculating new result")
            cache_dict[keys] = func(*args, **kwargs)
        return cache_dict[keys]
```

You can use `return` only once.

vsmutok (changes requested):

vsmutok (approved):
Great!

PR #1700, opened 2025-05-22
Solution

Hetmarum:


app/main.py
```diff
@@ -1,6 +1,18 @@
...
-    # Write your code here
-    pass
+    storage = {}
+
+    @wraps(func)
+    def wrapper(*args) -> None:
+        if args in storage:
+            print("Getting from cache")
+            return storage[args]
+        else:
```
MNDonut: no need to use `else` statement because `if` statement has `return` keyword, thus if this statement returns `storage[args]`, then everything else will be unreachable

MNDonut (changes requested):

AnyoneClown (approved):
LGTM

PR #1718, opened 2025-06-06
Solution

PavloReutskiy:


app/main.py
```diff
@@ -2,5 +2,18 @@
 
 
 def cache(func: Callable) -> Callable:
-    # Write your code here
-    pass
+    cashed_data = {}
+
+    def wrapper(*args, **kwargs) -> Callable:
+        cashed_key = tuple(args)
```
MNDonut, line 8: args has tuple type by default, so no need to convert it

app/main.py
```diff
@@ -2,5 +2,18 @@
...
+
+        if cashed_key in cashed_data:
+            print("Getting from cache")
+            return cashed_data[cashed_key]
+        else:
```
MNDonut, line 13: no need to use `else` because `if` statement has `return`, so in this case `else` statement will be unreachable

MNDonut (changes requested):

PR #1731, opened 2025-06-17
Solution

Orixara:


DrGregoryHouse4676 (approved):

app/main.py
```diff
@@ -1,6 +1,18 @@
...
+from typing import Callable, Any
 
 
 def cache(func: Callable) -> Callable:
-    # Write your code here
-    pass
+    cache_dict = {}
+
+    def wrapper(*args, **kwargs) -> Any:
+        key = (func, args, tuple(sorted(kwargs.items())))
```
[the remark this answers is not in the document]
Orixara: @mateacademy-ai-mentor but we have that condition in our task:
"Also note, that decorator cache should work correctly with few decorated functions simultaneously and correctly return for every function separately."
That means if we have situation like that:

@cache
def func1(x):
    return x * 2

@cache  
def func2(x):
    return x * 3

func1(5)  # result: 10
func2(5) # result: 10
func2 have to be 15, but we have 10 due to the fact that in cache we have that keys but don't have exact function

Orixara:
@mateacademy-ai-mentor 
I have reviewed your comments, and I'd like to clarify a few points that were misunderstood:

we have that condition in our task:
"Also note, that decorator cache should work correctly with few decorated functions simultaneously and correctly return for every function separately."
That means if we have situation like that:

@cache
def func1(x):
return x * 2

@cache
def func2(x):
return x * 3

func1(5) # result: 10
func2(5) # result: 10
func2 have to be 15, but we have 10 due to the fact that in cache we have that keys but don't have exact function

app/main.py
```diff
@@ -1,6 +1,18 @@
...
+
+        if key in cache_dict:
+            print("Getting from cache")
+            return cache_dict[key]
+        else:
+            print("Calculating new result")
+            result = func(*args, **kwargs)
+            cache_dict[key] = result
+            return result
```
L1nk3rrr: You can try to write it shorter.
Also here can be applied 1 return instead of 2

L1nk3rrr (changes requested):

app/main.py
Orixara: Thank you, can you check my changes? did I get it right? Or can I make it  shorter more?

AnyoneClown (approved):
LGTM

PR #1732, opened 2025-06-17
Solution Cache decorator

yagginov:


app/main.py
```diff
@@ -3,16 +3,19 @@
 
 
 def cache(func: Callable) -> Callable:
-    cache_dict = {}
+    cache_dict = {}  # variable for simple cache
```
DrGregoryHouse4676, line 7: Why are the decorators being so scattered?

DrGregoryHouse4676 (approved):

app/main.py
```diff
@@ -1,6 +1,28 @@
...
 
 
 def cache(func: Callable) -> Callable:
-    # Write your code here
-    pass
+    cache_dict = {}  # variable for simple cache
+
+    @wraps(func)
+    def inner(*args, **kwargs) -> Any:
+        nonlocal cache_dict
```
L1nk3rrr: You don't need to set it as a nonlocal variable, cause inner func is inside of cache and has access to its variable scope

app/main.py
```diff
@@ -1,6 +1,28 @@
...
+
+        func_name = func.__name__  # save function name in variable
+        # if a function is called for the first time,
+        # then we create a new place for it
+        if func_name not in cache_dict:
+            cache_dict[func_name] = {}
```
L1nk3rrr: As the decorator is set per function, you don't need to save a function name. So your cache must looks like 
key - args, kwargs (hashable object)
value - result of function

L1nk3rrr (changes requested):
Please review my comments to make your cache function simpler

L1nk3rrr (approved):
Well done!

PR #1735, opened 2025-06-18
Solution

Plovuezz:


DanilHamster (approved):

app/main.py
```diff
@@ -1,6 +1,28 @@
...
+    @wraps(func)
+    def wrapper(*args: tuple) -> int:
+        if func.__name__ in _cache_dict:
+            if args in _cache_dict[func.__name__]:
+                print("Getting from cache")
+                return _cache_dict[func.__name__][args]
+
+        for arg in args:
+            if not isinstance(arg, (int, float, bool, str, tuple)):
+                return func(*args)
```
L1nk3rrr: what the reason of this?

L1nk3rrr (changes requested):
What about not only args but also kwargs, if we want to talk about any functions?

Also, it looks too complicated, so I try to simplify it.
Also - test a cache decorator for different functions, see how it applies, and then think about - do you need a func? _ _ name__ in your key of _cache_dict

AnyoneClown (approved):
lgtm

PR #1744, opened 2025-06-20
cache decorator implemented

dimashevchukk:


MakovM (approved):

app/main.py
```diff
@@ -2,5 +2,16 @@
...
+
+    def wrapper(*args, **kwargs) -> int:
+        if args in cached_data:
+            print("Getting from cache")
+            return cached_data[args]
+        else:
+            print("Calculating new result")
+            result = func(*args, **kwargs)
+            cached_data[args] = result
+            return result
```
L1nk3rrr: You can use here 1 return instead of 2

L1nk3rrr (changes requested):
Also look for AI comment about kwargs and try to implement it.

L1nk3rrr (approved):
Well done!

PR #1749, opened 2025-06-21
Solution

SensibleN00B:


Vskesha (approved):
This decorator works correctly. Code style is good. You can also add **kwargs handling if you want to extend functionality of the decorator.

PR #1753, opened 2025-06-23
Solution

zqpiq:


app/main.py
```diff
@@ -2,5 +2,15 @@
...
+
+    def wrapper(*args, **kwargs) -> None:
+        key = (args, tuple(sorted(kwargs.items())))
+        if key in result:
+            print("Getting from cache")
+            return result[key]
+        else:
+            print("Calculating new result")
+            result[key] = func(*args, **kwargs)
+            return result[key]
```
L1nk3rrr: You can use 1 return instead of 2.

Simply make `return result[key]` out of `if-else` block

L1nk3rrr (changes requested):
Well done! Minor changes needed for improvements!

AnyoneClown (approved):
LGTM

PR #1756, opened 2025-06-24
created decorator for functions

trellle:


app/main.py
```diff
@@ -1,6 +1,27 @@
...
+                my_store[str(func)].update({str(args): result})
+                return result
+            elif str(func) in my_store and str(args) in my_store[str(func)]:
+                print("Getting from cache")
+                return my_store[str(func)][str(args)]
+            elif str(func) not in my_store:
+                print("Calculating new result")
+                result = func(*args)
+                my_store.update({str(func): {str(args): result}})
+                return result
```
MNDonut: Ahh, you can simplify it and make it in more pythonic way.
At first you don't need to validate args. It isn't mentioned that there may be wrong values. Even if there are invalid values (that isn't our case), you don't print anything because in this case FLAG = False and you don't return anything inside wrapper in this case

The second thing is that your dict has keys in a format:
`args: calculated value`
So you can just check if args in this dict, then return value, otherwise calculate this value and print an appropriate message

MNDonut (changes requested):
Need to refactor it

app/main.py
trellle: my dict has format {func: {args: calculated value}}, there're different functions in tests

app/main.py
```diff
@@ -1,6 +1,18 @@
...
+from typing import Callable, Any
 
 
 def cache(func: Callable) -> Callable:
-    # Write your code here
-    pass
+    my_store = {}
+
+    def wrapper(*args) -> Any:
+        if str(func) in my_store and str(args) in my_store[str(func)]:
```
MNDonut: why do you convert everything in string? you don't need to do that

app/main.py
```diff
@@ -1,6 +1,18 @@
...
+            print("Getting from cache")
+            return my_store[str(func)][str(args)]
+        print("Calculating new result")
+        result = func(*args)
+        if str(func) in my_store and str(args) not in my_store[str(func)]:
+            my_store[str(func)].update({str(args): result})
+        else:
+            my_store.update({str(func): {str(args): result}})
+        return result
```
MNDonut: the same question here

also what is the difference between
`my_store[str(func)].update({str(args): result})`
and
`my_store.update({str(func): {str(args): result}})`
?
In both cases you update a value for a function anyway, so it's the same expression meaning you don't need this if statement at all

app/main.py
```diff
@@ -1,6 +1,18 @@
...
 
 def cache(func: Callable) -> Callable:
-    # Write your code here
-    pass
+    my_store = {}
+
+    def wrapper(*args) -> Any:
+        if str(func) in my_store and str(args) in my_store[str(func)]:
+            print("Getting from cache")
+            return my_store[str(func)][str(args)]
```
MNDonut: why str?
func is immutable, args are tuple that is also immutable

MNDonut (changes requested):

app/main.py
trellle: yeah, strings really were excessive, when I wrote code I had error with if statements because they were mutable. I've removed them

app/main.py
trellle: first line updates value via key (function), second line creates function key with dict of parameters. I tried to use second lin for both cases but tests return two errors
`================================================================ short test summary info ================================================================ 
FAILED tests/test_main.py::test_cache_single_function - AssertionError: 

FAILED tests/test_main.py::test_cache_multiple_functions - AssertionError:
`

trellle: mby update overwrites values of func key?

app/main.py
```diff
@@ -1,6 +1,18 @@
...
+        if func in my_store and args in my_store[func]:
+            print("Getting from cache")
+            return my_store[func][args]
+        print("Calculating new result")
+        result = func(*args)
+        if func in my_store and args not in my_store[func]:
+            my_store[func].update({args: result})
+        else:
+            my_store.update({func: {args: result}})
+        return result
```
MNDonut: 1. Why do you save func in this dict? When you decorate two functions with `cache` decorator, both of them have a separate `my_store` dict, so you can simplify structure of this dict
2. You don't need to use `update` method, just assign a new value to a specific key

MNDonut (changes requested):

MNDonut (approved):
And now it looks pretty good :)

PR #1763, opened 2025-06-28
Solution

reinvvay:


app/main.py
```diff
@@ -1,6 +1,17 @@
-from typing import Callable
+from typing import Callable, Any
 
 
 def cache(func: Callable) -> Callable:
-    # Write your code here
-    pass
+    func_cache = {}
+
+    def wrapper(*args: Any) -> Any:
```
Kostenikov, line 7:
```suggestion
    def wrapper(*args: tuple) -> Any:
```
Annotation is not needed in this case, or if you want do this: use **`tupple`** for args.

app/main.py
```diff
@@ -1,6 +1,17 @@
...
+        if args in func_cache:
+            print("Getting from cache")
+            return func_cache[args]
+        else:
+            result = func(*args)
+            print("Calculating new result")
+            func_cache[args] = result
+            return result
```
Kostenikov, line 15:
```suggestion
            func_cache[args] = func(*args)
            print("Calculating new result")            
            return func_cache[args]
```
Code can be more simplify if you dont use one more variable.

Kostenikov (approved):
Look at my comments.

PR #1767, opened 2025-07-03
Solution

Evgeniya0208:


app/main.py
```diff
@@ -2,5 +2,17 @@
...
+        key = (func, args, tuple(sorted(kwargs.items())))
+        if key in result_func_dict:
+            result = result_func_dict[key]
+            print("Getting from cache")
+            return result
+        else:
+            result = func(*args, **kwargs)
+            result_func_dict[key] = result
+            print("Calculating new result")
+            return result
```
```suggestion
        if key in result_func_dict:
            print("Getting from cache")
        else:
            result_func_dict[key] =  func(*args, **kwargs)
            print("Calculating new result")
        return result_func_dict[key]
```

app/main.py
```diff
@@ -2,5 +2,17 @@
 
 
 def cache(func: Callable) -> Callable:
-    # Write your code here
-    pass
+    result_func_dict = {}
+
```

vsmutok (changes requested):

app/main.py
Evgeniya0208: fixed

app/main.py
Evgeniya0208: fixed

vsmutok (approved):
Looks Good!

PR #1799, opened 2025-08-04
Implement py-cache-decorator

Anton-V10:


linubah (approved):
Nicely done — the task was completed as expected. 👍
P.S. One suggestion for improvement: currently, if the same input values are passed using different argument formats — e.g., f(2, 3) f(2, b=3) — the cache treats them as separate calls. This results in duplicated cache entries for logically identical inputs. Normalizing args and kwargs into a unified key could help.

PR #1894, opened 2025-10-17
add chache, pass all tests

VadymVynokur:


app/main.py
```diff
@@ -1,6 +1,16 @@
...
+    def inner(*args) -> Any:
+        if args in used:
+            print("Getting from cache")
+            return used[args]
+
+        else:
+            print("Calculating new result")
+            used[args] = func(*args)
+            return used[args]
+    return inner
```
petrykivd, line 16: Well done! 
You can also simplify this solution a little by using only one "return" after the "if else" construction.

petrykivd (approved):

PR #1902, opened 2025-10-22
Added solution

vadym-zahrebelnyi:


stanislav-kishechko (approved):
I advise you to add a doc string for the decorator. On line 7, change the type that is returned to a dictionary. The moment with wrapper.cached_calls = dict() It is better to simply create a dictionary, rather than create an attribute for the wrapper as this complicates the readability of the code

PR #1903, opened 2025-10-22
Solution

stanislav-kishechko:


app/main.py
```diff
@@ -1,6 +1,55 @@
...
+        func (Callable): The function to be decorated. Should accept only
+                        immutable arguments for proper cache key generation.
+
+    Returns:
+        Callable: A wrapped version of the function with caching functionality.
+    """
+    func_cache = {}
+
+    @wraps(func)
+    def wrapper(*args, **kwargs) -> Optional[Dict]:
```
leemonch1k-space, line 31: annotaion "Optional[Dict]" mean that we will return Dict or None, but in this case we return result of long function which has type int or float. Of course we do not know what object we will get as an argument so for safty more accaptable will be " -> Any"

stanislav-kishechko: More correct annotation type is Dict | Callable

leemonch1k-space (approved):
The task is well solved. More of it, you added very detailed and simple to understanding documentation. Well done

PR #1904, opened 2025-10-22
Solution

leemonch1k-space:


app/main.py
```diff
@@ -2,5 +2,17 @@
 
 
 def cache(func: Callable) -> Callable:
-    # Write your code here
```
Taras-Ivashchuk, line 5: please remove comments and 'pass'  word

Taras-Ivashchuk (changes requested):

app/main.py
leemonch1k-space: What are you talking about? I have no comments, and "pass" in my commit?

PR #1909, opened 2025-10-24
Solution

oleksii-shcherbak-code:


AngryAndrii (approved):
Good work! But using "Any" for wrapper type is bad idea. It's better to type it as "Callable"

oleksii-shcherbak-code:
> Good work! But using "Any" for wrapper type is bad idea. It's better to type it as "Callable"

Thanks. 
Could you please tell me why in this particular case it is better to use Callable rather than Any?

AngryAndrii:
> > Good work! But using "Any" for wrapper type is bad idea. It's better to type it as "Callable"
> 
> Thanks. Could you please tell me why in this particular case it is better to use Callable rather than Any?

You could use "Any", in this task it is not critical. But it's bad practice to do it all the time, becouse your code became less informative and  more unpredictable

PR #1915, opened 2025-10-24
Solution

Najttsu:


app/main.py
```diff
@@ -1,6 +1,20 @@
...
+from functools import wraps
 
 
 def cache(func: Callable) -> Callable:
-    # Write your code here
-    pass
+    _cache = {}
+
+    @wraps(func)
+    def wrapper(*args: Any) -> Any:
```
leemonch1k-space: wrapper should take named argument as parameter too ->def wrapper(*args, **kwargs)  -> Any:
cache should contain kwargs as key too, you can use a tuple to combine args and kwargs in this case

leemonch1k-space (changes requested):
If the cache key logic will be rewritten and you use it to find cache value, the task will be completed.

app/main.py
Najttsu: > wrapper should take named argument as parameter too ->def wrapper(*args, **kwargs) -> Any: cache should contain kwargs as key too, you can use a tuple to combine args and kwargs in this case

It is not necessarily for this task, couse we dont get kwargs as a parameters, only args

leemonch1k-space (approved):
Well done

PR #1918, opened 2025-10-24
Released @cache decorator

n-korobko:


app/main.py
```diff
@@ -1,6 +1,38 @@
-from typing import Callable
+from typing import Callable, Any
 
+_function_caches = {}
```
tradewcs: Remove global _function_caches variable. It’s better to keep the cache in the closure, so the decorator is self-contained.

app/main.py
```diff
@@ -1,6 +1,38 @@
...
 
-def cache(func: Callable) -> Callable:
-    # Write your code here
-    pass
+
+def cache(func: Callable) -> Callable[..., Any]:
+    _function_caches[func] = {}
+
+    def wrapper(*args, **kwargs) -> Any:
```
tradewcs, line 9: add wrap decorator from functools to keep func metadata (like \_\_name\_\_, \_\_doc\_\_)

app/main.py
```diff
@@ -1,6 +1,38 @@
...
+        key = args + tuple(sorted(kwargs.items()))
+        if key in _function_caches[func]:
+            print("Getting from cache")
+            return _function_caches[func][key]
+        else:
```
tradewcs: else is redundant in this case

tradewcs (changes requested):

tradewcs (approved):

PR #1922, opened 2025-10-27
made 2 functions make_key, cache

AlexMishKo:


tigersaymon (changes requested):
Failed flake8
Double check that you really need another func for creating a key, it's better to implement def make_key into def cache

PR #1935, opened 2025-11-05
'Solution'

Nahorni:


app/main.py
```diff
@@ -2,5 +2,15 @@
 
 
 def cache(func: Callable) -> Callable:
-    # Write your code here
-    pass
+    cache_dict = dict()
+
+    def wrapper(*args) -> int | list:
+        nonlocal cache_dict
```
vsmutok: 1. If the function also accepts keyword arguments, you should take them into account as well.
2. Don’t forget to use `functools.wraps`.

vsmutok (changes requested):

app/main.py
```diff
@@ -1,6 +1,19 @@
...
+    def wrapper(*args, **kwargs) -> Any:
+        nonlocal cache_dict
+        key = (args, tuple(sorted(kwargs.items())))
+        if key not in cache_dict:
+            cache_dict[key] = func(*args, **kwargs)
+            print("Calculating new result")
+            return cache_dict[key]
+        else:
+            print("Getting from cache")
+            return cache_dict[key]
```
```suggestion
        else:
            print("Getting from cache")
        return cache_dict[key]
```

vsmutok (changes requested):

vsmutok (approved):
Great!

PR #1948, opened 2025-11-12
Solution

tarasmosiichuk01-ship-it:


app/main.py
```diff
@@ -1,6 +1,18 @@
-from typing import Callable
+from typing import Callable, Any
 
 
 def cache(func: Callable) -> Callable:
-    # Write your code here
-    pass
+    cache_storage = {}
+
```
vsmutok, line 7: don’t forget to use`@functools.wraps`

app/main.py
```diff
@@ -1,6 +1,18 @@
...
+    def wrapper(*args: Any, **kwargs: Any) -> Any:
+        key = (args, tuple(sorted(kwargs.items())))
+        if key in cache_storage:
+            print("Getting from cache")
+            return cache_storage[key]
+        else:
+            print("Calculating new result")
+            result = func(*args, **kwargs)
+            cache_storage[key] = result
+            return result
```
```suggestion
        if key in cache_storage:
            print("Getting from cache")
        else:
            print("Calculating new result")
            result = func(*args, **kwargs)
        return cache_storage[key]
```

vsmutok (changes requested):

app/main.py
tarasmosiichuk01-ship-it: I don't understand what's wrong here

DaTrEvTeR (approved):
Well done!

PR #1973, opened 2025-12-02
Created cache decorator using dict

andriiyarotskiy:


andriiyarotskiy:
Чи правильно я використав тут dict для зберігання пари ключ:значення де ключі це args? І чи це взагалі нормальна практика зберігати аргументи як ключі?

app/main.py
```diff
@@ -1,6 +1,15 @@
...
+from typing import Callable, Any
 
 
 def cache(func: Callable) -> Callable:
-    # Write your code here
-    pass
+    cache_storage = {}
+
+    def decorator(*args) -> Any:
+        if args not in cache_storage:
```
vsmutok, line 8: What if the function accepts keyword arguments (kwargs) and not only positional ones (args)?
Then your logic will not work. So you need to check both kwargs and args

Also, don’t forget to use `functools.wraps`

vsmutok (changes requested):

app/main.py
andriiyarotskiy: @vsmutok Thanks. Good point about `functools.wraps`. 

I didn't cover this case with kwargs because in the description of the task we had
> Decorator cache creating for decorating only functions that take immutable arguments.

And as far as I know, kwards is a dict data type and it can't be a key of `cache_storage`: dict in my solution, unlike tuple (args), can it? As I think, in real scenario we will have to use maybe some id or extract mutable data values from kwargs for further comparison and assigning them as a dict key.

PR #2027, opened 2026-01-12
Solution

py-ovarion:


py-ovarion:
@mateacademy-ai-mentor according to readmi.md "Decorator cache creating for decorating only functions that take immutable arguments." but **kwargs is mutable. So how can I add it to the cache?

PR #2054, opened 2026-01-25
Solution

AndriiOne:


app/main.py
```diff
@@ -1,6 +1,19 @@
...
+        key = (args, frozenset(kwargs.items()))
+
+        if key in stored_results:
+            print("Getting from cache")
+            return stored_results[key]
+        else:
+            result = func(*args, **kwargs)
+            stored_results[key] = result
+            print("Calculating new result")
+            return result
```
vsmutok, line 20:
```suggestion
        if key in stored_results:
            print("Getting from cache")
        else:
            stored_results[key] = func(*args, **kwargs)
            print("Calculating new result")
        return stored_results[key]
```

app/main.py
```diff
@@ -1,6 +1,19 @@
 from typing import Callable
+from typing import Any
 
 
 def cache(func: Callable) -> Callable:
-    # Write your code here
-    pass
+    stored_results = {}
+
+    def wrapper(*args, **kwargs) -> Any:
```
vsmutok, line 10: Don’t forget to use `@functools.wraps`.

vsmutok (changes requested):

app/main.py
AndriiOne: Added fuctools. Thx

app/main.py
AndriiOne: Add your suggest, thank you

PR #2063, opened 2026-01-29
Solution

Ensenkei:


DaTrEvTeR (approved):
In your solution, you ignore the possibility that the data in the arguments may be unhashable or have different keyword arguments, but for the conditions of this problem, that's great!

PR #2067, opened 2026-02-02
Implement cache-decorator

AndS9:


AndS9 (commented):
I changed solution according to your comments

AndS9 (commented):
Fix cache decorator according to yours requirement

PR #2076, opened 2026-02-07
Solution

KovalchukAlexey:


app/main.py
```diff
@@ -1,6 +1,16 @@
...
 
 
 def cache(func: Callable) -> Callable:
-    # Write your code here
-    pass
+    cache_data = {}
+
+    def wrapper(*args, **kwargs) -> ReturnType:
+        key = (args, tuple(sorted(kwargs.items())))
+        if key not in cache_data.keys():
```
vsmutok, line 10: This check works correctly, but it's more idiomatic and slightly more efficient in Python to check for key existence directly in the dictionary, like this: `if key not in cache_data:`.
 
The `.keys()` method creates a view object of the keys, which is an unnecessary step here.

app/main.py
```diff
@@ -1,6 +1,16 @@
...
+from typing import Callable, TypeVar
+ReturnType = TypeVar("ReturnType")
 
 
 def cache(func: Callable) -> Callable:
-    # Write your code here
-    pass
+    cache_data = {}
+
+    def wrapper(*args, **kwargs) -> ReturnType:
```
vsmutok, line 8: Don’t forget to use `@functools.wraps`.

vsmutok (changes requested):

PR #2082, opened 2026-02-11
Solution by Kryvutsa

pawunder1992:


app/main.py
```diff
@@ -2,5 +2,15 @@
...
-    # Write your code here
-    pass
+    result = {}
+
+    def wrapper(*args) -> int:
+        if args not in result:
+            result[args] = func(*args)
+            print("Calculating new result")
+            return result[args]
+        else:
```
SerhiiHryniuk: You can actually drop the else and just keep the return statement at the bottom to keep the indentation level low but that is not a such a big problem

SerhiiHryniuk (approved):

app/main.py
pawunder1992: Ok, i understand, thnx

PR #2111, opened 2026-03-01
complete task

PavloDanylkiv:


app/main.py
```diff
@@ -2,5 +2,18 @@
...
+        key = (args, tuple(sorted(kwargs.items())))
+
+        if key in storage:
+            print("Getting from cache")
+            return storage[key]
+
+        print("Calculating new result")
+        result = func(*args, **kwargs)
+        storage[key] = result
+        return result
```
vsmutok, line 17:
```suggestion
        if key in storage:
            print("Getting from cache")
        else:
            print("Calculating new result")
            storage[key] =  func(*args, **kwargs)
        return storage[key]
```
It will look a bit more compact)

vsmutok (changes requested):

PR #2118, opened 2026-03-07
'Solution'

matus332:


app/main.py
```diff
@@ -2,5 +2,16 @@
...
+
+    def wrapper(*args) -> Callable:
+        if args in cache_dict:
+            print("Getting from cache")
+            return cache_dict[args]
+        else:
+            print("Calculating new result")
+            result = func(*args)
+            cache_dict[args] = result
+            return result
```
```suggestion
        if args in cache_dict:
            print("Getting from cache")
        else:
            print("Calculating new result")
            cache_dict[args] = func(*args)
        return cache_dict[args]
```

app/main.py
```diff
@@ -2,5 +2,16 @@
 
 
 def cache(func: Callable) -> Callable:
-    # Write your code here
-    pass
+    cache_dict = {}
+
+    def wrapper(*args) -> Callable:
+        if args in cache_dict:
```
vsmutok, line 10: Don’t forget to use `functools.wraps`.

vsmutok (changes requested):

PR #2148, opened 2026-03-24
Solution

artradkevych:


alisahoys (approved):
Only works with positional arguments (*args)

PR #2152, opened 2026-03-25
Cache

VictorMatveyshyn:


VictorMatveyshyn:
ready

app/main.py
```diff
@@ -1,6 +1,18 @@
...
 
 
 def cache(func: Callable) -> Callable:
-    # Write your code here
-    pass
+    cache2 = {}
+
+    def wrapper(*args, **kwargs) -> Any:
+
+        key = ":".join([str(num) for num in args])
```
vsmutok: The key must include both `args` and `kwargs`, because different functions can have the same `args` but different `kwargs`.

app/main.py
```diff
@@ -1,6 +1,18 @@
-from typing import Callable
+from typing import Callable, Any
 
 
 def cache(func: Callable) -> Callable:
-    # Write your code here
-    pass
+    cache2 = {}
```
```suggestion
    cache = {}
```

app/main.py
```diff
@@ -1,6 +1,18 @@
...
+
+    def wrapper(*args, **kwargs) -> Any:
+
+        key = ":".join([str(num) for num in args])
+        answer = cache2.get(key)
+        if answer is None:
```
vsmutok: Checking `if answer is None:` is not a reliable way to see if a result is in the cache. If the decorated function legitimately returns `None`, your code will treat it as a cache miss and re-calculate it every time. A more robust approach is to check for the key's existence directly in the `cache1` dictionary.

app/main.py
```diff
@@ -1,6 +1,18 @@
-from typing import Callable
+from typing import Callable, Any
 
 
 def cache(func: Callable) -> Callable:
-    # Write your code here
-    pass
+    cache2 = {}
+
+    def wrapper(*args, **kwargs) -> Any:
```
vsmutok, line 9: Don’t forget to use `functools.wraps`.

vsmutok (changes requested):

vsmutok (changes requested):
Please do what your mentor recommends.

app/main.py
```diff
@@ -1,6 +1,26 @@
-from typing import Callable
+from typing import (
+    Callable,
+    Any
+)
```
```suggestion
from typing import Callable, Any
```

app/main.py
```diff
@@ -1,6 +1,26 @@
...
-    pass
+    cache = {}
+
+    @wraps(func)
+    def wrapper(*args, **kwargs) -> Any:
+
+        key = ":".join([str(num) for num in args])
+        key2 = ":".join([str(key) + ":" + str(val)
+                         for key, val in kwargs.items()])
+        key = key + ":" + key2
```
```suggestion
        key = args, tuple(sorted(kwargs.items()))
```

app/main.py
```diff
@@ -1,6 +1,26 @@
...
+        if key not in cache:
+            print("Calculating new result")
+            answer = func(*args, **kwargs)
+            cache[key] = answer
+        else:
+            answer = cache.get(key)
+            print("Getting from cache")
```
vsmutok, line 15:
```suggestion
        if key not in cache:
            print("Calculating new result")
            cache[key] = func(*args, **kwargs)
        else:
            print("Getting from cache")
```

app/main.py
```diff
@@ -1,6 +1,26 @@
...
+        return answer
```
```suggestion
        return cache.get(key)
```

vsmutok (changes requested):

PR #2156, opened 2026-03-25
Develop

andriihrechko:


app/main.py
```diff
@@ -2,5 +2,15 @@
 
 
 def cache(func: Callable) -> Callable:
-    # Write your code here
-    pass
+    cache_data = {}
+
+    def wrapper(*args, **kwargs) -> None:
```
andriihrechko: upd: def wrapper(*args, **kwargs) -> **int:**

app/main.py
```diff
@@ -2,5 +2,15 @@
...
+        if args in cache_data:
```
Sacchar20: Logic error with kwargs: Your current implementation only uses args as a cache key: if args in cache_data. This means if the function is called with different keyword arguments (e.g., func(a=1) and func(a=2)), the decorator will incorrectly return the same cached result for both. You need to include both args and kwargs in your cache key to make it unique.

Sacchar20 (changes requested):

Sacchar20 (approved):

olha12232 (approved):
good

PR #2172, opened 2026-03-30
Solution

dem1hh:


app/main.py
```diff
@@ -1,6 +1,17 @@
...
+from typing import Any
 
 
 def cache(func: Callable) -> Callable:
-    # Write your code here
-    pass
+    cache_memory = {}
+
+    def inner(*args, **kwargs) -> Any:
+        if args in cache_memory:
```
vsmutok: **You are currently checking and using only `args` as the key, but it is possible for a function to receive the same `args` with different `kwargs`.**

**You need to create a combined key from both `args` and `kwargs`.**

app/main.py
```diff
@@ -1,6 +1,17 @@
 from typing import Callable
+from typing import Any
 
 
 def cache(func: Callable) -> Callable:
-    # Write your code here
-    pass
+    cache_memory = {}
+
+    def inner(*args, **kwargs) -> Any:
```
vsmutok, line 10: **Also, don’t forget to use `functools.wraps`.**

vsmutok (changes requested):

app/main.py
dem1hh: Fixed: include kwargs in cache key

Boyok1337 (approved):
LGTM!

PR #2176, opened 2026-04-01
Solution

lewszyk-karolina:


Czernich:
@mateacademy-ai-mentor I don’t think the cache is shared here — stored_dict lives inside the decorator, so each function gets its own copy via closure.

Czernich (approved):
Approved

PR #2180, opened 2026-04-03
Solution

mykytatitar-rgb:


app/main.py
```diff
@@ -1,6 +1,17 @@
...
+
+    def wrapper(*args, **kwargs) -> Any:
+        key = (args, tuple(sorted(kwargs.items())))
+        if key in storage_of_data:
+            print("Getting from cache")
+            return storage_of_data[key]
+        print("Calculating new result")
+        result = func(*args, **kwargs)
+        storage_of_data[key] = result
+        return result
```
```suggestion
        key = (args, tuple(sorted(kwargs.items())))
        if key in storage_of_data:
            print("Getting from cache")
        else:
          print("Calculating new result")
          storage_of_data[key] = func(*args, **kwargs)
       return storage_of_data[key]
```

app/main.py
```diff
@@ -1,6 +1,17 @@
...
+from typing import Callable, Any
 
 
-def cache(func: Callable) -> Callable:
-    # Write your code here
-    pass
+def cache(func: Callable) -> Any:
+    storage_of_data = {}
+
+    def wrapper(*args, **kwargs) -> Any:
```
vsmutok, line 9: Don’t forget to use `@functools.wraps`.

vsmutok (changes requested):

app/main.py
mykytatitar-rgb: Hello, Vasyl! Thank you, for you review. Yeah, line 11 and 13 were a bit extra)
You used "else" to make code more structured and readable?

app/main.py
mykytatitar-rgb: is there any significant difference between?:

1. from functool import wraps
@wraps(func)

2.@functools.wraps

PR #2183, opened 2026-04-04
Solution

gospodenkolubomir-oss:


app/main.py
```diff
@@ -1,6 +1,19 @@
 from typing import Callable
+import functools
 
 
-def cache(func: Callable) -> Callable:
-    # Write your code here
-    pass
+def cache(func: Callable) -> Callable[..., None]:
```
vsmutok, line 5: The return type hint `Callable[..., None]` is incorrect. The decorator returns a `wrapper` function which in turn returns the result of the original function, not `None`. Consider using `typing.Any` to indicate that the wrapped function can return a value of any type.

app/main.py
```diff
@@ -1,6 +1,19 @@
...
+    local_cache = {}
+
+    @functools.wraps(func)
+    def wrapper(*args, **kwargs) -> None:
```
vsmutok, line 9: The `wrapper`'s return type is hinted as `None`, but the function returns a value on lines 13 and 18. This should be corrected to reflect the actual return type, for example by importing `Any` from `typing` and using it here.

app/main.py
```diff
@@ -1,6 +1,19 @@
...
+        key = (args, tuple(sorted(kwargs.items())))
+        if key in local_cache:
+            print("Getting from cache")
+            return local_cache[key]
+        else:
+            print("Calculating new result")
+            result = func(*args, **kwargs)
+            local_cache[key] = result
+            return result
+    return wrapper
```
vsmutok, line 19:
```suggestion
        if key in local_cache:
            print("Getting from cache")
        else:
            print("Calculating new result")
            local_cache[key] = func(*args, **kwargs)
        return local_cache[key]
    return wrapper
```

vsmutok (changes requested):

PR #2208, opened 2026-04-21
Implement caching decorator for function call optimization

melmonreaz:


ostap1232 (changes requested):
This task is not part of the **Data Analyst** course.  
If you'd like to switch to a different course, please let your coordinator know.

app/main.py
```diff
@@ -2,5 +2,15 @@
...
 
 def cache(func: Callable) -> Callable:
-    # Write your code here
-    pass
+    memory = {}
+
+    def wrapper(*args, **kwargs) -> int:
+        # nonlocal memory
+        if args not in memory:
+            memory[args] = func(*args, **kwargs)
```
vsmutok: You are using only `args`, but you should also use `kwargs`, since a function can have both positional and keyword arguments.

app/main.py
```diff
@@ -2,5 +2,15 @@
 
 
 def cache(func: Callable) -> Callable:
-    # Write your code here
-    pass
+    memory = {}
+
+    def wrapper(*args, **kwargs) -> int:
```
vsmutok, line 9: Don’t forget to use `functools.wraps`.

vsmutok (changes requested):

vsmutok (approved):
Good Job!

PR #2282, opened 2026-06-14
feat: implement cache decorator

neeeksa:


app/main.py
```diff
@@ -2,5 +2,17 @@
 
 
 def cache(func: Callable) -> Callable:
-    # Write your code here
-    pass
+    cached = {}
+
+    def wrapper(*args, **kwargs) -> None:
+        items = str(args)
```
vsmutok, line 8: I recommend checking not only `args` but also `kwargs`, since the function may receive the same positional arguments while having different keyword arguments.

vsmutok (changes requested):

PR #2287, opened 2026-06-18
Create docorator

Kuper18:


app/main.py
```diff
@@ -1,6 +1,22 @@
...
 
 
 def cache(func: Callable) -> Callable:
-    # Write your code here
-    pass
+    cached_data = {}
+
+    @wraps(func)
+    def wrapper(*args, **kwargs) -> Any:
+        cache_key = (args, *kwargs, *kwargs.values())
```
DonDmytron, line 10: if I call the function first without named arguments and then with them, the calculation will be performed twice. define the key based on how the arguments are passed to you.

